PR #6, opened 2016-02-29
Fix conflicts with itsdangerous version

servingUpAces:


utils/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.1.2'
+__version__ = '0.1.3'
```
quis: Should be > 0.2.0

PR #10, opened 2016-03-08
Don’t return additional personalisation from a CSV

quis:
A valid CSV can have columns that aren’t in a template’s placeholders.

The API only accepts request that contain exactly the personalisation (values for placeholders) required for a template.

Previously the `.personalisation` and `.recipients_and_personalisation` properties of the `RecipientsCSV` util were returning values for _all_ columns of a CSV.

This commit changes these properties to only return the values that match the placeholders in the template.


ashimali:
:+1:

PR #11, opened 2016-03-10
Don’t mix errors with normal rows

quis:
Showing a preview of the CSV file where errors are intermingled with normal rows isn’t that clear. The reckon was that it would be good for context, but it’s probably clearer to only show the rows with errors.

This means that the util for processing CSVs can be made simpler, because its methods only have to return rows with errors, or all rows, never a mixture of the two. Which is why this commit is able to delete a load of tests.

Major version bump because this removes/renames a bunch of methods.

***

### Annotate each field with some useful metadata

…

```python
list(RecipientCSV("phone number,name\n+44123,Jo", template_type='sms').annotated_rows)
>>> {
>>>   'index': 0,
>>>   'phone number': {
>>>     'data': '+44123',
>>>     'error': 'Must be a valid UK mobile number',
>>>     'ignored': False
>>>   },
>>>   'name': {
>>>     'data': 'Jo',
>>>     'error': None,
>>>     'ignored': True
>>>   }
>>> }
```

`.initial_annotated_rows` will do the same thing, but limited to the number of
rows specified when passing `max_initial_rows_shown` to the constructor.

`.annotated_rows_with_errors` will do the same thing, but only return rows that
have errors.

`._initial_annotated_rows_with_errors` will show only errors, up to a limit of
`max_errors_shown` (as passed to the constructor).


quis:
(demo’d to @minglis )

PR #12, opened 2016-03-18
Add method to get a template as a HTML email

quis:
When we send email notifications, they should have consistent branding, ie with the GOV.UK header.

This commit adds:
- the email template from https://github.com/alphagov/email-template/blob/43b4e34bf5bc4910dc0358aa7a55f66bc7d530df/email_template_2.html (hardcoded as a string as a quick and dirty hack, but probably better than reading it from disk every time)
- a method to get the contents of a template, wrapped up as a HTML email

It also replaces linebreaks in the template content with `<br>` tags, so that all the text doesn’t get smushed into one line.


quis:
Demo’d to Pete and Martyn.

PR #13, opened 2016-03-21
Only prefix template with service name if it’s SMS

quis:
Only SMS templates should be prefixed with the name of the service. This commit modifies the behaviour of the Template class to only persist the prefix for SMS templates. This means that, if an email template is passed a prefix:

1. it won’t fail, in order to make the code in apps that use this class cleaner, ie don’t make them have to check the template type every time they call the template class.

2. it won’t return the prefix in the contents of the template, thus fixing the bug.

https://www.pivotaltracker.com/story/show/115879891

Minor version bump, because it fixes a bug in a non-breaking way.


servingUpAces:
:+1:

PR #14, opened 2016-03-24
Strip trailing empty rows from CSV files

quis:
Numbers on OS X will include in its export any empty cells which are displayed on screen. This results in a CSV file with lots of this at the end of it:

```
,,,
,,,
,,,
```

Currently the CSV validation throws an error, saying that data is missing for these rows.

Instead it should just ignore them.

This commit add test data and a fix in order to do so.


mannickutd:
:+1:

utils/recipients.py
```diff
@@ -28,7 +28,7 @@ def __init__(
         max_errors_shown=20,
         max_initial_rows_shown=10
     ):
-        self.file_data = file_data
+        self.file_data = file_data.strip(', \n\r')
```
mannickutd: maybe \t as well if you want

quis: Good point, added in 772639f

PR #15, opened 2016-03-24
Make phone validation more flexible

quis:
## Allow phone numbers with no leading 0

When exporting from some spreadsheet software (OS X numbers, LibreOffice) phone numbers with a leading zero get converted to numbers, so 07123456700 becomes 7123456700.

This commit adds numbers without leading zeros to the range of acceptable phone numbers.

image

## Add tests for formatting of phone number

This function was untested. This commit adds a test that checks each of the valid phone numbers gets outputted in the same format.

## Add single method to format phone numbers

Currently using the formatting phone numbers method requires doing `format_phone_number(validate_phone_number(123…))`

This is clunky, this commit adds an alias so that you can do `validate_and_format_phone_number(123…)` instead


mannickutd:
:+1:

PR #16, opened 2016-04-04
Add recipient whitelist

quis:
```python
recipients = RecipientCSV(
  """
    phone number
    07700900111
  """,
  template_type='sms',
  whitelist=['07700900999']
)
list(recipients.rows_with_errors)
>>> [1]
```

***

Add option to pass a whitelist of recipients

> We don't want to encourage people to upload real data into the system
> while they're in trial mode, they should be using their own team members
> addresses. So rather than processing and ignoring, we should reject and
> let them know that the numbers/addresses they're uploading aren't
> allowed to be sent to.

This commit adds an extra check when processing the rows in a CSV to annotate the recipients field with an error if an (optional) whitelist is provided and the recipient isn’t in the whitelist.

https://www.pivotaltracker.com/story/show/116767967

## Make whitelist work for any recipient format

Phone numbers and email addresses uploaded to the system could come in any format/or mixture of cases.

The whitelist validation should ignore differences in format and/or case when checking whether or not a recipient is allowed.

(as a separate commit because it changes quite a bit of code)

## 	Refactor whitelist validation so API can use it

The API does the same whitelist checks here:
https://github.com/alphagov/notifications-api/blob/514d490d2f6146ff82c8b5aefc4b081e8c03c396/app/notifications/rest.py#L362

This commit factors out the code that checks if a recipient is in a whitelist so that it can be used by the API and by the `RecipientCSV` class.


utils/recipients.py
```diff
@@ -220,8 +227,25 @@ def validate_email_address(email_address):
...
     return {
         'email': validate_email_address,
         'sms': validate_phone_number
     }[template_type](recipient)
+
+
+def format_recipient(recipient):
+    try:
+        return validate_and_format_phone_number(recipient)
+    except InvalidPhoneError:
```
servingUpAces: If the phone number does not have enough digits, it will then validate as an email address, then get an InvalidEmailError. Is that right?

quis: Almost. The `format_email_address` method doesn’t do any validation so will never raise an `InvalidEmailError`. So there’s no need to catch it.

quis: Thinking about it more, I agree that it wasn’t that clear. Rewritten here: https://github.com/alphagov/notifications-utils/pull/16/files#diff-e3ab47367cb27430e6a12689b0b0a0f5R247

servingUpAces:
:+1:

PR #17, opened 2016-04-05
Fix bug where whitelist was being ignored

quis:
The `RecipientCSV` class was accepting an iterator for the whitelist, eg
https://github.com/alphagov/notifications-admin/blob/b912fa3f6686775d8fad90d5902406ada052418f/app/main/views/send.py#L230

This meant that once it had read the whitelist once, it was now empty.
Since an empty whitelist is treated as no whitelist at all, this meant
that all subsequent phone numbers or email addresses were let through.

This commit makes sure that the whitelist is always cast to a list when
the class is initialized, and the iterator is thrown away.

Minor version bump because fixing a bug in a non-breaking way.

***

Tested locally:

image


servingUpAces:
:+1:

PR #18, opened 2016-04-06
Add method to get phone number in human-readable format

quis:
Turns out that if you put spaces in a phone number then spreadsheet applications won’t try to interpret it as an ‘actual’ number.

It is also more human-readable when spaced out, although this may have the downside of coming back in a different format to the one in which the user entered it.

So, TL;DR, this commit adds a method which can take a UK mobile number in any format any return it in the format `07700 900 762`.

Minor version bump; new feature introduced in a non-breaking way.


mannickutd:
:+1:

PR #20, opened 2016-04-12
Template subject now replaces placeholders the same as content. All t…

mannickutd:
…ests passing.


notifications_utils/template.py
```diff
@@ -48,9 +48,11 @@ def formatted_as_markup(self):
...
     @property
     def placeholders(self):
-        return OrderedSet(re.findall(
-            Template.placeholder_pattern, self.content
-        ))
+        oset = OrderedSet()
+        if self.subject:
+            oset = OrderedSet(re.findall(Template.placeholder_pattern, self.subject))
+        oset.update(OrderedSet(re.findall(Template.placeholder_pattern, self.content)))
+        return oset
```
quis: You could just look for all the placeholders in the template and subject combined, eg:
`self.content + (self.subject or '')`

mannickutd: Agree fixed.

PR #24, opened 2016-04-14
Try to prevent GOV.UK becoming a link

quis:
When we talk about GOV.UK things, we don’t necessarily want to link to
the www.gov.uk homepage.

Email clients disagree.

This commit replaces all instances of GOV.UK in an email with
```
guv dot unicodezerowidthspace you kay
```

…which seems to stop email clients treating it as a link.

It ignores genuine links, like gov.uk/register-to-vote or wwww.gov.uk?search=blah

***

image

***

https://www.pivotaltracker.com/story/show/115929585


servingUpAces:
:+1:

PR #26, opened 2016-04-27
Work out whether a message will be too long for each row of a CSV

quis:
Previously a row was a dictionary, keyed by the columns. It also had an extra key, `index`, which said which row number it was in the file.

For example:
```python
{
    'name': 'Chris',
    'day': 'Friday',
    'index': 0
}
```

If your CSV file looked like this, however:
```csv
phone number, index
07815838437, 27364
```
…you would lose the data in the index column.

This will only get worse if we add additional properties to the row which might clash with the name of the columns. So this commit changes the structure of the row dict to contain the columns, ie:
```python
{
    'columns': {
        'name': 'Chris',
        'day': 'Friday',
    },
    'index': 0
}
```

It then adds an extra property which specifies whether the message that would result from processing that row would be too long:
```python
{
    'columns': {
        'name': 'Chris',
        'day': 'Friday',
    },
    'index': 0,
    'message_too_long': True
}
```

This is considered to be an error on the row.

***

More detail of how this works in the individual commits…


mannickutd:
:+1:

quis:
@mannickutd 🎉 https://github.com/alphagov/notifications-utils/releases/tag/5.1.0

PR #27, opened 2016-04-28
Ignore case and spacing in CSV file column headers

quis:
Users often won’t have control over the data that they are uploading. But we mandate, for example that their recipient column is `email address` or `phone number` with that exact case. This is also a problem
if they put a trailing space in one of their placeholders, or use Title Case when their file has UPPERCASE columns, etc.

_However_, we don’t want to change how we show column headers on the front end (should match what the user uploaded).

So this commit replaces the `dict` of columns that we return for each row with an object that:
- behaves like a `dict`, ie you can do `.get()` on it and, subscript it
- ignores case and spacing, so ‘phone number’, ‘PHONENUMBER’ and ‘  Phone_number  ’ will all give the same value


notifications_utils/recipients.py
```diff
@@ -158,7 +176,17 @@ def column_headers(self):
 
     @property
     def missing_column_headers(self):
-        return set([self.recipient_column_header] + list(self.placeholders)) - set(self.column_headers)
+        required = {
+            Columns.make_key(key): key
+            for key in set([self.recipient_column_header] + self.placeholders)
+        }
+        return set(
```
mannickutd: Not sure you need the last set creator here

notifications_utils/recipients.py
```diff
@@ -169,15 +197,15 @@ def column_headers_with_placeholders_highlighted(self):
...
             try:
                 validate_recipient(value, self.template_type)
             except (InvalidEmailError, InvalidPhoneError) as error:
                 return str(error)
             if list(self.whitelist) and not allowed_to_send_to(value, self.whitelist):
                 return 'You can’t send to this {}'.format(self.recipient_column_header)
 
-        if key not in self.placeholders:
+        if key not in self.placeholders_as_column_keys:
             return None
```
mannickutd: None is not needed on the end of the return

mannickutd:
Apart from those small things :+1:

PR #30, opened 2016-05-11
Make email template more GOV.UK-y

quis:
This commit makes a number of changes intended to increase users’ trust in our email templates:
- making the crown in the header visually match the website as closely as possible
- adding a grey footer, with the crest image

image


servingUpAces:
👍

PR #32, opened 2016-05-16
Handle CSV files with blank column headers 

quis:
Fixes https://www.pivotaltracker.com/story/show/119527051

```csv
phone number,name,cat
7739468050,Pete,Socks,01/01/17
7739468050,Chris,Pickles,01/01/17
```

Should have a ‘date’ heading, but if it doesn't it throws a 500 rather than
telling you that.

CSV `DictReader` returns a dictionary for each row of a CSV, eg:
```python
{'name': 'Pete', 'cat': 'Socks'}
```

If a column doesn’t have a header, then `DictReader` appends its value to a
list inside the dictionary, keyed by `None`, eg
```python
{'name': 'Pete', 'cat': 'Socks', None: ['01/01/17', …]}
```

This was causing problems because we were expecting all dictionary keys
to be strings. This commit:
- adds a bunch of test data where there are blank column headings
- handles the errors that this causes

So the user will see a ‘you need a column called date’ error message, rather
than a 500.


ashimali:
👍

PR #33, opened 2016-05-17
Ensure email header is full-width

quis:
Android email clients are reluctant to make the black ‘GOV.UK’ header
bar full-width. This commit adds some inline CSS to persuade them.

As suggested by: http://stackoverflow.com/questions/25630139/100-width-tables-dont-work-in-gmail-android

Also makes some fixes for the email header not being full-width on Outlook.

And removes the footer because it’s causing more problems than it’s solving
for now.


quis:
Demo’d the EOA results to Pete.

PR #34, opened 2016-05-17
Add an `allowed_to_send_to` property to a CSV

quis:
It is useful to whether, as a whole, a user is allowed to send to all the recipients in a CSV file. It will help us give better error messages on the front end, without having to look through each row in the admin app.

This commit adds just such a property.

Minor version bump.

***

https://www.pivotaltracker.com/story/show/118919845


mannickutd:
:+1: Code looks fine, will need to update the version number.

PR #35, opened 2016-05-18
Property added with test.

mannickutd:
Update version.


notifications_utils/recipients.py
```diff
@@ -158,6 +158,15 @@ def personalisation(self):
             yield self._get_personalisation_from_row(row)
 
     @property
+    def annotated_recipients_and_personalisation(self):
```
quis: Can this be called `enumerated_recipients_and_personalisation` instead? thnx.

quis:
👍

PR #37, opened 2016-05-31
Add utils for showing changes to a template

quis:
_Minor version bump, adds new features in a non-breaking way_

## Add a class for comparing templates

When someone edits a template, we want to know whether this changes the data that needs to be provided (via spreadsheet or API).

This commit adds a class which, when given two templates, can determine:
- if they have different placeholders
- if the latter template adds any placeholders…
- …or removes any placeholders

## Add compare method to the `Template` class

This commit adds a method to the `Template` class which allows it to compare itself to another (newer) template.

It does this by returning an instance of the `TemplateChange` class.


mannickutd:
:+1:

PR #38, opened 2016-06-01
Remove ‘you can’t send to this phone number’ error

quis:
When you try to send to recipients that aren’t in the whitelist, we give you an error for each of these recipients. This is not helpful, because the problem isn’t less with any specific recipient, and more with the idea that there is a whitelist (because Trial Mode) that you probably didn’t know about.

_However_, sending to non-whitelisted recipients should still be an error

This commit adds trying to send to recipients which aren’t in the whitelist to the set of things which cause the `.has_errors` property to return `True`.

This is done explicitly, because the rows with these recipients won’t appear in `.rows_with_errors` any more.

## _Also, improvements to `.has_errors`_

Detailed in the commits, but:
- Increase test coverage
- Make it run faster

## Major version bump

Includes breaking changes to the return values of:
- `rows_with_errors`
- `has_errors`
- `annotated_rows`
- `annotated_rows_with_errors`

…when you try to send to recipients that aren’t in the whitelist.


servingUpAces:
👍

PR #40, opened 2016-06-02
Preserve brackets when highlighting placeholders

quis:
It’s confusing to sometimes show placeholders with the double brackets as you typed them, and sometimes hide them. Especially when users actually call them ‘double brackets’ sometimes.

This commit changes the template class to always return the placeholders surrounded by double brackets, except when they’ve been replaced by some data.


mannickutd:
:+1:

PR #41, opened 2016-06-02
Make template change case and space insensitive

quis:
Related to @leohemsted’s comment on https://github.com/alphagov/notifications-admin/pull/631#issuecomment-223238872

***

When you make changes to a template it doesn’t actually matter if you change the case or whitespace in a placeholder, eg changing (( FIRSTNAME )) to ((first_name)) is not a breaking change.

This commit changes the `TemplateChange` class to only recognise ‘meaningful’ changes to placeholders, ie changes that modify more than just case or spacing.

It uses the same code for dealing with case/space insensitivity as we do for looking at the column headers in CSV files to ensure consistency.


leohemsted:
👍

PR #42, opened 2016-06-06
Make links out of URLs in HTML emails

quis:
Some email clients make URLs in emails clickable, some don’t. We always want URLs to be clickable.

This commit adds a fairly noddy regex which just looks for anything starting with `http`/`https` followed by a series of non-spaces.


leohemsted:
👍

PR #43, opened 2016-06-07
Escape quotes in URLs

quis:
OWASP says:
> Except for alphanumeric characters, escape all characters with ASCII values less than 256 with the %HH escaping format.

– [https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet#RULE_.235](https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet#RULE_.235_-_URL_Escape_Before_Inserting_Untrusted_Data_into_HTML_URL_Parameter_Values)

I don’t think we need to be that strict because:
- we clean any HTML we find before attempting to make links
- we’re looking at the whole URL, not just the querystring parameters, and we don’t want to encode the slashes in `http://`, for example

But we _do_ need to escape quotes, so that extra attributes can’t be snuck into our links.

We need to un-url-encode the string first, to prevent double-encoding any percent symbols that themselves are encoding a character.


leohemsted:
👍

PR #44, opened 2016-06-30
Add breathing room around GOV.UK logo in emails

quis:
## After

***

On small screens, the GOV.UK logo was sitting flush against the edge of its black containing box. This doesn’t look good.

This commit makes sure there’s always at least 10px of space between the edge of the logo and the edge of the containing box. It does the same for the content of the message so that it stays aligned with the logo.

This doesn’t change how the email displays on larger screens.


servingUpAces:
👍

PR #45, opened 2016-07-04
Extract HTML email template and make it more flexible

quis:
Functional changes:
- HTML email template is no longer in code
- it can take some arguments, and in the future will be able to take more arguments (eg branding)

Aesthetic changes:
- adds the blue bar to the GOV.UK banner

Refactoring changes:
- d923aa5b578b60db3fe1972f37bd7d5d428547c9 — this moves quite a lot of stuff around, so it’ll be easiest to review this commit separately

***


leohemsted:
👍

PR #47, opened 2016-07-06
Add renderers for different message types

quis:
# TL;DR

Let’s us do stuff like:

```python
sms = Template(
  template_model.__dict__,
  values={} if not notification.personalisation else notification.personalisation,
  renderer=SMSMessage(prefix=service.name)
)

html_email = Template(
    template_dict,
    values=notification.personalisation,
    renderer=HTMLEmail()
)

plain_text_email = Template(
    template_dict,
    values=notification.personalisation,
    renderer=PlainTextEmail()
)
```

# Complete changes

## Rename `html_email.py` to `renderers.py`

It’s going to be used for more than just HTML emails.

## Add renderers for plain text email, text messages, etc

We now have a class to render a HTML email.

This commit adds classes with the same interface to render plain text emails, text messages, and email and text previews (for use in the admin app).

By default the `Template` class will return an email or text message preview when `.replaced` or `.formatted` are called.

Different renderers can be specified by passing them to the `Template` class when it’s initialised.

## Drop `as_html_email` property and remove `nl2br`  …

This is handled by the renderers now.

Clients should:
- pass the `HTMLEmail` renderer when initialising the template
- use the `.replaced` property

`nl2br` is only done inside the renderers now

## Add a class for chaining string operations

Because nesting functions, eg `a(b(c('d')))` is really ugly unless you really like Lisp and is hard to read because you have to go right-to-left to work out what’s going on.

This commit adds a class to make this kind of code easier to read:
```python
Take('d').then(c).then(b).then(a).as_string
```

## Bump version to 8.0.0

Drops a few properties from the template class, renames the `html_email` module, so a breaking change.


leohemsted:
👍

PR #48, opened 2016-07-07
Speed up processing of CSV files

quis:
# TL;DR

Makes processing a 100,000 row file at least 6× faster.

***

## Cache column key generation for speedup

Treat columns as case and space insensitive means generating a key for each one. Doing some rudimentary profile revealed that we were doing this operation a (not literally) inordinate number of times:

For a 100,000 row file:
```
  ncalls    tottime  percall cumtime   percall filename:lineno(function)
 3600006    1.956    0.000   19.564    0.000 columns.py:30(make_key)
 3600006    5.610    0.000   17.608    0.000 {method 'join' of 'str' objects}
  300000    1.073    0.000   16.557    0.000 recipients.py:133(<dictcomp>)
34000054    9.209    0.000   11.998    0.000 columns.py:35(<genexpr>)
```

This commit caches calls to this function so when it is called with the same argument (which it nearly always is) we don’t have to do the same intensive string manipulation.

**This cuts the time for processing a 100,000 row file from 32s to 10s.**

## Only annotate rows once

Most of our processing of CSV files is based on the `.annotated_rows` property. `.annotated_rows` takes the raw CSV data and adds some useful metadata, eg does a row have errors.

This is quite expensive, and we were doing it _loads_ of times.

This commit changes the class to only annotate the rows once, when the class is initialised.

This is a tradeoff, because it’s going to consume more memory, but **it cuts processing a 100,000 row file from about 10s to 5s.**

***

Now every call is within an order of magnitude of the number of rows in the file:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    5.508    5.508 {built-in method builtins.exec}
        1    0.000    0.000    5.508    5.508 <string>:1(<module>)
        1    0.000    0.000    5.508    5.508 pluggy.py:586(execute)
        1    0.133    0.133    5.508    5.508 python.py:257(pytest_pyfunc_call)
        1    0.053    0.053    5.375    5.375 test_recipient_csv.py:140(test_big_list)
        1    0.024    0.024    3.500    3.500 recipients.py:25(__init__)
   100001    0.348    0.000    3.476    0.000 recipients.py:128(get_annotated_rows)
   200002    0.286    0.000    2.364    0.000 recipients.py:93(rows)
   200002    1.339    0.000    1.708    0.000 csv.py:106(__next__)
   100000    0.314    0.000    1.608    0.000 recipients.py:133(<dictcomp>)
   200000    0.278    0.000    1.127    0.000 recipients.py:226(_get_error_for_field)
   300000    0.284    0.000    0.562    0.000 columns.py:6(__init__)
   100000    0.089    0.000    0.518    0.000 recipients.py:324(validate_recipient)
   100000    0.063    0.000    0.429    0.000 recipients.py:310(validate_email_address)
   300001    0.227    0.000    0.419    0.000 recipients.py:66(placeholders_as_column_keys)
        1    0.000    0.000    0.382    0.382 recipients.py:74(has_errors)
        1    0.022    0.022    0.373    0.373 recipients.py:105(rows_with_missing_data)
   100000    0.064    0.000    0.366    0.000 re.py:160(match)
   100001    0.137    0.000    0.351    0.000 recipients.py:108(<genexpr>)
      101    0.030    0.000    0.319    0.003 recipients.py:157(initial_annotated_rows_with_errors)
```


leohemsted:
not sure if it'll make a big difference now that it's not being called as much, but not splitting and rejoining has a factor of 10 improvement

```
(api) gds0260:dev leohemsted$ python -m timeit "''.join(c.lower() for c in 'PhOne_number -' if c not in ' _-')"
100000 loops, best of 3: 3.58 usec per loop
(api) gds0260:dev leohemsted$ python -m timeit "'PhOne_number -'.lower().replace(' _-', '')"
1000000 loops, best of 3: 0.364 usec per loop
```

quis:
@leohemsted looks cleaner like that anyway. Changed in 50f0c3b.

leohemsted:
👍

PR #49, opened 2016-07-07
Add a ‘pass through’ renderer

quis:
In the API send email/sms endpoint we need to return the contents of a notification, without doing any transformation of it.

This commit adds a renderer to do that, rather than using a `lambda` in the API code itself.


leohemsted:
👍

PR #50, opened 2016-07-11
Add markdown-style formatting for emails

quis:
Authors of HTML emails can now add:

```
# One level of heading

* lists
* with
* multiple items

^ and callout boxes
```

Which look like:

image


notifications_utils/renderers.py
```diff
@@ -3,6 +3,15 @@
 from os import path
```
leohemsted: think re and urllib aren't needed in this file anymore

quis: @leohemsted Good point, removed them and squashed into 4e13a53.

leohemsted:
👍

PR #51, opened 2016-07-12
Add a new, conditional, type of placeholder

quis:
#### Users who send emails want to be able to insert chunks of text for some recipients but not others.

What we don’t want:
- users passing in these chunks of text as the values of placeholders
- users replicating their business logic in our templates, eg `(if ((age)) < 18)` would be a really bad thing

What these chunks of text aren’t:
- ‘snippets’, ie text that is shared between multiple templates
- data about the recipient—the data is whether or not the recipient’s application has been granted, whether they need to supply an extra document, etc.—they are _content_, so they belong in the template.

---

This pull request defines a syntax to display a chunk of text based on the truthy/falsey value of their data.

**Example template**

```
Hello ((name)). ((granted??Your application has been granted)).
```

**Data**

name | granted
---------|------------------
Chris    | yes
Philip   | no

**Result**

```
Hello Chris. Your application has been granted.
```

```
Hello Philip.
```

***

In the admin app, this will look like:

image


leohemsted:
👍

PR #52, opened 2016-07-12
Strip HTML from placeholder values

quis:
If we are going to preview HTML emails then we need to make sure that no HTML can be injected into the email beforehand.

This means

1. Stripping HTML from the template (which we did in https://github.com/alphagov/notifications-api/pull/135)
2. Stripping HTML from the values of the placeholders (which we weren’t doing)

This commit makes sure that placeholders are never replaced with HTML, in emails (subject and body) or text messages.

---

This is not _yet_ a vulnerability, but will be when we turn on previewing of templates in the front end.

---

Patch version bump, fixing this bug does not make breaking changes.


notifications_utils/template.py
```diff
@@ -1,5 +1,6 @@
 import re
 import math
+import bleach
```
leohemsted: should we now be using bleach instead of urllib.parse?

https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/formatters.py#L25

leohemsted:
👍

notifications_utils/template.py
quis: Discussed IRL, bleach only strips out complete HTML, it doesn’t handle breaking out of attributes, eg:
```
https://example.com"onclick="alert('hi')
```

PR #54, opened 2016-07-15
Use a real Markdown parser

quis:
### Reviewers: the substantive changes are all in df9f167

***

The regex-based solution for transforming our flavour of Markdown into HTML was never going to work long term. This commit replaces it with a parser called [mistune](https://pypi.python.org/pypi/mistune) which:

- is easily customisable
- does proper lists with a containing `<ul>`
- and I wish I’d found before

The main changes I’ve made are:

- changing the blockquote character from `>` to `^`
- adding inline styles to paragraphs, headings, lists, etc so that they still work in HTML emails
- disabling all the other Markdown features like images, multiple heading levels, etc

This does give us extra features that we didn’t have before and are probably OK:

- ordered lists
- nesting (eg put a list inside a blockquote)
- multiline blockquotes


leohemsted:
👍

PR #55, opened 2016-07-19
Add sub-branding for email templates

quis:
If a user’s experience of a service includes the department/agency/etc’s branding then the emails they receive should also include this branding.

This commit adds parameters to the email template which allows any app consuming it to specify branding. There are three parts to branding:

- a colour
- a logo image
- a name

Only the logo is required. Name and colour will be shown if they are provided.

This means that we still need to figure out:

- how to model/manage brands in the API
- where to serve the logos from
- how (if at all) services can choose their branding (or how we choose it for them)

image

image


leohemsted:
👍

PR #56, opened 2016-07-22
add optional remaining_messages option to recipients

leohemsted:
represents the number of messages we want to send - if the CSV has more rows, then has_errors will be set to true

blocks https://github.com/alphagov/notifications-admin/pull/809


notifications_utils/recipients.py
```diff
@@ -78,7 +82,8 @@ def has_errors(self):
             (not self.allowed_to_send_to) or
             self.rows_with_missing_data or
             self.rows_with_bad_recipients or
-            self.rows_with_message_too_long
+            self.rows_with_message_too_long or
+            self.more_rows_than_can_send
```
quis: This should probably go before `rows_with_missing_data` because it’ll be faster.

quis:
👍 Other than the one comment.

PR #57, opened 2016-07-22
Mark tables as presentational in email template

quis:
We make extensive use of tables in our HTML email because email client support for CSS layout is not good.

This could have a side effect for screen reader users: content being read out as if it were data. 

Browsers have heuristics to work out whether a table is being used for data or layout (eg by seeing if it has any column headers). That means this isn’t currently a problem, at least in the testing I’ve done (voiceover with Chrome on Mac).

However, it can’t hurt to make it explicit that the table is not data, and the way to do that is with the `role` attribute.


leohemsted:
👍

PR #58, opened 2016-07-28
Fix linkify formatter messing with newlines

quis:
The linkify formatter was stripping trailing newlines if they immediately followed a URL.


quis:
Closing in favour of https://github.com/alphagov/notifications-utils/pull/59

PR #59, opened 2016-07-29
Fix markdown email oddities

quis:
## Fix linkify formatter messing with newlines

The linkify formatter was stripping trailing newlines if they immediately followed a URL.

## Make Markdown take notice of single newlines

Ordinarily Markdown treats single newlines as wrapping, not soft returns.

So something like this:
> Thanks,
> Government

Comes out as:
> Thanks, Government

We want Markdown to treat newlines as newlines. The proper syntax for doing this is to add two spaces `  ` at the end of the line, which is super obscure and unintuitive. So rather than try to document this weird syntax, let’s just do the conversion for people.

## Explicitly set text colour in HTML emails

We were setting text colour on the whole body of the email, but clients like GMail seemed to override this with purple when sending yourself multiple emails (like when you forward an email).

Fixing this needs a brute force approach of setting the colour on each paragraph, heading, etc explicitly.

## Allow long links to wrap onto two lines

Links can get very long, and in some email clients (Apple Mail) this causes them to expand the width of the email. They should just wrap onto two or more lines instead.

We already do this when preview emails in the admin app: https://github.com/alphagov/notifications-admin/commit/6909146feea23324e0a3ed6f9ef44013f57dc29e


quis:
Long links still broken in Apple mail:

image

quis:
## Specify max width on main content column

This forces anything that is set to `word-wrap: break-word` to wrap at 560px, rather than expand its container.

Fixes broken long links in Apple Mail.

minglis:
👍

PR #61, opened 2016-08-05
Added statsd logging into the after_request handler

minglis:
- Usage:

in Flask application pass an initialized statsd client into the logger.
only the API does this now.

then setup the data required for the after_request hook

 @app.before_request
    def record_start_time():
        g.start = monotonic()
        g.endpoint = request.endpoint

this records the start time and the endpoint (module and method) that the request will invoke.

The logger will then use these to create statsd metrics to record the event (count) and the time taken


leohemsted:
👍

PR #62, opened 2016-08-08
render brand_colour correctly if user passes in leading # for hex

leohemsted:
previously would render '##f00' for example


quis:
👍

PR #63, opened 2016-08-08
Include email-template dir in spacer image path

quis:
Makes it clear that it’s not part of the app.

Depends on:
- [x] alphagov/notifications-admin#851


leohemsted:
👍

PR #66, opened 2016-08-23
Don’t prefix text messages if sender name is set

quis:
> We have an alphanumeric sender name for texts. We also inject the service name. This can be useful but it can be a bit repetitive.
>
> So let's make this more flexible by only enforcing the service name at the front of the text if there's no sender name. It's then possible for services to add it anyway to their templates.
>
> I think there is a pattern to emerge here, but not certain what it is yet, so let's open it up a bit and see what people want to go with.

https://www.pivotaltracker.com/story/show/126047943

### Make template pass sender through

There are quite a few places in the admin app where we don’t call the `SMSMessage` or `SMSPreview` renderers directly, so there’s no way to pass them the SMS sender name directly.

Need the template class to pass it through, like it does with the prefix.

### Don’t make prefix a property of `Template`

It’s just passing it through, and having it as a property implies that it will respect reassignment, which it won’t.


leohemsted:
👍

leohemsted:
is this a breaking change tho?

quis:
Breaking change explained in 2280a90318a799b0ed4a54d61dce986c137bb01a. Depends whether you think that removing:
- a ‘private’ method
- that I don’t think any of the apps are using
really constitutes a breaking change?

PR #67, opened 2016-09-05
Add blank `alt` attribute to presentational images

quis:
> Simply leaving the alt attribute off the img tag isn’t good enough, in this case the screen reader will read the full URL of the image, not good. So instead use alt="" to let the screen reader know this is intended to be blank.

http://blog.rebelmail.com/accessibility-in-email-part-ii/


cfq:
👍

PR #68, opened 2016-09-12
Do some more optimisation of the CSV processing

quis:
This makes about a 30% improvement if you upload a big file when you’re in trial mode, and 8% for other cases.

## Make big list test more rigorous

Adding a whitelist argument to the big list test makes it do more work, because it has to check whether or not each recipient is in the whitelist.

_10 loops, best of 1: 7.71 sec per loop_

## Add caching to the format recipient method

We call format recipient with the same recipient a lot. This is because we use it to normalize recipients, so for each message in the list we’re also formatting every message in the whitelist.

Putting a cache in front of this makes it faster, because the whitelist has a small number of recipients that don’t change.

_10 loops, best of 1: 5.38 sec per loop_

## Only compute placeholders as column keys once

Computing `placeholders_as_column_keys` again and again is inefficient. The only time it will change is if the placeholders change, so let’s compute it once in the `placeholder` property’s setter.

_10 loops, best of 1: 5.13 sec per loop_

## Add a setter method for template type

This lets us compute the `recipient_column_header` property only once, rather than derive it from the template type every time.

_10 loops, best of 1: 4.97 sec per loop_


minglis:
👍

PR #69, opened 2016-09-14
Fix HTML lists in Microsoft Outlook

quis:
*Click screenshots to enlarge*

# Before 

image

# After

image

---

There are still some spacing inconsistencies but fixing them will need more creativity 😞. Let’s get this fixed first.


servingUpAces:
👍

PR #70, opened 2016-10-05
Email regex

leohemsted:
borrow regex from wtforms - should be a bit more robust, especially regarding
trailing punctuation. also should make us a bit more consistent between forms
(sign-ups, invites) and sending notifications via csv/api

plz double check test cases my local python is broken to fuck and i couldn't run locally


imdadahad:
Ran tests locally and they pass

imdadahad:
Yeah that's fair. 👍

PR #71, opened 2016-10-07
make email regex more restrictive with " and ; characters

leohemsted:
don't allow double quotes or semicolons anywhere in the email - while they're
technically allowed under some circumstances by RFC6530, chances are much
higher that the input data has accidentally been modified.

also tweak the tests to reduce logic and separate good from bad emails, and
rename some test cases to make it more apparent what is wrong


quis:
All of the technical failures this morning were email addresses the mistakenly started with a `"`. So this would have caught it.

👍  but need to be careful when you send it out so that it doesn’t interfere with any scheduled jobs.

quis:
Also needs to catch `..`

tests/test_recipient_validation.py
```diff
@@ -88,6 +86,8 @@
     'email+middle"-quotes@domain.com',
     '"quoted-local-part"@domain.com',
     '"quoted@domain.com"',
+    'lots-of-dots@domain..gov..uk',
```
quis: Can you add a test case for `lots..of..dots@domain.com` with a comment to say that it’s technically valid but that we’re not allowing it?

quis:
BTW, this should be rolled out to admin first, then wait 24 hours before rolling out to API (to let any scheduled jobs flush through).

PR #74, opened 2016-10-13
make email regex stricter

leohemsted:
don't allow spaces anywhere, or underscores in the domain (fine in local part)

jesus hopefully this is good enough


quis:
🙏

quis:
👍

quis:
@leohemsted added an extra test for `firstname.o'lastname@example.com` in 33f7bc0.

PR #75, opened 2016-10-19
Fix multiple newlines in emails

quis:
We do this so that the Markdown parser interprets them as single newlines, without the author of the template having to know this trick (see 11a42ba for full explanation).

It wasn’t working properly:
- when there was a series of single lines of text, like in a street address (it was only adding spaces after every other line)
- when the author added the double spaces, or any other number of spaces at the end of a line themselves

This commit makes the regex a bit smarter by using a lookahead to determine if it’s a single, not double newline.

> `(?=...)`
> Matches if `...` matches next, but doesn’t consume any of the string. This is called a lookahead assertion. For example, `Isaac (?=Asimov)` will match 'Isaac ' only if it’s followed by 'Asimov'.

https://docs.python.org/3/library/re.html#regular-expression-syntax

It also removes trailing whitespace on single newlines before normalizing it at 2 spaces.
---

Thanks to @timblair for spotting this bug.


notifications_utils/formatters.py
```diff
@@ -12,7 +12,7 @@
 )
 
 single_newlines = re.compile(
-    r'^(.+)\n([^\s])',
+    r'^(.+)\n(?=[^\n])',
```
leohemsted: 😷

PR #76, opened 2016-10-27
Email regex changes

leohemsted:
# ✉️  🔥

notable changes:
- long TLDs. previously we checked for a TLD of length 2-10 - turns out `.enterprise` is a valid TLD.
- leading/trailing whitespace. we strip it now
- unicode TLDs. they're a thing. we allow them now


notifications_utils/recipients.py
```diff
@@ -341,13 +339,42 @@ def validate_and_format_phone_number(number, human_readable=False):
...
+    # if the part after the last . is not a valid TLD then bail out
+    if not tld_part.match(parts[-1]):
+        raise InvalidEmailError
+
     return email_address
 
 
 def format_email_address(email_address):
-    return email_address.strip().lower()
+    return email_address.lower()
```
imdadahad: Any reason why we no longer want to strip, seems super strict to say an email like: ' someone@mail.com' is invalid only because of the leading whitespace?

leohemsted: cos we're now doing it as part of the validation process, which this happens after - so the email is already stripped when this function is called

imdadahad: Cool

imdadahad:
Looks good 👍

PR #77, opened 2016-11-01
Add method to get `Columns` instance as a `dict`

quis:
Factored out from [alphagov/notifications-api/blob/548525/app/celery/tasks.py#L75-L78](https://github.com/alphagov/notifications-api/blob/5485252b85d2bfdeaacbb76ca2304a729f522d5e/app/celery/tasks.py#L75-L78)
for reusability.


imdadahad:
Looks good 👍

PR #79, opened 2016-11-04
Refactor renderer into a getter/setter

quis:
So there’s less logic in the `Template`’s `__init__` method.


quis:
Martyn 👍 ’d IRL.

PR #83, opened 2016-11-16
Factor out placeholder replacement/highlighting

quis:
We probably want to use this logic outside the `Template` class, eg to replace the ‘to’ address when rendering letters.

This commit makes a class called `Field`, which deals only with the replacement and highlighting of placeholders in an arbitrary string of text.

The `Template` class then, indirectly, uses an instance of `Field` for the `content` and `subject` of a template.

This commit also removes unused methods from the `Template` and `RecipientCSV` classes to make this an easier change.


quis:
Closing in favour of https://github.com/alphagov/notifications-utils/pull/84

PR #84, opened 2016-11-17
Refactor interaction between `Template` and `Renderer` classes

quis:
## Factor out placeholder replacement/highlighting

We want to use this logic outside the `Template` class, eg to replace the ‘to’ address when rendering letters.

This commit makes a class called `Field`, which deals only with the replacement and highlighting of placeholders in an arbitrary string of text.

The `Template` class then, indirectly, uses an instance of `Field` for the content and subject of a template.

This commit also removes unused methods from the `Template` and `RecipientCSV` classes to make this an easier change.

## Let the `LetterRenderer` add an address block to a letter

Significant refactoring so that `Renderer` classes take a `dict` representation of a `Template` instance. 

This means the `LetterRenderer` has all the information it needs to piece together the subject, content and recipient of the letter into one HTML representation of a letter.

## Generate complete template previews from renderers

Currently, the `Renderer` classes only generate previews of the body of the message.

This means that the preview (including showing the recipient, rendering other wrappers, etc) is split between the admin app and the `utils` stuff.

This commit brings all that stuff in from the admin app, so that calling `template.replaced` gives you (for example) a complete preview of an email, with the from address, service name, recipient’s email, etc. all populated.


notifications_utils/template.py
```diff
@@ -90,99 +82,34 @@ def renderer(self, value):
...
-
-    @property
-    def replaced_content_count(self):
-        return len(self.replaced.encode(self.encoding))
+        return Field(self.subject or '').placeholders | Field(self.content).placeholders
 
     @property
     def content_count(self):
-        return len(self.content.encode(self.encoding))
+        print(self.rendered)
```
leohemsted: 🖨 ❌

notifications_utils/template.py
```diff
@@ -46,19 +38,19 @@ def __init__(
...
         self._sms_sender = sms_sender
         self.renderer = renderer
 
-    def __str__(self):
-        return self.replaced if self.values else self.content
-
     def __repr__(self):
         return "{}(\"{}\", {})".format(self.__class__.__name__, self.content, self.values)  # TODO: more real
 
+    def __iter__(self):
```
quis: Fair point. Changed it to `.to_dict()`

notifications_utils/template.py
quis: Squashed out.

PR #85, opened 2016-11-22
Pulled the statsd and redis clients out of the API and into the utils repo

minglis:
Straight forward lift and shift, excepting I bumped the version of pytest to match that in the API. Caused 2 warnings. 1 I've fixed, one 1 can't trace.

1st step to sharing them with the Admin application.


tests/test_this_is_setup_properly.py
```diff
@@ -1,2 +0,0 @@
-def test_for_testing_sake():
-    assert 1 == 1
```
servingUpAces: what is this?

quis:
This modifies a bunch of the Markdown tests. Bad merge/rebase?

tests/test_this_is_setup_properly.py
minglis: Dunno, I removed it. Think is was the original test from the empty repo to prove pytest worked

minglis:
@quis Yeah. I bumped the version of py.test to match what we use elsewhere and that broke class based testing. So I had to change those tests otherwise they'd be ignored.

PR #86, opened 2016-11-28
Add renderers for generating PDF/image previews of letters

quis:
## Add a prototype HTML preview of the letter

This is just some HTML/CSS I threw together to make something that looks a bit like a letter.

It doesn’t represent the work that Sonia/Maria have been doing on the actual letter template.

We’ll generate PDF previews from this HTML.

## Add a renderer to preview a letter

This new renderer takes a template and returns:

- an `` preview of the letter
- wrapped in a link to the PDF preview of the letter

Generating the URLs manually feels a bit dirty. And probably isn’t going to be flexible enough when we want to show a preview of a letter with the placeholders replaced, or a preview of a previous version of a letter. Works for now as a first pass, so should refactor later.


notifications_utils/formatters.py
```diff
@@ -72,6 +72,10 @@ def prepend_postal_address(body, address_field):
     )
```
servingUpAces: prepend_postal_address is only used in tests and unused import.

quis:
@servingUpAces 

- removed `prepend_postal_address` and squashed into 2375269
- added a test for Markdown in letter templates in 498b5c4

PR #87, opened 2016-11-30
Make RecipientCSV understand quoted values

pcraig3:
Setting `csv.QUOTE_NONE` means the string

```01,"Nov, 2016",hello```

will be interpreted as

`01`, `"Nov`, `2016"`, `hello`

rather than as

`01`, `Nov, 2016`, `hello`

Adjusting to `csv.QUOTE_MINIMAL` fixes this problem.
It'll consider delimiters and quotecharacters as special fields,
which it seems like is what we want.

[Docs for csv.QUOTE_MINIMAL](https://docs.python.org/2/library/csv.html#csv.QUOTE_MINIMAL)


quis:
Is this still WIP? Looks fine to me apart from needing a version bump.

I can pick up the admin side of things separately if you let me know the branch it’s on.

pcraig3:
Yeah, it's just WIP as a formal thing until the tests are in the admin app; I wasn't planning any more code changes other then the version.

PR #88, opened 2016-11-30
Add service id into logging

minglis:
Update the after_request logging.
- if present will log the service ID against the API calls made
- if present will use the service ID as a key in the statsd calls - allows filtering by service ID in graphite by outcome , call, method, package etc.


notifications_utils/logging.py
```diff
@@ -16,46 +16,60 @@
 logger = logging.getLogger(__name__)
 
 
+def build_log_line(extra_fields):
+    fields = []
+    if 'service_id' in extra_fields:
+        fields.append(str(extra_fields.get('service_id')))
+    fields += [extra_fields.get('method'), extra_fields.get('url'), str(extra_fields.get('status'))]
```
leohemsted: i'd prefer these to be `[` `]` - otherwise it'll crash when trying to `join` and a KeyError here would be much easier to debug than `TypeError: sequence item 1: expected str instance, NoneType found`

leohemsted: if any of method, url, or status aren't in, you'll get `TypeError: sequence item 1: expected str instance, NoneType found` when it calls `' '.join(fields)`. otherwise there's no point in calling `.get` instead of `extra_fields['method']`

notifications_utils/logging.py
```diff
@@ -16,46 +16,60 @@
...
+    if 'time_taken' in extra_fields:
+        fields.append(extra_fields.get('time_taken'))
+    return ' '.join(fields)
+
+
+def build_statsd_line(extra_fields):
+    fields = []
+    if 'service_id' in extra_fields:
+        fields.append(str(extra_fields.get('service_id')))
+    fields += [extra_fields.get('method'), extra_fields.get('endpoint'), str(extra_fields.get('status'))]
```
leohemsted: same as fn above

notifications_utils/logging.py
leohemsted: in fact this whole function can be de-complexified
```
field = ['service_id', 'method', 'url', status']
return ' '.join(extra_fields[field] for field in fields if field in fields)
```

notifications_utils/version.py
```diff
@@ -1 +1 @@
-__version__ = '10.0.0'
+__version__ = '10.0.1'
```
leohemsted: minor thing (heh) - should probably be 10.1.0 as it's new/changed functionality rather than bug fixes/refactors.

actually, possibly it should be 11.0.0 as it's a breaking change to downstream integrators (all our statsd graphs will need to be rewritten?)

notifications_utils/logging.py
```diff
@@ -16,46 +16,60 @@
...
-                        endpoint=g.endpoint,
-                        status_code=response.status_code),
-                    time_taken
-                )
-
-        if 'time_taken' in extra_fields:
-            current_app.logger.info('{method} {url} {status} {time_taken}', extra=extra_fields)
-        else:
-            current_app.logger.info('{method} {url} {status}', extra=extra_fields)
+                statsd_client.timing(build_statsd_line(extra_fields), time_taken)
```
leohemsted: DRY: you can build_statsd_line just once in a variable above rather than calling it twice

notifications_utils/version.py
minglis: thinking of them as new/additional graphs. We can ship this and not affect graphs. The existing ones can be maintained by starting out the service id?

minglis: Take point about statsd. Nothing else should be affected - it doesn't change signatures?

notifications_utils/logging.py
```diff
@@ -30,7 +31,9 @@ def build_statsd_line(extra_fields):
     fields = []
     if 'service_id' in extra_fields:
         fields.append(str(extra_fields.get('service_id')))
-    fields += [extra_fields.get('method'), extra_fields.get('endpoint'), str(extra_fields.get('status'))]
+    standard_fields = [extra_fields.get('method'), extra_fields.get('endpoint'), extra_fields.get('status')]
+    fields += [str(field) for field in standard_fields if field is not None]
+    print(fields)
```
leohemsted: 🖨

notifications_utils/version.py
minglis: Gone 10.1.0

PR #94, opened 2016-12-01
Unite `Template` and `Renderer` classes

quis:
So this is the work that I wanted to do as part of https://github.com/alphagov/notifications-utils/pull/84 but wasn’t sure whether it was worth the effort. But @leohemsted’s comments on that PR got me thinking again, and I had 2 uninterrupted hours on the train this morning, so here we are.

***

This is a big refactor. It’s addressing three issues which were making the existing code bloated and overly complicated:

1. The interaction between the `Template` and `Renderer` classes was still weird, because an instance of `Renderer` had effectively become a method of a `Template` instance

2. The `Renderer` classes themselves weren’t really proper classes, they were basically just functions that took a template and turned it into a string, but architected in a slightly odd way. There are people who consider a class that only has `__init__` and `__call__` methods an anti-pattern (see [Jack Diedrich’s talk _Stop writing classes_](https://www.youtube.com/watch?v=o9pEzgHorH0)

3. There were methods on the `Template` class (eg `content_count`) which were very specific to one kind of template (eg a text message).

This commit makes three main changes to address these issues:

1. Makes `Template` into an abstract-ish base class, and makes the various `Renderer`s into `Template`s which inherit from the base template class (ie `HTMLEmailTemplate` instead of a renderer called `HTMLEmail`).

2. Moves the functionality of the renderers to `__str__` methods on template instances. This is a better model of what they do (turn templates into strings).

3. Methods that are only relevant to one type of template (eg text messages) now live only on that type of template.

~There is one thing that’s still slightly weird about this. Service-level things on a template (eg the text message sender name) aren’t set through an `__init__` method any more because there’s no appropriate place for one to exist.~

~Instead they are set directly on the instance (eg `template.sms_sender = 'Service name'`). This means doing some slightly funny jiggery pokery around default and required ‘arguments’, which
would be more intuitive with an `__init__`. But I _think_ the tradeoff is worth it for the simpler code elsewhere.~

[`super()` considered super!](https://rhettinger.wordpress.com/2011/05/26/super-considered-super/)

## Also, add subject property back onto certain templates

It’s really ugly having to do this sort of thing all over the API:
```python
subject=str(Field(plain_text_email.subject, notification.personalisation))
```

But it doesn’t really belong in the base `Template` class because not all templates have subjects.

So this commit re-adds subjects (and the highlighting/replacing thereof) to those templates which do have subject lines.

So the API code can just change to:

```python
subject=template.subject
```


notifications_utils/template.py
```diff
@@ -66,40 +73,50 @@ def values(self, value):
...
     @property
-    def sms_fragment_count(self):
-        if self.template_type != 'sms':
-            raise TypeError("The template needs to have a template type of 'sms'")
-        return get_sms_fragment_count(self.content_count)
+    def content_count(self):
+        return len(str(self).encode(self.encoding))
 
     @property
     def content_too_long(self):
```
leohemsted: unused property? it certainly errors

notifications_utils/template.py
```diff
@@ -109,21 +126,198 @@ def content_too_long(self):
...
+
+    def __str__(self):
+        return Markup(self.jinja_template.render({
+            'body': get_html_email_body(
+                self.content, self.values
+            ),
+            'subject': self.subject,
+            'from_name': self.from_name,
+            'from_address': self.from_address,
+            'recipient': Field("((email address))", template.get('values', {}), with_brackets=False),
```
leohemsted: think this should be `self._template`? it's currently crashing when i try and send an email on admin app, so make sure there's a test for this

quis:
Addresses unreachable code in a separate PR: https://github.com/alphagov/notifications-utils/pull/101

notifications_utils/template.py
quis: Yeah, the API was doing this comparison itself. Removed in: 7997fc7d7c30a4f1f26280997404ab73837f50a7

notifications_utils/template.py
quis: `self.values`. I think I’d fixed this already and squashed it out but not pushed it or something ¯\_(ツ)_/¯

PR #99, opened 2016-12-06
Use different statsd identifier for admin api calls

minglis:
Prefix service ID statsd identifiers with service-id
Don't do this for notify-admin calls, to allow them to be extracted from the grafana dashboards easily


minglis:
Review by Chris

PR #100, opened 2016-12-08
Make letter previews reflect what the letter will look like

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-admin/pull/1050

Branched off #94 – compare: https://github.com/alphagov/notifications-utils/compare/refactor-template-renderer...accurate-letter-preview

image

## Parametrize markdown tests to expect different output

Right now, we expect both the Markdown used in letters and the Markdown used in emails to convert to the same HTML. But the Markdown in letters doesn’t need to have all the inline styles, because we’re not working around the limitations of email clients.

This commit takes the first step towards removing those inline styles, which is to add extra parametrization to the tests. This way we can test the same source Markdown, but assert that the output is correct for letters and emails, even when that output differs.

## Remove inline styles from letters HTML

The HTML of our letters is only consumed by the thing that turns it into a PDF. This means that we don’t have to worry about:
- any issues arising from multiple different clients like we do with emails
- any clients which strip out non-inline CSS

So there’s no reason to use inline CSS anymore. This commit removes it
from the parts of the code that convert Markdown into HTML for letters.

## Use default markdown renderers where possible

For letters, the Markdown conversion is a lot closer to the default implementation of Markdown. Where possible we shouldn’t write our own methods to do the conversion.

So this commit:
- removes these custom conversions where they are functionally
  equivalent to the Markdown default
- does not remove the tests for these conversions, because we still want a definition of the output we’re producing (even though this is unit testing an external library)

I expected to be able to remove the custom paragraph convertor, but this isn’t possible because it treats empty lines or strings as paragraphs, which isn’t behaviour we want.

## Add CSS to make a letter look real

This commit adds the CSS to make a letter look like the template that Sonia and Maria have worked on. The image assets will need to be deployed to production on admin for this to work (same as we do for email assets).

Right now it sets the logo to be the HM Government one, which might be the default for departments that don’t want/need their own branding on a letter. Hat tip to @harrytrimble for that idea.


quis:
## Add styling for placeholders in letter templates

This commit highlights placeholders in letter previews the same way that we do for emails and text messages. The implementation is slightly different because WeasyPrint doesn’t support box-shadow (see https://github.com/Kozea/WeasyPrint/pull/149#issuecomment-236052554 for reasons).

Luckily WeasyPrint _does_ support CSS3 multiple backgrounds, so we can mask the curved edges with SVG background images instead. This wouldn’t be a great solution on the web because it you might get a flash of unstyled placeholder as the images loaded. But because the PDF is rendered completely before it’s served it shouldn’t be a problem.

***

image

quis:
## Add date to the letter

image

tests/test_template_types.py
```diff
@@ -105,13 +130,13 @@ def test_markdown_in_templates(template_class):
...
-        '<p style="Margin: 0 0 20px 0; font-size: 19px; line-height: 25px; color: #0B0C0C;">'
-        'jumped over the lazy dog</p>'
-    ) in str(template)
+    ))
+    if template_class == HTMLEmailTemplate:
+        mock_notify_email_markdown.assert_called_once_with(result)
+    elif template_class == LetterPreviewTemplate:
+        mock_notify_letter_preview_markdown.assert_called_once_with(result)
+    else:
+        assert False
```
leohemsted: i really don't like `if` statements in tests - you could pass in the mock string as a parametrise option and then write something like

```
def test_markdown_in_templates(template_class, result, markdown_render):
    with mock.patch(markdown_renderer) as mock:
        str(template) [...]
    mock.assert_called_once_with(result)
```

that saves us from the icky `else: assert False` line too

PR #104, opened 2016-12-20
Switch to cloudfront

leohemsted:
# before:

image

# after:

image

look at dem cloudfront headers!

url changes from `https://www.{NOTIFY_URL}/static/{FILE_PATH}` to `https://static.{NOTIFY_URL}/{FILE_PATH}`


quis:
Need to fix the ones in this file as well:https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/letter_pdf_template.jinja2

(my bad)

quis:
Actually, serving the letter images on localhost is the right thing to be doing. They’re only accessed internally, so it’s probably quicker to request them directly rather than go out to the internet.

quis:
You can see that they’re still working here: 

image

leohemsted:
fair enough @quis - a pdf preview is also significantly lower usage than an image in a template, which was our primary concern - server load rather than render time

PR #105, opened 2016-12-28
Fix email address validation for CSV files

quis:
TL;DR We weren’t catching any bad email addresses in CSV files.

So I’m not sure how this ever worked, tried digging back through `git blame` but even back as far as March 2016 there’s not a change which is the obvious culprit, but it’s some combination of these additions:
- https://github.com/alphagov/notifications-utils/commit/1dc7f0fe84010a6b3c794f59de35a195e235ccba#diff-e3ab47367cb27430e6a12689b0b0a0f5R64
- https://github.com/alphagov/notifications-utils/commit/c7a89ed96e2becf899e99e814350c387779ebcce#diff-e3ab47367cb27430e6a12689b0b0a0f5R115

Anyway, the reason is that `BaseException.message` was deprecated in Python 2.6. Therefore setting `self.message` in the constructor doesn’t do anything.

The code that catches these exceptions just passes through the error message. The code that counts which rows have errors treats an `error` message of `''` on a column equivalent to `None`, ie not an error.

The fix for this is to make sure that the code for setting the default error message on email addresses passes the message through to the `Exception` that it’s subclassing.

I’ve also added a more generic test for invalid addresses at the bottom of a file. This is initially what I thought the problem was, but I think it’s worth keeping this test because:

- it does catch this specific bug
- it catches a bug which could, potentially, be introduced


leohemsted:
just fyi this should actually have been a major change, as it breaks external compatibility (exceptions now don't have `e.message`)

PR #106, opened 2017-01-06
Cloudfoundry logging

leohemsted:
if the config of the app you're setting up logging for has the
`CLOUDFOUNDRY` variable set to True, then set up logs to return
JSON to stdout.

also added tests to ensure we set up logging correctly based on
debug, cloudfoundry, etc

- [x] test it against an actual app


notifications_utils/logging.py
```diff
@@ -100,6 +101,9 @@ def get_handlers(app):
     if app.debug:
         handler = logging.StreamHandler(sys.stderr)
         handlers.append(configure_handler(handler, app, standard_formatter))
+    elif app.config['CLOUDFOUNDRY']:
```
bandesz: Usually a feature switch is better for a certain functionality, therefore I think it's better to use LOGGING_STDOUT_JSON=1, or LOGGING=stdout-json or similar.

leohemsted:
it works locally if i put in a fake VCAP_SERVICES and VCAP_APPLICATION :)

```
(api) [leohemsted:~/dev/api]$ python application.py runserver
{"name": "app", "levelname": "INFO", "message": "Logging configured", "pathname": "/Users/leohemsted/.virtualenvs/api/lib/python3.5/site-packages/notifications_utils/logging.py", "lineno": 93, "time": "2017-01-06T15:43:50", "requestId": "no-request-id", "application": "api", "logType": "application"}
[...]
```

PR #108, opened 2017-01-16
Output a template in the DVLA-specified format

quis:
# TL;DR

Makes this:
```python
str(LetterDVLATemplate(
    {
        "content": (
            'Dear ((name)),\n\n'
            'Thank you for applying to register a lasting power of '
            'attorney (LPA) for property and financial affairs. We '
            'have checked your application and...'
        ),
        'subject': 'Your ((thing)) is due soon',
    },
    {
        'thing': 'application',
        'name': 'Henry Hadlow',
        'address line 1': 'Mr Henry Hadlow',
        'address line 2': 'Managing Director',
        'address line 3': '123 Electric Avenue',
        'address line 4': 'Great Yarmouth',
        'address line 5': 'Norfolk',
        'address line 6': '',
        'postcode': 'NR1 5PQ',
    },
    numeric_id=1,
))
```

Return this:
```
140|001|001||201604290000001|29042016||||||||||||Mr Henry Hadlow|Managing Director|123 Electric Avenue|Great Yarmouth|Norfolk|||NR1 5PQ||||||||Your application is due soon|Dear Henry Hadlow,<cr><cr>Thank you for applying to register a lasting power of attorney (LPA) for property and financial affairs. We have checked your application and...<cr><cr>
```

## Add a letter output template class

We need to send a notification to DVLA in a format they understand.

This is different to, for example, how we format a letter notification to preview it.

The DVLA have specified how they want us to send them each letter to print. The format is a text file with one row per letter, and various fields which are pipe delimited.

Therefore this commit adds a new subclass of `Template` whose `__str__` method outputs one line of the file we send to DVLA.

The specification for what goes in each column is here:
https://github.com/alphagov/notify-ftp-documentation/blob/bd2164066e3a6b38956db7fb06ba6ced9a2cd50f/dvla_documentation/GDS%20Notify%20-%20DVLA%20File%20Format%20v1_2.xlsx

I’ve copied the contents of this spreadsheet into the tests so that it’s a bit easier to understand what each field is.

## Dates

The DVLA spec calls for dates in two places. One is for internal use and one is printed on the letter. They have specified what format the dates should be in.

The date that we print on the letter should be whatever the date is at the time of sending. Their example added an extra day to this, but I think that from the (government) user’s perspective what they see on the printed letter should match what they see on the preview at the time of sending.

## Addresses

The address to send the letter to should be whatever the government user has given to us in their CSV file or API call. The DVLA spec has one field for each line of the address.

They have allowed seven address lines plus postcode. We only collect 6 address lines, plus postcode. Therefore the 7th line of the address is hardcoded to be blank. DVLA will strip out any blank lines in the address before they print the letter.

## Notification IDs

DVLA require us to give them a unique identifier for each letter:

> Unique identifier for each notification, consisting of the current date and a numeric counter that resets each day. Supports a maximum of 10 million notification events per day.
>
> Format:  CCYYMMDDNNNNNNN
>
> Where:   CCYY = Current year
>          MMDD = Current month and year (zero padded)
>          NNNNNNN = Daily counter (zero padded to 7 digits)

The template instance only needs to care about taking whatever ID we want to generate and:
- validating it (just as a safety check)
- outputting it in the right place

Managing the generation of sequential, unique IDs should be done by whatever job is creating the file.

## Markdown

DVLA’s printing software supports some rudimentary markup for styling text.

The specification is here:
https://github.com/alphagov/notify-ftp-documentation/blob/bd2164066e3a6b38956db7fb06ba6ced9a2cd50f/dvla_documentation/Notify%20Markup%20Specification.v0.1.docx

This commit adds the transformation of Markdown into the specified markup.


quis:
# Make subject line render as `<h1>` not `<h2>`

In the content of letters, we let people add a heading with the normal markdown `#` syntax. We convert this to a `<h2>` in the HTML, because the `<h1>` is reserved for the subject.

Unfortunately, the subject was being inserted by prepending it to the content of the letter as markdown, and then letting the Markdown renderer convert it to HTML. But, as described above, the Markdown renderer was converting it to a `<h2>`. This meant that the type size for the subject line was too small.

Fixing this means hard-coding the subject to be a `<h1>` in the template, which is fine because all letters will have a subject, and the user should have no control over its type size.

## Before

screen shot 2017-01-27 at 16 27 43

## After

screen shot 2017-01-27 at 16 27 24

tests/test_template_types.py
```diff
@@ -366,3 +421,478 @@ def test_templates_handle_html(
...
+            'attorney (LPA) for property and financial affairs. We '
+            'have checked your application and...<cr><cr>'
+        ),
+    }
+]
+
+
+@freeze_time("2016-04-29 12:00:00.000000")
+@pytest.mark.parametrize('field', dvla_file_spec)
+@pytest.mark.parametrize('template', [
```
leohemsted: this doesn't need to be in a parametrize

tests/test_template_types.py
```diff
@@ -366,3 +421,478 @@ def test_templates_handle_html(
...
+        numeric_id=1,
+    )
+])
+def test_letter_output_template(field, template):
+    # To debug this test it’s helpful uncomment the following line:
+    # print(field)
+    assert str(template).split('|')[int(field['Field number']) - 1] == field['Example']
+
+
+@pytest.mark.parametrize('template', [
```
leohemsted: also doesnt need to be parametrize

tests/test_template_types.py
```diff
@@ -366,3 +421,478 @@ def test_templates_handle_html(
...
+
+    assert len(str(template).split('|')) == len(dvla_file_spec)
+    assert 'Pipes  pipes  everywhere' in str(template)
+
+
+@pytest.mark.parametrize('id, expected_exception', [
+    (None, 'numeric_id is required'),
+    ('abc123', 'numeric_id must be a number'),
+    (12345678, 'numeric_id cannot be longer than 7 digits'),
+    pytest.mark.xfail((1234567, 'numeric_id cannot be longer than 7 digits')),
```
leohemsted: i'm not a huge fan of this test being in the same parametrize stuff, because it's not apparent whether this'll xfail because it doesn't raise a TypeError, or because it does raise a TypeError but the error string is different.

quis: Fixed in 57f07bf

tests/test_template_types.py
quis: Fixed in 43452a9

tests/test_template_types.py
quis: Fixed in 43452a9

notifications_utils/template.py
```diff
@@ -331,6 +357,83 @@ def __str__(self):
...
+
+    def __str__(self):
+        return '|'.join(line.replace('|', '') for line in [
+            '140',
+            '001',
+            '001',
+            '',
+            datetime.utcnow().strftime(
+                '%Y%m%d{0:07d}'.format(self.numeric_id)
+            ),
```
quis: Done as part of 22d3e6e

notifications_utils/template.py
```diff
@@ -331,6 +357,83 @@ def __str__(self):
...
+        if len(str(numeric_id)) > 7:
+            raise TypeError('numeric_id cannot be longer than 7 digits')
+        try:
+            int(numeric_id)
+        except ValueError:
+            raise TypeError('numeric_id must be a number')
+        self.numeric_id = int(numeric_id)
+
+    def __str__(self):
+        return '|'.join(line.replace('|', '') for line in [
```
quis: Yeah. That’s a better idea. Changed it in 499a71a.

notifications_utils/template.py
```diff
@@ -331,6 +332,80 @@ def __str__(self):
...
+    ):
+        super().__init__(template, values)
+        if not numeric_id:
+            raise TypeError('numeric_id is required')
+        if len(str(numeric_id)) > 7:
+            raise TypeError('numeric_id cannot be longer than 7 digits')
+        try:
+            int(numeric_id)
+        except ValueError:
+            raise TypeError('numeric_id must be a number')
```
quis: Think we can lose any casting and just validate. Changed in 7890454.

PR #114, opened 2017-02-07
Added methods for saving and retrieving hash maps from redis.

servingUpAces:
This PR implements methods to save and retrieve a hash in the redis client. 

The first use case for this is to cache the template statistics for the dashboard page, the key will be service_id-template-counter, the value will be a key-value map of template_id + number of notifications.

The cache will be given a number of seconds to expire the cache.

The increment_hash_value method may not be used immediately but may be useful later, I have left it in since I had already built it.


notifications_utils/clients/redis/redis_client.py
```diff
@@ -4,10 +4,10 @@
 
 class RedisClient:
     redis_store = FlaskRedis()
-    active = False
+    active = True
 
     def init_app(self, app):
-        self.active = app.config.get('REDIS_ENABLED')
+        # self.active = app.config.get('REDIS_ENABLED')
```
leohemsted: dont think this or the `active = True` above should be changed

notifications_utils/version.py
```diff
@@ -1 +1 @@
-__version__ = '13.3.1'
+__version__ = '13.4.1'
```
leohemsted: should be `13.4.0`

notifications_utils/clients/redis/redis_client.py
servingUpAces: Shoot I missed that, I changed it to do a real test.

PR #115, opened 2017-02-08
Enable Markdown-style links in emails

quis:
image

We want to do research into how users respond to links which are ‘hidden’ behind some text. The easiest way for us to do this is to send the emails from Notify. And the easiest way to make this links is using the default Markdown syntax.

So this commit turns on this feature of Markdown temporarily.

**If you are a user of Notify you should not rely on this. It is not a supported feature. It could stop working at any time.**


quis:
cc/ @leohemsted 

# 5836864 – Use neg lookbehind to ignore URL in Markdown links

This is the proper way to do this with a regex. It’s better because it doesn’t created a matching group, so you don’t risk replacing any of the content preceding the link.

notifications_utils/formatters.py
```diff
@@ -27,8 +27,9 @@ def unlink_govuk_escaped(message):
 
 def linkify(text):
     return re.sub(
-        #             url is anything without whitespace, >, or <
-        r'(https?:\/\/[^\s\<\>]+)($|\s)',
+        #  url is anything without whitespace, >, or <
+        #  Can’t be preceded by ( because that’s probably a Markdown link
+        r'(?<!\()(https?:\/\/[^\s\<\>]+)($|\s)',
```
leohemsted: quick thing - i think we should specifically limit this to `](`, because it seemsthat people have been using brackets round links in text anyway, eg

quis:
I’ve completely written this and squashed out some of the intermediate changes. Reason being:

> Implementing this also means replacing our hand-crafted auto-linker with one that uses Markdown to detect links. Two reasons for doing this:
>
>- avoiding awkward conflicts between our linker and Markdown’s link syntax
>
>- it’s generally better, for example it handles (http://links.in/brackets) properly, which ours doesn’t
>
> The only reason why we didn’t use Markdown’s autolinker in the first place is because were weren’t using Markdown at all when we added autolinking to email templates.
>
> Link detection was added on 6 June 2016: a66383a
>
> Markdown was added 15 July 2016: df9f167

PR #116, opened 2017-02-09
convert non-gsm characters to gsm-compatible versions in text messages

leohemsted:
GSM is an ascii-like (but not ascii) character set that text messages
are typically sent in. The providers we work with handle non-GSM chars
differently - one downgrades non-gsm to GSM compatible, and one sends
in unicode encoding, which takes three times as much space, and so
changes the amount billable.

This commit changes SMSTemplate to downgrade the content to GSM when
stringifying. For implementation details, read the docstrings in gsm.py


tests/test_base_template.py
```diff
@@ -122,9 +122,10 @@ def test_extracting_placeholders(template_content, template_subject, expected):
     "content,prefix, expected_length, expected_replaced_length",
     [
         ("The quick brown fox jumped over the lazy dog", None, 44, 44),
-        ("深", None, 3, 3),
+        # should be replaced with a ?
+        ("深", None, 1, 1),
         ("'First line.\n", None, 12, 12),
-        ("\t\n\r", None, 0, 0),
+        (" \n\r", None, 0, 0),
```
quis: Shouldn’t need to change this test – tab should still be replaced with space and then count as `0` length.

tests/test_gsm.py
```diff
@@ -0,0 +1,61 @@
...
+    assert gsm.get_unicode_char_from_codepoint(codepoint) == char
+
+
+@pytest.mark.parametrize('bad_input', [
+    '',
+    'GHIJ',
+    '0001";import sys;sys.exit(0)"'
+])
+def test_get_unicode_char_from_codepoint_rejects_bad_input(bad_input):
+    with pytest.raises(AssertionError):
```
quis: Is assertion error the best thing to raise here? Does it make sense if it should up in logs?

tests/test_base_template.py
leohemsted: good point, i modified it when it was returning `?` and forgot to change it back later

tests/test_gsm.py
leohemsted: an assertion felt good here, rather than a value error, because it's an internal function that typically acts on the output of unicodedata.decomposition, so it's not like the caller is expected to catch a valueerror and act on that (eg pass messages to people or something). i think if you get to this fn with something that isn't a unicode codepoint, something has gone  pear shaped somewhere

quis: OK, that sounds reasonable. 👍

leohemsted: actually, apparently pyc removes asserts as an optimisation step, so i'll change to raising a value error

leohemsted:
@quis should the `is_gsm_compatible` function return a list of non-gsm characters, so we can easily show them on the front end?

thinking something like `return set(content) - GSM_CHARACTERS`, which'll return a set containing just the characters not in GSM

quis:
@leohemsted Yeah, that seems better than straight true/false. Can assert on the set being empty/not in the same way.

tests/test_gsm.py
```diff
@@ -0,0 +1,71 @@
...
+    ('The quick brown fox jumps over the lazy dog', 'The quick brown fox jumps over the lazy dog'),
+])
+def test_encode_string(content, expected):
+    assert gsm.encode(content) == expected
+
+
+@pytest.mark.parametrize('content, expected', [
+    ('The quick brown fox jumps over the lazy dog', set()),
+    ('The “quick” brown fox has some downgradable characters', {'“', '”'}),
+    ('Need more 🐮🔔', {'🐮', '🔔'})
```
leohemsted: @bandesz

bandesz: It still needs more I think, but accepted! :)

tests/test_gsm.py
```diff
@@ -0,0 +1,71 @@
...
+    ('Łódź', '?odz'),
+    ('The quick brown fox jumps over the lazy dog', 'The quick brown fox jumps over the lazy dog'),
+])
+def test_encode_string(content, expected):
+    assert gsm.encode(content) == expected
+
+
+@pytest.mark.parametrize('content, expected', [
+    ('The quick brown fox jumps over the lazy dog', set()),
+    ('The “quick” brown fox has some downgradable characters', {'“', '”'}),
```
quis: Question about whether this should return `set()`. Is it annoying to get an error that says “You can’t use _em dash_ in your template” when we could downgrade it silently for you?

leohemsted: i see where you're coming from - though if you're a content designer who knows about correct quotes, would you be confused by it replacing them?

i suppose that might be nicer than leaving it up to the user to figure out where in their SMS they've used the illegal character

quis: I think that downgrading is the sensible thing – it’s not like the user has a choice about whether they can use the downgraded version or not.

quis: Downgrading is also more forgiving to existing templates (otherwise you could make an unrelated change and get the error message).

tests/test_gsm.py
```diff
@@ -64,8 +64,8 @@ def test_encode_string(content, expected):
 
 @pytest.mark.parametrize('content, expected', [
     ('The quick brown fox jumps over the lazy dog', set()),
-    ('The “quick” brown fox has some downgradable characters', {'“', '”'}),
+    ('The “quick” brown fox has some downgradable characters', set()),
```
quis: `set('👏', '👍', '🎉')`

PR #120, opened 2017-02-23
Fix minor typography issues in letter previews

quis:
The font size we were using for the address was too big. This is the info we have from DVLA:

> The top line is the MDI (Mailer Defined Information) this is in a custom mono-spaced font (based on Arial) and is 6pt.
> The rest of the address is in Arial, always upper case and 8pt.
> The line spacing for the whole block is Exactly 0.3 cm.

So this commit:
- uses the correct font size and line spacing for the address
- transforms the given address (but not the placeholders) to uppercase
- adds an example MDI line, to eep the preview faithful to what the final letter will look like

# DVLA

image004

# Us

image


quis:
# Also, add a line break between paragraphs and lists

It felt a bit tight before. A full line break between matches what GOV.UK Elements does.

## Before

screen shot 2017-02-23 at 12 19 24

## After

screen shot 2017-02-23 at 12 19 40

quis:
# One more thing: fix indent of ordered lists in letters

Since numbers are wider than bullets, they need to be indented a bit more.

## Before

screen shot 2017-02-23 at 12 50 47

## After

screen shot 2017-02-23 at 12 51 17

PR #123, opened 2017-02-27
Allow letters to have an optional contact block

quis:
Utils side of implementing this bit of letters:

img_6409

## Define the interface for the contact block

Like email branding or SMS sender it gets passed in at the time we create the template.

I think a string (as opposed to `list` of lines) is the most obvious interface for a block of text, even if we collect it and output it as a list.

## Pass contact block through to DVLA

This takes any string passed as a contact block to the `LetterDVLATemplate` and outputs it in the format they require, which is a number of (up to 10) columns in a file. We pad any unused columns with empty strings.

## Pass contact block through to preview

This commit takes any string passed to `LetterPreviewTemplate` as a contact block and generates a preview with this information in the top-right of the rendered letter (according to the dimensions on Sonia and Maria’s design).

Because the preview is generated from HTML, we have to convert newlines to HTML `<br>`s.

---

Minor version bump to 13.9.0


quis:
Rebased to fix version number conflicts.

PR #125, opened 2017-03-03
Fix type size for letters contact block

quis:
Should be smaller than the body text of the letter. Missed this when setting it up the first time because I had no way of previewing it.

Address and page numbers should be 10/12pt

Date and body are 12/16.5pt

image


quis:
Gonna yolomerge this one. It’s just two lines of CSS…

PR #127, opened 2017-03-22
Fill recipient address from bottom on letters

quis:
The DVLA print software fills the address lines from the bottom upwards. In other words the postcode (as the last line of the address) is always in the same place on the page, no matter how many lines of address precede this.

This commit makes this work the same in our previews.

The calculation used to get to the mm figure is:

```
A4 height − top/bottom margins − address top position − address height

   297mm  −   (10mm + 15mm)    −        29.5mm        −     26.7mm

= 215.8mm
```

Before | After 
--- | ---
image | image


quis:
# Make letter contact details wider

Typical contact details for services who send letters don’t fit in the space we’d allotted.

This commit increases the width from a third column to a two-fifths column. This is based on some experimentation that Sonia has been doing.

It will also need moving vertically, but I’ll do that work separately once we’ve decided on the best alignment.

## Before

image

## After 

image

quis:
Will need to move the date as well, but that’s a separate bit of work.

PR #128, opened 2017-03-24
Allow placeholder values to be lists, as well as strings

quis:
We need a way for users to put lists of variable length in their messages. It might be a list of properties you’re getting alerts for, or a list of documents that you need to provide.

## So, with this spreadsheet

phone number| list| list| list
---|---|---|---
07900900001| cat| rat| gnat

## …and this template

> Here are some animals:
>
> ((list))

## We make this `dict` internally…
```python
{'list', ['cat', 'rat', 'gnat']}
```

***

https://www.pivotaltracker.com/story/show/142228133


notifications_utils/field.py
```diff
@@ -106,3 +126,16 @@ def str2bool(value):
...
+    if len(list_of_strings) == 0:
+        return None
+    if len(list_of_strings) == 1:
+        return list_of_strings[0]
+    if len(list_of_strings) == 2:
+        return '{} and {}'.format(*list_of_strings)
+    return '{} and {}'.format(
+        ', '.join(list_of_strings[:-1]),
+        list_of_strings[-1],
+    )
```
leohemsted: i will do this myself on monday

PR #135, opened 2017-04-05
Handle any JSON type as placeholder value

quis:
One of our users was passing an integer as the value of their placeholder. We didn’t anticipate this, but it did work.

When we changed to explicitly allow placeholders to be lists as well as strings we inadvertently started ignoring anything that wasn’t a list or string. This meant that HTML for marking up the placeholder started showing up in people’s text messages 😳

This commit adds code to handle (as sensibly as possible) any valid JSON type, which are:
- number (integer or floating point)
- string
- boolean
- array
- object
- null

This is based on the principle that we should be liberal in what we accept, and, by having tests for all these types, predictable.


tests/test_field.py
```diff
@@ -46,6 +46,31 @@ def test_returns_a_string_without_placeholders(content):
...
+        ),
+        (
+            'before ((placeholder)) after',
+            {'placeholder': 123},
+            'before 123 after',
+        ),
+        (
+            'before ((placeholder)) after',
+            {'placeholder': 0.1 + 0.2},
+            'before 0.30000000000000004 after',
```
leohemsted: 🙈

notifications_utils/field.py
```diff
@@ -75,20 +75,28 @@ def get_replacement(self, match):
...
-            return self.sanitizer(unescaped_formatted_list(replacement, before_each='', after_each=''))
-
-        if isinstance(replacement, str):
-            return self.sanitizer(replacement) or ''
+            return self.get_replacement_as_list(list(filter(None, replacement)))
+
+        if isinstance(replacement, bool):
+            return str(replacement)
+
+        if replacement:
```
leohemsted: so if i've got this right here's the desired input/output for this:

```
None -> None
[] -> None

0 -> '0' (str)
'' -> '' (str)
bool -> 'bool' (str)

'truthy string' -> 'sanitized truthy str'
['truthy list'] -> get_replacement_as_list
```

don't think this'll catch the `0` case as it currently stands.

quis: Added code to handle the `0` case in 4c7ffd2

PR #136, opened 2017-04-05
Make list output conform to what works for DVLA

quis:
This is the markup for bulleted list that we have tested and know gives us the visually correct thing when we send it to DVLA:

```
Here is a list of bullets:  <cr><op><bul><tab>bullet 1  <op><bul><tab>bullet 2  <op><bul><tab>bullet 3  <p><cr><cr>And here is a list of long bullets…
```

This pull request changes the output for bulleted lists to match this.

It also changes the output for numbered lists, which was almost correct already, this is the tested, working markup:

```
Here is a list numbered:<cr><cr><np>bullet 1  <np>bullet 2  <np>bullet 3 <p><cr>And here is…
```
--- 

The only way I’ve differed is by putting one `<cr>` after the lists, not two. It looks like two is introducing too much space, but we need to double check this.


yahoopete:
:1st_place_medal: 

See how it goes with removing that (hopefully) extra cr tag.

PR #138, opened 2017-04-06
Refactor s3upload

servingUpAces:
Moved s3upload function to notifications-utlils, a common method used by api and admin.


notifications_utils/version.py
```diff
@@ -1 +1 @@
-__version__ = '15.0.0'
+__version__ = '15.0.1'
```
leohemsted: minor thing, but this is a major version (new functionality) - so should be `15.1.0`

notifications_utils/s3.py
```diff
@@ -0,0 +1,34 @@
+import uuid
+
+import botocore
+from boto3 import resource
+from flask import current_app
+
+
+def s3upload(filedata, region, bucket_name, file_location):
```
leohemsted: needs tests (even if they're just moved across from admin)

notifications_utils/version.py
quis: major thing, 15.0.0 to 15.1.0 is a minor version change :trollface:

PR #142, opened 2017-04-10
Make logos configurable in letter previews and files

quis:
Currently letters are hard-coded to use:
- _HM Government_ for the preview
- _HM Land Registry_ in the file we send to DVLA

We want to preview and print letters which have the logos of other organisations.

This commit makes configuring the logo an argument of the relevant `Template` subclass – the same way we solved this for email branding.

Still defaults to the same behaviour as currently, so that this isn’t a breaking change.


minglis:
Looks 👍

PR #146, opened 2017-04-12
Fix empty address placeholders becoming HTML

quis:
Situation
---
A CSV file has an optional address column (eg `address line 3`) but no values for this column in (some of) the rows.

Problem
---
The outputted data for those rows will have `<span class='placeholder'>…` in the address block, instead of empty strings.

Why it’s happening
---

`{'x': None}.get('x', '')` returns `None`, _not_ `''`.

`Field`s treat `None` as missing personalisation, and `''` as a valid replacement. So they were seeing missing personalisation when the intent of the code was to say _replace this bit of the address with an empty string because it’s optional_.

The fix
---
Converting `None` to `''` for all optional address lines (fix is in 01bd264, other commit is just whitespace changes).

### DIFF
--- a/notifications_utils/template.py
+++ b/notifications_utils/template.py
@@ -357,7 +357,7 @@ class LetterPreviewTemplate(WithSubjectTemplate):
         ).keys()
 
         return {
-            key: Columns(self.values).get(key, '')
+            key: Columns(self.values).get(key) or ''
             for key in keys
         }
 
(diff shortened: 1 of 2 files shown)


leohemsted:
although quick question - do we ever expect a mix of provided placeholders we want to replace, and missing placeholders we want to highlight?

quis:
@leohemsted Yeah – that’s this case https://github.com/alphagov/notifications-utils/pull/146/files#diff-e5050dea996611c8bdf32cfe6e8214cdR1023

I guess the principle is replace what we can, and show the brackets for the ones we can’t.

Feels like better feedback for the user because they know they’re on the right track. Rather than us saying ‘you got one thing wrong so I’m refusing to replace anything’.

PR #147, opened 2017-04-12
Don’t escape HTML in the data we output to DVLA

quis:
We have a problem where stuff like `&amp;` was showing up in our letters. We shouldn’t be doing this sanitising in the data we send to DVLA, but we also shouldn’t be letting users pass DVLA-markup like `<cr>` through us.


quis:
# Bonus commit: Don’t show pipes in letter preview

Since we strip pipes from the DVLA output we will never have them in our letters. This should be reflected in the preview we give to our users.

Doing as part of this PR to avoid merge conflicts.

PR #149, opened 2017-04-12
Changed the DVLA template

minglis:
- had previously a numericId
- this has changed to be a random string, 16 chars long.
- update the template and the tests to handle this


tests/test_template_types.py
```diff
@@ -928,43 +928,41 @@ def test_letter_output_pipe_delimiting():
...
+        notification_reference="1234567",
+    ).notification_reference == "1234567"
 
 
 @pytest.mark.parametrize('extra_args, expected_field', [
     ({}, '500'),
     ({'org_id': '001'}, '001'),
 ])
-def test_letter_output_numeric_id(extra_args, expected_field):
+def test_letter_output_notification_reference(extra_args, expected_field):
```
quis: This test should be called `test_letter_outputs_org_id` or something similar. My bad.

quis: Fixed in dae0ef67d9fa3d56194a39a91bbe70d0c08f87b6

quis:
Version number added, about to force push because of version number conflict.

PR #150, opened 2017-04-20
Generate image URLs for all pages of a letter

quis:
This is in order to show all pages of a letter in the admin app.

Takes the count of pages, returns HTML `` tags for each page in the letter.

Doesn’t (yet) remove the link to the PDF version because it’s probably still useful to have it for our testing.


tests/test_template_types.py
```diff
@@ -382,17 +384,52 @@ def test_letter_preview_renderer(
...
+    (
+        partial(LetterPDFLinkTemplate, page_count=1),
+        TypeError
+    ),
+    (
+        partial(LetterPDFLinkTemplate, preview_url='foo'),
+        TypeError
+    ),
+    (
+        partial(LetterPDFLinkTemplate, preview_url='foo', page_count='foo'),
```
leohemsted: nice use of partials 👌

PR #152, opened 2017-04-24
add international rates to notifications_utils

leohemsted:
(this also adds pyyaml to load the file)


tests/notifications_utils/test_international_billing_rates.py
```diff
@@ -0,0 +1,5 @@
+
+
+def test_international_billing_rates_exists():
+    from notifications_utils.international_billing_rates import INTERNATIONAL_BILLING_RATES
+    assert INTERNATIONAL_BILLING_RATES['1']['names'][0] == 'Canada'
```
kentsanggds: It's probably also a good idea to add tests to ensure that yaml file contains the correct fields.

PR #153, opened 2017-04-25
Rate limiting

minglis:
Rate limiting.
        - Uses Redis sorted sets
        - Also uses redis "multi" which is abstracted into pipeline() by FlaskRedis/PyRedis
        - Sends all commands to redis as a group to be executed atomically

        Method:
        (1) Add event, scored by timestamp (zadd). The score determines order in set.
        (2) Use zremrangebyscore to delete all set members with a score between
            - Earliest entry (lowest score == earliest timestamp) - represented as '-inf'
                and
            - Current timestamp minus the interval
            - Leaves only relevant entries in the set (those between now and now - interval)
        (3) Count the set
        (4) If count > limit fail request
        (5) Ensure we expire the set key to preserve space

        Notes:
        - Failed requests count. If over the limit and keep making requests you'll stay over the limit.
        - The actual value in the set is just the timestamp, the same as the score. We don't store any requets details.
        - return value of pip.execute() is an array containing the outcome of each call.
            - result[2] == outcome of pipe.zcard()
        - If redis is inactive, or we get an exception, allow the request


tests/test_redis_client.py
```diff
@@ -22,6 +27,7 @@ def build_redis_client(app, mocker):
                         return_value={b'template-1111': b'8', b'template-2222': b'8'})
     mocker.patch.object(redis_client.redis_store, 'hmset')
     mocker.patch.object(redis_client.redis_store, 'expire')
+    mocker.patch.object(redis_client.redis_store, 'pipeline', return_value=my_redis_mock)
```
leohemsted: one reason we don't want this to be a module level variable, i've just realised, is that `my_redis_mock` won't be reset between calls.  so any checks to say "did this function get called" will be looking at every test that has ran so far, and if a test that doesn't change `my_redis_mock.execute.return_value` runs after one that *does*, it'll use the results from the one that does

PR #154, opened 2017-04-25
Validate international phone numbers

quis:
This pull request adds validation that can be optionally enabled to allow international phone numbers. We consider an international phone number to be anything that starts with:

- `00` (but not `0`), and then a valid country prefix
- `+` and then a valid country prefix
- just a valid country prefix

As with UK numbers we ignore any brackets, dashes, spaces before validating.

There is one exception, Russia, whose country prefix is `7`. A number starting with `7` could be either a UK mobile number or a Russian phone number. So we only consider a phone number to be Russian if the part after the `7` is at least 10 digits.

Any number that starts with `7` and has fewer than 10 digits we treat as a UK mobile number. UK mobile numbers have 9 digits after the first `7`.

This pull request also does a bunch of other tidying up, but that’s the meat of it.

### DIFF
--- a/notifications_utils/recipients.py
+++ b/notifications_utils/recipients.py
@@ -406,6 +406,9 @@ def validate_phone_number(number, column=None, international=False):
     ):
         return validate_uk_phone_number(number)
 
+    if (number.startswith('0') and not number.startswith('00')):
+        return validate_phone_number(number)
+
     number = normalise_phone_number(number)
 
     if (
--- a/notifications_utils/recipients.py
+++ b/notifications_utils/recipients.py
@@ -380,7 +380,7 @@ def validate_phone_number(number, column=None):
 
     if not any(
         number.startswith(prefix)
-        for prefix in ['7', '07', '447', '4407', '00447']
+        for prefix in ['7', '447', '4407', '00447']
     ):
         raise InvalidPhoneError('Not a UK mobile number')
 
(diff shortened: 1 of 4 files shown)


notifications_utils/recipients.py
```diff
@@ -366,47 +369,63 @@ class InvalidAddressError(InvalidEmailError):
     pass
 
 
-def validate_phone_number(number, column=None):
+def normalise_phone_number(number):
```
imdadahad: Is it worth having a test for this?

tests/test_recipient_validation.py
```diff
@@ -60,6 +75,26 @@
...
+        (phone_number, error) for phone_number in group
+    ] for error, group in [
+        ('Not a valid country prefix', (
+            '800000000000',
+        )),
+        ('Not enough digits', (
+            '1234',
+        )),
+    ]
+], []))
```
imdadahad: On first glance this looks pretty confusing...

notifications_utils/recipients.py
imdadahad: Is it worth having a test for this, given that it does raise an exception and is used by the validation functions?

tests/test_recipient_validation.py
quis: good point, made it simpler in 756c140

notifications_utils/recipients.py
quis: Test added in 	ea56035

PR #156, opened 2017-04-26
Optionally allow international phone numbers in CSV files

quis:
Whether or not international phone numbers are permitted is on a service-by-service basis. So when we validate a CSV file we need to tell the bit of code that processes the CSV files whether or not the service is allowed to send to international phone numbers.

The CSV processing code can then decide what argument to pass to the bit of code that validates each individual recipient.


quis:
Version bump added in 6c24a7e

PR #166, opened 2017-05-30
Add human readable way to format a phone number

quis:
We got rid of this functionality before. But I think that now we’re doing inbound SMS, we might want to bring it back.

Two reasons:
- the number that people are texting from won’t be in the service user’s data already, so it doesn’t need to ‘match up’ (we’ll be doing the matching up)
- reading phone numbers is easier when they’re chunked up
- if we get inbound messages from international numbers, it’s more friendly to split the country code, and add a `+` to indicate that it’s international


quis:
# Use external library to format phone numbers

This commit brings in the `phonenumbers` library, which is [a port](https://github.com/daviddrysdale/python-phonenumbers) of [Google’s library for dealing with international phone numbers](https://github.com/googlei18n/libphonenumber).

The useful thing it gives us is locale-specific formatting of phone numbers. So in the USA the idiomatic way to format a phone number is:

> +1 202 555-5555

Whereas France doesn’t use dashes, and uses chunks of two digits:

> +33 1 12 34 56 78

So this commit changes our code to format numbers in whatever way people from that country are used to. So someone trying to match up a phone number (either visually or in conversation) with something they see in Notify should have an easier time. This is better than just chunking it at an arbitrary number of digits.

PR #169, opened 2017-06-15
Added ascii address validation

kentsanggds:
## What 

Added additional validation for addresses as DVLA doesn't support UTF-8 for address lines.

## How to review

Run `pytest -k test_non_ascii_address_line_raises_invalid_address_error tests` will carry out the tests to ensure that errors are raised.


notifications_utils/recipients.py
```diff
@@ -520,6 +521,17 @@ def validate_and_format_email_address(email_address):
...
 def validate_address(address_line, column):
+    if address_line:
+        def get_invalid_chars(s):
+            invalid_chars = [c for c in s if ord(c) > 255]
+            return sorted(set(invalid_chars))
+
+        invalid_chars = get_invalid_chars(address_line)
+        if len(invalid_chars) > 0:
+            raise InvalidAddressError(
+                "Can`t include: {}".format(
```
quis: Two things:
- can you remove the colon
- typographic apostrophe looks like `’`

notifications_utils/recipients.py
```diff
@@ -520,6 +521,17 @@ def validate_and_format_email_address(email_address):
 
 
 def validate_address(address_line, column):
+    if address_line:
+        def get_invalid_chars(s):
+            invalid_chars = [c for c in s if ord(c) > 255]
+            return sorted(set(invalid_chars))
+
+        invalid_chars = get_invalid_chars(address_line)
```
quis: Don’t think you need to define a function here:
```
invalid_chars = sorted(set(c for c in address_line if ord(c) > 255))
```

notifications_utils/recipients.py
```diff
@@ -520,6 +521,17 @@ def validate_and_format_email_address(email_address):
...
+        if len(invalid_chars) > 0:
```
quis: Don’t need to call `len` here. `if invalid_chars:` will work just the same.

notifications_utils/recipients.py
kentsanggds: ok

notifications_utils/recipients.py
kentsanggds: ok

notifications_utils/recipients.py
kentsanggds: ok

PR #171, opened 2017-06-23
Allow missing placeholders to ████████ redacted

quis:
We have a need to not show things like one time passwords or two factor codes when we show the content of messages.

We could just show the placeholders, but this might make people think that placeholders have accidentally been sent out to users.

What we’re really doing is redacting the content of the message. And the visual language of redaction is a thick black marker pen. This commit replicates that in unicode, so that our users are clear about what’s happening.

This is only relevant to `Template`, `WithSubjectTemplate`, `EmailPreviewTemplate`, `SMSPreviewTemplate` and`LetterPreviewTemplate`. These are the only ones which get used in the admin app, which is where the issue will show up.


quis:
Update: I’ve changed this to return HTML that we can style, rather than unicode block characters. bit more flexible, and better for screenreaders.

quis:
Update: Changed the invisible text to say ‘hidden’, not ‘redacted’ because this is the language we’re using to set this option on a template.

quis:
Changed it to `redact_missing_personalisation` – better describes what it does.

PR #175, opened 2017-07-05
hardcode SMS_CHAR_COUNT_LIMIT

leohemsted:
it doesn't really make sense to go through the faff of having it
passed in, when it's always the same - we weren't properly handling
it on the front-end anyway


quis:
Thumbs up to just hard coding this. But I think an even better place to put it would be in the `Template`, then this whole logic is self-contained. Pseudo code:

```python
class Template():

    def message_too_long():
        return False

class SMSMessageTemplate():

    char_count_limit = 999

    def message_too_long():
        if self.char_count > self.char_count_limit:
            return True
```

quis:
Or:
```python

def message_too_long(self):
    if getattr(self, 'char_count_limit') and self.char_count > self.char_count_limit:
        return True
```

tests/test_template_types.py
```diff
@@ -1293,3 +1293,24 @@ def test_lists_in_combination_with_other_elements_in_letters(markdown, expected)
...
+def test_message_too_long(template_class):
+    body = ('b' * 200) + '((foo))'
+    template = template_class({'content': body}, prefix='a' * 100, values={'foo': 'c'*200})
+    assert template.is_message_too_long() is True
+
+
+@pytest.mark.parametrize('template_class', [
+    EmailPreviewTemplate,
+    HTMLEmailTemplate,
+    PlainTextEmailTemplate
```
quis: Should test letter templates here too.

notifications_utils/recipients.py
```diff
@@ -63,12 +66,11 @@ def __init__(
         whitelist=None,
         template=None,
         remaining_messages=sys.maxsize,
-        sms_character_limit=0,
         international_sms=False,
     ):
         self.file_data = file_data.strip(', \n\r\t')
         self.template_type = template_type
-        self.character_limit = int(sms_character_limit) if self.template_type == 'sms' else sys.maxsize
+        self.character_limit = SMS_CHAR_COUNT_LIMIT if self.template_type == 'sms' else sys.maxsize
```
quis: This line and the constant it references can be removed now I think?

leohemsted: We still need this `459` value somewhere - it's used in the error messages on both admin and api.

But happy for it to be elsewhere in utils. thoughts? Just in the __init__? `from notifications_utils import SMS_CHAR_COUNT_LIMIT`?

quis: Yup, that sounds good.

tests/test_template_types.py
leohemsted: added

leohemsted:
@quis updated

PR #186, opened 2017-07-20
Add content_type arg for s3upload

kentsanggds:
Added another arg which can be passed in for s3uploads, 
default is set to `binary/octet-stream` as this is the content type that csv files are uploaded with.

This is needed as part of the create organisation logo story to allow platform admin users to add organisations without having to update the git repository.

Reference - 
https://www.pivotaltracker.com/story/show/148289053


quis:
When you have this in admin, can you test that uploading all of these files and reading them back still works: https://github.com/alphagov/notifications-admin/tree/master/tests/spreadsheet_files

PR #188, opened 2017-07-25
Automatically do nice typography in emails and letters

quis:
Fixes:
- errant whitespace before full stops (follows on from https://github.com/alphagov/notifications-utils/pull/181)
- straight quotes
- use of hyphens where en dashes would be better

# Before

image

# After 

image


quis:
Trying en dashes with hair spaces. Think it looks too tight:
image

cc/ @markhurrell

markhurrell:
ok true - how about a thin space?

quis:
@markhurrell It’s subtle.

thin-space

***

thin-diff

tests/test_template_types.py
```diff
@@ -618,26 +626,89 @@ def test_templates_handle_html_and_redacting(
...
+        mock.call(Markup('subject')),
+        mock.call(Markup('<p>content</p>')),
+        mock.call(Markup('subject')),
+        mock.call(Markup('subject')),
+    ]),
+    (LetterDVLATemplate, {'notification_reference': "1", 'contact_block': 'www.gov.uk  '}, [
+        mock.call('subject'),
+        mock.call('content<cr><cr>'),
+        mock.call('subject'),
+        mock.call('subject'),
```
leohemsted: These tests all feel very tautological - they test what they test because the code that they run produces the results that it does.

what would happen if tomorrow `str(LetterDVLATemplate)` only called make_quotes_smart with `'subject'` twice instead of three times?

quis: I guess the thing that I’m trying to test is that it gets called at least once – ie that we’re not skipping doing `make_quotes_smart` on the subject.

PR #189, opened 2017-07-25
Don’t check for values before counting characters in text messages

quis:
This conditional stuff about what to do when placeholders are/aren’t filled is handled by the `Field` instance now. Can remove it here to make the code cleaner. Tests for this are pretty robust:
https://github.com/alphagov/notifications-utils/blob/f90242731c54827b161efd334aad41129cecb000/tests/test_base_template.py#L122-L143


quis:
I didn’t understand this properly. Extra lines are needed.

PR #194, opened 2017-08-01
Allow multiple newlines in letter templates

quis:
Normally, the default Markdown behaviour of collapsing multiple newlines is desirable, because we want consistency in the amount/size of whitespace appearing in our emails and letters.  

However, for letters, we need to give teams control over vertical space so that they can push content onto the next page of a letter. This is because they have a legislative obligation to do this. And while it’s a hacky way of meeting that obligation (compared to, say, a page break) it’s the best we can offer for now.

---

The implementation of this is even hackier.

A fundamental tenet of Markdown is that it ignores lines that are empty, and assumes they are just there to demarcate two blocks of text. This is a behaviour that’s hard to override in the parser (which is what we’ve done for other custom features) because it’s so fundamental.

Instead we have to hack around this behaviour: Because Markdown ignores empty lines, we can make it pay attention to those lines by making them not empty.

So this PR adds two functions: 

1. One to fill sequences of empty lines longer than one with some  special characters that make them non-empty. 

2. One to remove the special characters after the Markdown parsing is  done.  The special characters were chosen to be something that no-one would type in an actual message.


notifications_utils/formatters.py
```diff
@@ -44,6 +44,12 @@
 
 whitespace_before_commas = re.compile(r'\s+,')
 
+multiple_newlines = re.compile(r'((\n)\2{2,})')
+
+MAGIC_SEQUENCE = "🇬🇧🐦✉️"
```
leohemsted: oh man i'm going to write this all the time

tests/test_template_types.py
```diff
@@ -1386,3 +1386,69 @@ def test_non_sms_ignores_message_too_long(template_class, kwargs):
...
+)
+def test_multiple_newlines_in_letters(
+    content,
+    expected_preview_markup,
+    expected_dvla_markup
+):
+    print(expected_dvla_markup)
+    print(str(LetterDVLATemplate(
+        {'content': content, 'subject': 'foo'}, notification_reference=1
+    )))
```
leohemsted: 🖨 🖨

PR #197, opened 2017-08-02
Tweak spacing around title of letters

quis:
DVLA have got the line spacing how we want it now. But this has had a knock-on effect on the spacing around the title in letters. Controlling this spacing is more finesse than their software allows.

This means that we need to change the appearance of our letters to match what they can do, ie:
- a bit more space above the title
- a bit letter space below

The adjacent selector is needed because otherwise the top margin of the `<h2>` would mean any reduction in the top margin of the `<h1>` would have no effect because of [margin collapsing](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing).

***

The other thing that needs adjusting is the vertical position of the main text block, but DVLA can do this (it needs to move down by 4mm for the two examples below to line up).

## Before 

nvd

## After

Let’s see how this works…


quis:
Going to YOLOmerge this because it’s just CSS changes.

PR #208, opened 2017-08-22
Update the Flask dependency to 0.12.2

servingUpAces:
The api and admin app have already been updated to this version.

This will get rid of warnings during start up due to the deprecated use of flask.ext.redis.


leohemsted:
we should probably put @pyup-bot on utils ( @quis )

quis:
Or could we import just Flask-Redis rather than the whole of Flask?

https://pypi.python.org/pypi/Flask-Redis

PR #217, opened 2017-08-31
use a WatchedFileHandler for log messages

leohemsted:
We currently use TimedRotatingFileHandler to rotate log files every
day. However, this causes issues when multiple celery workers are
logging to the same file - when the file rotated at midnight, handlers
wouldn't be correctly set up for all processes to point to the new
file. As a result, we'd lose log messages if the box had been alive
since the previous day :grimacing:.

This commit uses a WatchedFileHandler instead, and relies on separate
processes to do the rotation. WatchedFileHandler is a builtin file
handler that checks that the file is the same before writing each line,
so if the old file is moved, it'll happily create a new file and write
to that, and each process will update their handlers accordingly. No
old logs overwritte, no new logs lost. :tada:


notifications_utils/logging.py
```diff
@@ -106,9 +106,8 @@ def get_handlers(app):
 
     if not app.debug:
         # machine readable json to file
-        handler = logging.handlers.TimedRotatingFileHandler(
-            filename='{}-{}.json'.format(app.config['NOTIFY_LOG_PATH'], os.getpid()),
-            when='midnight'
+        handler = logging.handlers.WatchedFileHandler(
+            filename='{}-{}.json'.format(app.config['NOTIFY_LOG_PATH'], os.getpid())
```
sakisv: A bit late to the party, do we need the pid anymore?

leohemsted: think we do for FTP while we still have logs from both gunicorn and celery

PR #221, opened 2017-09-11
Migrate code for letter timing from admin app

quis:
We want to return letter timings on the API side as well. This means the code should be shared between the two apps. Utils is the place where shared code lives.

Code was originally introduced here: https://github.com/alphagov/notifications-admin/pull/1362

Also migrates the code for converting UTC to GMT from the admin app, because its needed by the letter timings stuff (because DVLA print at 5pm clock time, not UTC – makes sense). This code didn’t have any unit tests, so I’ve also added a couple.


quis:
## Make delivery estimates default to 4pm

The delivery estimate should be able to give you a reasonable chance of guessing whether a letter has been delivered or not.  According to Royal Mail:

> We aim to deliver to addresses in urban areas by 3pm every working day and addresses in rural areas by 4pm every working day
– http://www.royalmail.com/sites/default/files/docs/pdf/106831_col_del_pdf_02.pdf..unpublished 

So if you wait until after 4pm you can be reasonably sure that a letter has been delivered.  Only changing this because the API _has_ to return an hour (to be a valid timestamp) so it might as well return the most sensible hour I can think of.

quis:
## Ensure datetimes returned are UTC

We deal with UTC everywhere, up to the point where we present a time back to the user on the frontend. So all code that is not specifically to do with presenting a date on the frontend should take dates assuming they’re in UTC (done) and return them as UTC. This commit does the latter.

PR #223, opened 2017-09-14
Updated email template to allow branding banners

gov-cjwaszczuk:


notifications_utils/email_template.jinja2
```diff
@@ -104,6 +104,53 @@
...
+                  <span style="
+                    display: block;
+                    padding-left: 5px;
+                    font-family: Helvetica, Arial, sans-serif;
+                    font-weight: 700;
+                    font-size: 19px;
+                    color: #ffffff;
+                    line-height: 25px;
+                    margin: 14px 0;
+                    {% if not brand_logo %}
```
quis: Should this not be `if brand_logo` – ie if there’s an image, add some space between it and the text?

tests/test_template_types.py
```diff
@@ -54,6 +54,47 @@ def test_govuk_banner(show_banner):
         assert "GOV.UK" not in str(email)
 
 
+def test_brand_banner_shows():
```
quis: I think this test would be better if it passed in `brand_logo`, `brand_name` and `brand_colour` but set `brand_banner=False`, and then assert that the logo, name and colour don’t appear in the HTML. A bit more solid than asserting for arbitrary HTML I think.

PR #227, opened 2017-09-20
Preview reply to address in email template

quis:
image

This pull request:
- deprecates the from address argument, and stops showing it in the frontend preview
- adds a new argument to pass in a reply to address for an email, which will be shown in the front end preview

The reply to address argument is optional argument – if it’s not passed in then nothing will be shown. In the future this means that we could:
- show the reply to address for all email previews
- show it only in the case where you’ve had to choose from one or more reply to addresses


quis:
Also:

##  Optionally show sender in text message preview

This will be used when picking from multiple text message senders.

screen shot 2017-09-22 at 10 55 54

PR #229, opened 2017-09-28
Add alt tag for GOV UK Crown

mathieuhendey:
While submitting the MOT Reminders service for an accessibility audit,
we found that the GOV UK Crown in the email notifications  was missing
an alt tag. This was flagged up during the audit, and it was recommended
that we add the alt tag to reduce confusion for users using screen
readers.

### DIFF
--- a/notifications_utils/email_template.jinja2
+++ b/notifications_utils/email_template.jinja2
@@ -48,7 +48,7 @@
             <td width="70" bgcolor="#0b0c0c" valign="middle">
               <a href="https://www.gov.uk" style="text-decoration: none;"><img
                       src="https://static.notifications.service.gov.uk/images/gov.uk_logotype_crown.png"
-                      alt="" height="32" border="0" style="margin-top: 4px; margin-left: 10px;"></a>
+                      alt="GOV UK Crown" height="32" border="0" style="margin-top: 4px; margin-left: 10px;"></a>
             </td>
             <td width="100%" bgcolor="#0b0c0c" valign="middle" align="left">
               <span style="padding-left: 5px;"><a href="https://www.gov.uk" style="


quis:
Hey @mathieuhendey, thanks for raising this. As far as I understand it the crown is presentational. This means that it should be ignored by screenreaders. The way to make screenreaders ignore presentational images is to give them empty `alt` text.

This is the same thing that GOV.UK itself does:
screen shot 2017-09-28 at 12 54 21

I can double check this with the @alphagov/accessibility team as well if you’d like?

quis:
Just to clarify: the source of GOV.UK uses alt=“”, it’s just Chrome dev tools that strips out the =“”

nickcolley:
@quis I saw a presentation from the FT where they added role="presentation" as well as the alt="" due to some screenreaders still trying to read an image out...

selfthinker:
@nickcolley, the FT mentioned which screen readers but I suspect they meant the combination of Chrome and JAWS or NVDA (I don't remember which one). That indeed reads out the `src` and completely ignores an empty `alt` parameter. But that is a) a browser bug and b) in a combination of browsers and screen readers which is extremely rare. There are not many screen reader users who use that combination because it is so buggy (on many levels). We wouldn't recommend to use `role="presentation"` because it's a bit like fixing things for IE6 and the very few people using that combination will be used to that bug, but it doesn't hurt adding it either.

I agree with @quis that an empty `alt` parameter is theoretically correct here because it's of a decorative nature. Ask yourself, would it make sense to you if the image wasn't there and the text would be there instead, i.e. "GOV UK Crown GOV.UK"? If it doesn't, that's an indicator that it doesn't need the alternative text (or it needs a different one).
The reason why the auditor is actually sort of right here is because the crown as well as the text "GOV.UK" is a link. And you shouldn't have links with no text (which is the case for a link around an image with an empty `alt` text). And because the link goes to gov.uk, it should say that, i.e. the `alt` parameter should be "GOV.UK" as well. (A link saying it is for "GOV UK Crown" doesn't make much sense.) But the problem with that is that it is repeating the same thing again, which also violates some accessibility guideline.

What I would suggest to do instead is:
* either remove the link from the crown completely (which might be a usability issue because people might still want to click on it)
* or try to put both the crown and the text "GOV.UK" onto the same link (as it's done on gov.uk), no idea if it's easy or difficult to do in an email template

By the way, lots of people get the copy for `alt` parameters wrong. A very good resource I can recommend for that is the [W3C's alt decision tree](https://www.w3.org/WAI/tutorials/images/decision-tree/).

mathieuhendey:
Agree with all the points above, but as I say, the auditors picked it up and said it was causing them issues. The auditors we use are themselves reliant on screenreaders as they are visually impaired, so it stands to reason that the issues they flagged up to us would affect all users of screenreaders.

If @alphagov/accessibility could comment that would be great, as we could then go back to the auditors with this information.

Thanks all.

selfthinker:
@mathieuhendey, I am part of the @alphagov/accessibility team. Before I posted, I also discussed this with @accessiblewebuk who is also part of the team and he agreed.

As I said, the auditors are not wrong with flagging it up but giving the `alt` parameter a text would be the wrong solution.

quis:
Thanks @nickcolley and @selfthinker for the comments.

As a first attempt, we’re going to look into whether we can put the crown and GOV.UK text into one `<a>`, rather than two separate `<a>`s.

I’ll leave this PR open as an issue, until we have a better solution.

selfthinker:
@edwardhorsford, the **FT** was testing images in that unusual/buggy combination, not the auditors. Those are two unrelated issues we were talking about here.

mathieuhendey:
Our auditors used a combination of Chrome and JAWS. Is this really that unusual a combination? Chrome being probably the commonest browser and JAWS being one of the commonest screenreaders?

I've shared everyone's comments with the rest of my team, some interesting information in here that we didn't know so thanks for all your contributions to the conversation!

selfthinker:
I was sure about Chrome but not about which screen reader it was. I'm pretty sure now it was the combination of **Chrome and VoiceOver** on a Mac, not NVDA or JAWS. That is the combination I meant above when I wrote that it is so rare because it is so very buggy. Sorry about the confusion.

Others will know more about this, but as far as I know the combination of Chrome and JAWS is less common than IE and JAWS but should be alright. The combination of Chrome and VoiceOver is much, much rarer.

quis:
There’s some good investigation here too: http://john.foliot.ca/aria-hidden/

selfthinker:
Using ARIA to hide the crown and the link around it is a viable option as well.
Putting crown and text into one link would still be preferable because it is more semantic and follows the [first rule of ARIA](https://www.w3.org/TR/using-aria/#rule1). But if that doesn't work for whatever reason, hiding it with ARIA would fix the issue as well.

gov-cjwaszczuk:
https://github.com/alphagov/notifications-utils/pull/281 This has been continued here

quis:
@mathieuhendey I’m closing this in favour of https://github.com/alphagov/notifications-utils/pull/281

We’ll be deploying it in the next few days.

Thanks again for prompting us to get this right.

PR #232, opened 2017-10-03
Fixed accessibility bug with email header

gov-cjwaszczuk:
## Problem

There was an accessibility bug raised with the header of the emails where the crown was causing confusion for screen readers. In the previous implementation, the images's src was read out alongside the text.

## Solution

The text for the header was added into the same link as the crown, the anchor tag also now has an aria-label on it that reads out "Go to the GOV.UK homepage". A title attribute was attempted first however the screen readers still read out the image src.

***

Closes #229


quis:
Are these the test results for this change? https://www.emailonacid.com/app/acidtest/qCCoL1UkZMKQ9TqI6DY5NvAHE11sjO2RVLLcpFdFoittB/list

gov-cjwaszczuk:
https://www.emailonacid.com/app/acidtest/MnkEN0WrJUFI1ov0jpXox4QRHFv5jS7f8e8EwOcFrJmDy/list should be the correct results

selfthinker:
When you say "the screen readers still read out the image src", do you mean "VoiceOver in Chrome on a Mac still reads out the image src"? VoiceOver on a Mac should generally be only tested with Safari as Chrome is so buggy that the majority of screen reader users are not using it.
Only beginning of this week have we finally updated our guidance, including which assistive tech and browser combinations to test in: https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#testing-with-assistive-technologies

A `title` should be fine, although no `title` or `aria-label` should also be fine as "GOV.UK" by itself should be clear enough. But having either of them doesn't hurt either.
I noticed we use a `title` on the gov.uk logo link. It would probably make sense to be consistent with that.

quis:
Can we also try this trick with the margins to see if it fixes the squashed-together crown and GOV.UK text in Outlook? 

https://www.emailonacid.com/blog/article/email-development/outlook.com-does-support-margins

selfthinker:
Why was this closed? And, what was the holdup with merging it?

gov-cjwaszczuk:
Apologies, we're trying to solve a different problem at the same time, It will be moved into a different PR with the other problem aswell.

selfthinker:
Okay, just make sure to add explanations to PRs when closing them and also to reference the new PR so people can follow what's going on.

PR #235, opened 2017-10-16
Update email validation regex to reject commas in local part of address

klssmith:
- Also added a test to check that email addresses with a comma in the
domain part are not validated.


tests/test_recipient_validation.py
```diff
@@ -138,6 +138,7 @@
     'spaces-in-domain@dom ain.com',
     'underscores-in-domain@dom_ain.com',
     'pipe-in-domain@example.com|gov.uk',
+    'comma,in-domain@gov.uk',
```
quis: This is `comma,in-local` technically (but functionally it doesn’t make a difference).

klssmith: Thanks - now updated.

quis:
Can you add a test for a comma in the domain part as well (and make it pass if it doesn’t already)?

klssmith:
@quis I've added in a test for commas in the domain part - (commas in the domain were already considered invalid, but we had no test for it)

PR #242, opened 2017-10-27
use python35 in tests                                                …

leohemsted:
we already use python 35 on paas


leohemsted:
merging as while this PR keeps open other tests are gonna fail due to 3.5 binaries conflicting with other PRs trying to run in 3.4

PR #250, opened 2017-11-14
Add secret request header check

sakisv:
This is a **backwards incompatible** change as it changes the name of the module.

It adds a function that is bound as `before_request` function on both admin and api, which checks whether the request has the appropriate headers set by the proxies.

It's important to note that the current implementation fails closed, so for example if there are no keys configured on the app to check the header value against, then all the requests will be rejected.


notifications_utils/request_helper.py
```diff
@@ -45,3 +45,47 @@ def rewrite_response_headers(status, headers, exc_info=None):
...
+
+    # if there isn't any non-empty secret configured we fail closed
+    if not any(secrets):
+        return False, "Secrets are not configured"
+
+    key_used = None
+    for i, secret in enumerate(secrets):
+        if header_secret == secret:
+            key_used = i + 1  # add 1 to make it human-compatible
+            break
```
allait: Instead of `break` here you could `return True, ...` right away, and then replace the rest of the code with the default case of `return False, ...`

tests/test_request_header_authentication.py
```diff
@@ -0,0 +1,93 @@
+from unittest import mock  # noqa
+from werkzeug.test import EnvironBuilder
+
+from notifications_utils.request_helper import CustomRequest, _check_proxy_header_secret
+
+
+def test_request_header_contains_correct_secret():
+    builder = EnvironBuilder()
```
allait: These 3 lines could make a nice py.test fixture since they're repeated in most tests in this file. Alternatively, the tests are probably similar enough to be grouped and run with `@pytest.mark.parametrize`.

notifications_utils/request_helper.py
sakisv: Nice one, done 👍

tests/test_request_header_authentication.py
sakisv: Updated to use the `@pytest.mark.parametrize`. 

Although I'm not sure how I feel about adding logic to tests 🤔

tests/test_request_header_authentication.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@pytest.mark.parametrize('test_input,expected', [
+    ({'header_name': 'X-Custom-Forwarder', 'header_value': 'right_key', 'secrets': ['right_key', 'old_key']}, (True, 'Key used: 1')),  # noqa
+    ({'header_name': 'X-Custom-Forwarder', 'header_value': 'right_key', 'secrets': ['right_key']}, (True, 'Key used: 1')),  # noqa
+    ({'header_name': 'X-Custom-Forwarder', 'header_value': 'right_key', 'secrets': ['right_key', '']}, (True, 'Key used: 1')),  # noqa
+    ({'header_name': 'My-New-Header', 'header_value': 'right_key', 'secrets': ['right_key', '']}, (True, 'Key used: 1')),  # noqa
+    ({'header_name': 'X-Custom-Forwarder', 'header_value': 'right_key', 'secrets': ['', 'right_key']}, (True, 'Key used: 2')),  # noqa
+    ({'header_name': 'X-Custom-Forwarder', 'header_value': 'right_key', 'secrets': ['', 'old_key', 'right_key']}, (True, 'Key used: 3')),  # noqa
+    ({'secrets': ['old_key', 'right_key']}, (False, 'Header missing')),
```
leohemsted: i'd pull out this into its own test since it has less in common with the other tests

tests/test_request_header_authentication.py
```diff
@@ -0,0 +1,31 @@
...
+    ({'header_name': 'X-Custom-Forwarder', 'header_value': '', 'secrets': ['right_key', 'old_key']}, (False, 'Header exists but is empty')),  # noqa
+    ({'header_name': 'X-Custom-Forwarder', 'header_value': 'right_key', 'secrets': ['', None]}, (False, 'Secrets are not configured')),  # noqa
+    ({'header_name': 'X-Custom-Forwarder', 'header_value': 'wrong_key', 'secrets': ['right_key', 'old_key']}, (False, "Header didn't match any keys")),  # noqa
```
leohemsted: i'd make the parametrize have three values: `headers, secrets, expected_output`. then the first arg would be `{'X-Custom-Forwarder': 'wrong_key'}`, second would be `['right_key', 'old_key']`, and third would be as it is.

You can then get rid of the logic in the test - and just call `builder.headers.update(headers)`

tests/test_request_header_authentication.py
sakisv: Done

tests/test_request_header_authentication.py
sakisv: Done

PR #253, opened 2017-11-16
Add header check feature flag

sakisv:


sakisv:
Although this raises the question of whether we need to check the `DEBUG` setting 🤔

leohemsted:
yeah good point, i guess we don't any more

PR #255, opened 2017-11-17
Don't spam the logs

sakisv:
Even on preview we were generating way too many `info` logs.


sakisv:
Merging to create the version and get the rest of the tests to run.

PR #257, opened 2017-11-23
add try_validate_and_format_phone_number, which doesnt err if invalid

leohemsted:
This is for use with numbers we receive messages from - we don't want to crash out and error if they're not real phone numbers (because they might have been sent by a shortcode or alphanumeric).


notifications_utils/recipients.py
```diff
@@ -553,8 +568,11 @@ def format_recipient(recipient):
 
 
 def format_phone_number_human_readable(phone_number):
-
-    phone_number = validate_phone_number(phone_number, international=True)
+    try:
+        phone_number = validate_phone_number(phone_number, international=True)
+    except Exception:
```
quis: There’s a custom exception you can catch here (`InvalidPhoneError`).

PR #264, opened 2017-12-05
Allow spreadsheet rows to be looked up by index

quis:
In the future we want to be able to show a preview of any message in your file. This means being able to get the value of any row in the spreadsheet.

Currently the admin app looks at the first item in the `.rows` property of `RecipientCSV`[1].

This commit changes the class to allow access to any row in the file. It does this by implementing the `__getitem__` method, which is the magic method that gets called when evaluating `instance[i]`.

---

I haven‘t bothered implementing reverse indexing (`[-1]`) or slicing (`[1:2]`) because we don’t have a use case for them at the moment.


quis:
Don’t think there’s any other way because the CSV library returns a generator.

quis:
Also the front end isn’t going to let you go past 50 or 100 rows so probably a non-issue performance wise.

PR #275, opened 2017-12-11
Add support for horizontal rules in emails

quis:
We’re finding that the GOV.UK email alerts are quite complex in terms of hierarchy.

We think that a horizontal rule (with some spacing around it) is an effective way to break up the content a bit, thereby making it clearer what the structure of the content is.

This will be an undocumented feature to start with, though we think it will be useful to more than just the GOV.UK email team.

---

Tested here: https://www.emailonacid.com/app/acidtest/4svNdjgI2ZuwjSStM17lP4vFbGiFe0hC4WWoQKb53qcth/list

---


quis:
@leohemsted `***` or `---`; test is here: https://github.com/alphagov/notifications-utils/pull/275/files#diff-30615124ca17c3eee3ddd0a2b3e988a8R324

PR #281, opened 2017-12-19
Fix broken outlook header and content / make header accessible

gov-cjwaszczuk:
### Fixed accessibility bug
There were problems with the link in the header not being read correctly on accessibility devices, to solve this we placed the image and the text within the same a tag. Uses tables to work with Outlook clients. 
### Fixed Outlook 2016 bugs
Outlook clients were having issues with ordered lists within content. Margin was not working and the ordered list ended up tabbed in. Margin issues were solved using tables and the indent issue was solved by decreasing the text indent within list items.

Outlook 2016 on windows 10 was also having a jarring line run through the header, problem was solved by moving the font-size and line-height onto the parent `<td>`, rather than on the span itself.

**Email on acid link:** https://www.emailonacid.com/app/acidtest/ZcmhbFa4nettJYBG5Q8lCgDo6dvZs9PdK254OvlX40S7y/list
### Before 
image
image
### After
image


notifications_utils/formatters.py
```diff
@@ -364,22 +364,34 @@ def linebreak(self):
 
     def list(self, body, ordered=True):
         return (
-            '<ol style="Margin: 0 0 20px 0; padding: 0; list-style-type: decimal;">'
+            '<table style="padding: 0 0 20px 0;">'
```
quis: You should add `role='presentation'` to any tables which are being used for layout (so they don’t confuse screenreaders).

gov-cjwaszczuk:
@selfthinker This is the new PR for the a11y issue with the header being read incorrectly, I tried using title, rather than aria-label and it seemed to not do anything within gmail on safari, so I reverted back to aria. Do you have any other thoughts on the issue?

quis:
Nice work. The lists and the crown look 👌 

Think you might have introduced a new problem where the content doesn’t line up with the logo in some versions of Outlook though:

image

gov-cjwaszczuk:
I'll run over it again and make the tweak in a bit

selfthinker:
When you say "it seemed to not do anything within gmail on safari", I guess you tested in VoiceOver?
Title and aria-label are not the same. When you use `title` it will be read out additionally to the link text (so, something like "link - Go to the GOV.UK homepage - GOV.UK"). When you use `aria-label` it will be read **instead of** the link text (so, probably "link - Go to the GOV.UK homepage").
Here is a good resource which demonstrates which screen reader combination does it correctly: [A "click here" link with TITLE attribute](https://www.powermapper.com/tests/screen-readers/labelling/a-title/) and [Link text replaced by aria-label attribute](https://www.powermapper.com/tests/screen-readers/labelling/a-aria-label/).

I personally think screen readers usually don't need to read out anything different to what the sighted user sees. But as we do give them extra context on gov.uk, it makes sense to keep that consistent.

I wouldn't change code just to suit a buggy screen reader (here: VoiceOver) unless it presents a real barrier. Most screen readers will be fine with it.

selfthinker:
I forgot another reason not to use `aria-label` but `title`. According to the spec `aria-label` should generally only be used where the link text is not visible.

> If the label text is visible on screen, authors SHOULD use aria-labelledby and SHOULD NOT use aria-label.

from https://www.w3.org/TR/wai-aria-1.1/#aria-label

gov-cjwaszczuk:
@quis See most recent Email On Acid for new commit changes.

PR #312, opened 2018-01-16
Fixed border issue on tables in admin

gov-cjwaszczuk:


quis:
I’d prefer if we did this with CSS in the admin app. 

Doing it here means that we’re including this code in every email we send, even though it isn’t needed by any email clients.

PR #354, opened 2018-02-21
Added a new utils files for pdf and a method to count the number of Pages in a PDF

richardc0:
This will be used so that we can bill the service correctly, as we need
to count the pages in the pdf and record this against the notification.

* Added pdf utils file with a count pdf pages method
* Added tests to test in various conditions
* Added a const file with the pdf data in it and how the value was created
* Updated the version number
* Added test files for provenance


quis:
Is this better than/does it replace the code we have to do this in template preview already?

cf https://github.com/alphagov/notifications-template-preview/blob/7ce96fbf7a3432d2d02e60da0cb8db329aca888d/app/preview.py#L50-L52

richardc0:
@quis Its better for this implementation because:

1. Its native Python so no need for the docker image etc and we don't have to make a round trip to Template Preview
2. It's quicker because we are not converting to an image (a one page pdf is 80 times faster using this library than using the Image) to get the page count and we are not making the round trip to template preview

I think this could be used for template preview (hence why I put it in utils) as it would be quicker than converting it to an image - I'll add it to the list to try out but as we are converting to an image as well there maybe other efficiencies which means using this is slower?

quis:
Makes sense 👍 

Should look into using it for template preview if it makes things quicker.

PR #381, opened 2018-03-05
Refactor and tidy up handling of CSV files

quis:
Was looking at this code the other day as part of handling duplicate columns. 

I realised that:
- the code could be factored a lot better 
- there were lots of unused methods which have been created and then dropped over the two years this code has been in use

There were also assumptions built in to the code based on how we thought it would be used; now we can see how it’s actually used in the admin and API apps and make the API of the `RecipientCSV` class geared towards how it’s used.

Detail in the commits…


notifications_utils/recipients.py
```diff
@@ -186,36 +182,17 @@ def get_rows(self):
...
-                for recipient_column in self.recipient_column_headers
-            )
-        )
-
-    @property
-    def rows_with_message_too_long(self):
-        return set(
-            row['index'] for row in self.annotated_rows if row['message_too_long']
-        )
+            if index < self.max_rows:
```
leohemsted: should this if check should be at the top of the for loop to avoid parsing columns at all if there are too many rows

notifications_utils/recipients.py
```diff
@@ -225,71 +202,38 @@ def more_rows_than_can_send(self):
...
-        for row in self.annotated_rows:
-            if RecipientCSV.row_has_error(row):
-                yield row
+    def displayed_rows(self):
+        if any(self.rows_with_errors) and not self.missing_column_headers:
+            return self.initial_rows_with_errors
+        return self.initial_rows
+
+    def _filter_rows(self, attr):
+        return filter(lambda row: getattr(row, attr), self.rows)
```
leohemsted: but [guido doesn't like lambdas](https://www.artima.com/weblogs/viewpost.jsp?thread=98196) :(

`return (row for row in self.rows if getattr(row, attr))`

(i don't mind if you keep the filters tho)

quis: Changed in 8fb32c0fad99a68b9750db04a30de8c4bb1bf12b

notifications_utils/recipients.py
quis: Trouble is we don’t know how many rows there are until we’ve processed them…

notifications_utils/columns.py
```diff
@@ -1,40 +1,28 @@
...
-
-    def items(self):
-        return self._dict.items()
-
-    def pop(self, key):
-        return self._dict.pop(key)
+        try:
+            return self[key]
+        except IndexError:
+            return default
```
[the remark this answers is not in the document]
quis: Had a bit more of a think about how this should work, see 19d29e2

PR #387, opened 2018-03-06
Attempt to create the logging directory

sakisv:
Arises from alphagov/notifications-template-preview#102


notifications_utils/logging.py
```diff
@@ -93,6 +95,20 @@ def after_request(response):
...
 
+def ensure_log_path_exists(path):
+    """
+    This function assumes you're passing a path to a file and attempts to create
+    the path leading to that file.
+    """
+    path = Path(path)
+
+    # file next to the running script
+    if len(path.parts) == 1:
```
leohemsted: this won't catch `/home/vcap/apps` or `/user/leo/dev/notify-api` will it

i think you can see what `path.relative_to(Path.cwd())` returns?

leohemsted: in fact, i think this is much nicer

```
if not path.parent.exists():
    path.parent.mkdir(mode=755, parents=True)
```

notifications_utils/logging.py
```diff
@@ -93,6 +95,20 @@ def after_request(response):
...
+        return
+
+    Path(path.parent).mkdir(mode=755, parents=True, exist_ok=True)
```
leohemsted: `path.parent` should return a path, so you can just do `path.parent.mkdir`

notifications_utils/logging.py
sakisv: (I only just saw your comments here /facepalm) 

As per the docstring, this function will treat all the paths as pointing to files, so regarding the first comment no, it won't catch things like `/home/vcap/apps` which is ok as this path shouldn't reach this function anyway.

As for the second comment, agreed 100%. In fact, taken together with the comment below, the whole function can be reduced to:

```python
def ensure_log_path_exists(path):
    Path(path).parent.mkdir(mode=755, parents=True, exist_ok=True)
```

The less code, the better 👍

PR #422, opened 2018-03-23
Redis expiry

leohemsted:
* add expire function to redis client to set TTL for a key
*  remove raise_exception and standardise logging

we don't currently use the raise_exceptions flag anywhere - lets make the redis client more standard by getting rid of it and never raising an exception. Also, get rid of some mysterious kwargs we never use for `set` too.

nb: technically a breaking version change since i removed a couple of
kwargs


notifications_utils/clients/redis/redis_client.py
```diff
@@ -90,24 +94,31 @@ def increment_hash_value(self, key, value, raise_exception=False, incr_by=1):
...
+                self.redis_store.expire(key, expire_in_seconds)
+            except Exception as e:
+                self.__handle_exception(e, raise_exception, 'expire', key)
 
-    def __handle_exception(self, e, raise_exception, message=None):
-        current_app.logger.exception("{} \n {}".format(message, e))
+    def __handle_exception(self, e, raise_exception, operation, key_name):
+        current_app.logger.exception('Redis error performing {} on {}'.format(operation, key_name))
         if raise_exception:
             raise e
```
richardc0: Are all exceptions likely to be RedisError?
I'd usually go for raising a custom exception (Error) here i.e. RedisClientError.

As a consumer I would want to easily catch exceptions from a library that would encapsulate RedisError but also any other error e.g. TypeError etc

tests/test_redis_client.py
```diff
@@ -177,3 +179,9 @@ def test_should_not_call_rate_limit_if_not_enabled(mocked_redis_client, mocked_r
 
     assert not mocked_redis_client.exceeded_rate_limit('key', 100, 100)
     assert not mocked_redis_client.redis_store.pipeline.called
+
+
+def test_expire(mocked_redis_client):
```
richardc0: Add  some sort of test to check that the error message has the key and operation in it?

notifications_utils/clients/redis/redis_client.py
leohemsted: yeah our wrapper is super simple and I've never seen anything other than a `RedisError` thrown - if there's something else i'd probs want us to investigate it as a bug anyway.

PR #427, opened 2018-03-27
Don’t show HTML entities in plain text emails

quis:
Generally we shouldn’t be expecting HTML entities in our plain text emails, but _some_ teams are using `&nbsp;` as a workaround to get multiple newlines.

We don’t want to support multiple newlines officially, but they should be able to use this workaround without it sending garbage to plain text clients.


1pretz1:
Nice work, this will solve our plain text email `&nbsp;` appearing issue. Thanks!

PR #437, opened 2018-04-03
Make `Take` class inherit from `str`

quis:
Since:
- we only instantiate this class with a string value
- we always coerce to a string when using the result of this class

Doing this simplifies the class itself (fewer lines), and the interface (by allowing us to remove the `.as_string` property).


notifications_utils/take.py
```diff
@@ -1,22 +1,15 @@
...
-class Take():
-
-    def __init__(self, thing):
-        self.thing = thing
+class Take(str):
 
     @classmethod
     def from_field(cls, field):
-        return cls(str(field))
+        return cls(field)
```
leohemsted: does this from_field give us anything now?

as the casting is implicit now, so 

```
Take.from_field(my_field) == Take(my_field)
```

quis: Good point. We can lose `.from_field` too. See a072fb2.

quis:
@leohemsted Major version bump added in aa3f251.

PR #467, opened 2018-04-24
add zendesk client

leohemsted:
looks similar to the existing deskpro client.

* priority is set to "normal" unless P1, when it's "urgent". Don't know if we want to give more granularity over this
* type is passed in - can be `problem`, `incident`, `question`, or `task`. No default.

see details about zendesk API here: https://developer.zendesk.com/rest_api/docs/core/tickets


notifications_utils/clients/zendesk/zendesk_client.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # if no requester provided, then the call came from within Notify 👻
+        if user_email:
+            data['ticket']['requester'] = {
+                'email': user_email,
+                'name': user_name or '(no name supplied)'
+            }
+
+        response = requests.post(
+            'https://govuk.zendesk.com/api/v2/tickets.json',
```
quis: Feels weird to not have this split out into a constant like `NOTIFY_GROUP_ID`.

notifications_utils/clients/zendesk/zendesk_client.py
```diff
@@ -0,0 +1,88 @@
...
+        user_name=None,
+        user_email=None
+    ):
+        data = {
+            'ticket': {
+                'subject': subject,
+                'comment': {
+                    'body': message
+                },
+                'group_id': ZendeskClient.NOTIFY_GROUP_ID,
```
quis: `self.NOTIFY_GROUP_ID` is shorter (and better if we ever want to subclass this client).

PR #471, opened 2018-04-30
strip obscure whitespace on all emails

leohemsted:
see https://github.com/alphagov/notifications-admin/commit/9e78c5f5

"However we are still getting some bad email addresses through, for
example one that had a leading zero-width space character. This means
that the user sees a validation error; really we should just deal with
the mess for them.

So this commit also includes characters without Unicode character
property "WSpace=Y" (which includes zero-width space) to those which are
stripped from form submissions.

List taken from here: https://en.wikipedia.org/wiki/Whitespace_character

See issue and discussion here: https://bugs.python.org/issue13391"


leohemsted:
@quis added and squashed in

PR #476, opened 2018-05-10
Fix multiple lines separated by single line breaks

quis:
If you have this in your template:
```
A
B
C
```

You (unexpectedly) get:
```
A
B

C
```

This is because we use a hidden element with 0 height to represent line breaks (illustrated here by `—`):
```
A
—
B
—
C
```

However if we have two single linebreaks following each other then they should occupy space. And because the `B` and `C` in our example are text nodes, not HTML elements the browser considers the line breaks before and after `B` to be consecutive (ie it doesn’t ‘see’ the `B`).

So the second line break gets height applied, like this:
```
A
—
B
▢
C
```

---

This commit fixes the problem by using real line breaks (`<br>`) and paragraphs (`<p>`). The reason we had to stop using these before was for exact compatibility with DVLA (see c3dd782c016906b845c3aa9458687af3022992e3).

There is a bit of CSS trickery to hide the first and last `<br>`s inside a paragraph; the fact that a sequence of `<br>` get wrapped in a paragraph means that they take up two extra lines of vertical space (because the paragraph makes a newline and has a bottom margin).


quis:
Doesn’t seem to change paragraph spacing of existing letters:

letter-compare

quis:
Reverting this until we also have proper page breaks.

PR #482, opened 2018-05-15
Add S3 download method to utils

quis:
It has an upload method. A download method will be useful in:

- admin app (for downloading CSV files)
- template preview (for downloading cached letters)


notifications_utils/s3.py
```diff
@@ -39,3 +39,16 @@ def s3upload(filedata, region, bucket_name, file_location, content_type='binary/
         put_args['Tagging'] = tags
 
     key.put(**put_args)
+
+
+class S3ObjectNotFound(botocore.exceptions.ClientError):
+    pass
+
+
+def s3download(bucket_name, filename, raises=True):
```
allait: `raises` doesn't seem to be used

quis: squashed

PR #491, opened 2018-06-15
add Base64UUIDConverter for flask routes

leohemsted:
(and a couple of generic byte <-> base64 fns)

This enables us to have URLs which contain UUIDs (36 characters), but with only 22 characters used. Shorter! Nicer! Easier to get into a text message!


leohemsted:
closed in favour of https://github.com/alphagov/notifications-utils/pull/495

PR #494, opened 2018-06-20
change email local validation to be whitelist-based

leohemsted:
rather than blacklisting invalid characters, just whitelist the characters that are explicitly allowed.

Note: This rejects technically correct emails with quoted local parts, but that's okay. I don't think anyone actually uses one of them.


notifications_utils/recipients.py
```diff
@@ -448,11 +448,12 @@ def validate_email_address(email_address, column=None):  # noqa (C901 too comple
     if not match:
         raise InvalidEmailError
 
-    hostname = match.group(1)
-    # don't allow consecutive periods in domain names
-    if '..' in hostname:
+    # don't allow consecutive periods in either part
+    if '..' in email_address:
```
sakisv: Maybe use a regex like 

```python
r"\.{2,}"
```

to also catch more than 2 dots.

notifications_utils/recipients.py
```diff
@@ -44,10 +44,10 @@
 }
 
 # regexes for use in validate_email_address.
-# invalid local chars - whitespace, quotes and apostrophes, semicolons and colons, GBP sign
+# Valid characters taken from https://en.wikipedia.org/wiki/Email_address#Local-part
 # Note: Normal apostrophe eg `Firstname-o'surname@domain.com` is allowed.
-INVALID_LOCAL_CHARS = r"\s\",;:@£“”‘’"
-email_regex = re.compile(r'^[^{}]+@([^.@][^@]+)$'.format(INVALID_LOCAL_CHARS))
+VALID_LOCAL_CHARS = r"a-zA-Z0-9.!#$%&'*+/=?^_`{|}~\-"
```
sakisv: TIL that all these symbols are valid in the address 😱 🤔 

Edit:
However, reading the link above it says that some of them are only valid if quoted. Probably not something to be implemented now, but useful to keep in mind for a future iteration as and when necessary.

leohemsted: Shouldn't be! Which ones did you see? 

wikipedia lists this one `#!$%&'*+-/=?^_`{}|~@example.org` as a completely valid unquoted address.

notifications_utils/recipients.py
leohemsted: this'll catch more than 2 dots - it returns True after the first time it sees two dots

sakisv: Based on this https://regexr.com/3rb59 it matches 2 or more dots which I think it's what we're after?

notifications_utils/recipients.py
sakisv: Ah you're right; I read the bit below, but didn't notice the different character subset:

> space and "(),:;<>@[\] characters are allowed with restrictions (they are only allowed inside a quoted string, as described in the paragraph below, and in addition, a backslash or double-quote must be preceded by a backslash);

However it says that `..` is allowed if quoted:

> dot ., provided that it is not the first or last character unless quoted, and provided also that it does not appear consecutively unless quoted (e.g. John..Doe@example.com is not allowed but "John..Doe"@example.com is allowed);[6]

notifications_utils/recipients.py
leohemsted: ah this isn't a regex this is a simple python substring check. Should be more performant than a regex (worst case O(N))

notifications_utils/recipients.py
leohemsted: yeah i decided not to allow quoted local parts because i don't think i've ever seen an actual email address

PR #499, opened 2018-07-13
Add Microsort-specific spacing styles

tombye:
One of our users (using Outlook 2016 on Windows 10) reported issues with the top banner of our email template when set up with their branding:
- the spacing between the brand image and text was missing
- the column of the main content area was a lot larger than that for the banner

We have been unable to replicate this when testing with Email on Acid. The problems described can be seen with other combinations of email client and OS though. These fixes work on those versions (and don't break any existing code) so it's worth deploying them to see if they have an effect for this user.

The fixes are:
- replacing the padding and margin between the brand image and text with a spacer cell/image combo
- making higher DPI systems pick up our width styles by duplicating those set in HTML attributes to inline CSS
- adding some Microsoft Office-specific CSS which can remove spacing Outlook sets sometimes

I also noticed the blue bar below the GOV.UK banner was too big in Outlook due to the wrong width being set in HTML wrapped in Outlook-specific conditional comments. A fix for this is also included.

### DIFF
--- a/notifications_utils/jinja_templates/email_template.jinja2
+++ b/notifications_utils/jinja_templates/email_template.jinja2
@@ -142,7 +142,7 @@
                   <img
                     src="{{ brand_logo }}"
                     border="0"
-                    style="display: block; border: 0; margin-top: 10px; margin-bottom: 12px;"
+                    style="display: block; border: 0; Margin-top: 10px; Margin-bottom: 12px;"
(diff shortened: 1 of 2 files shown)


tombye:
I think there are better ways to control spacing so I'm closing this for now as I don't want to add any code that doesn't improve things.

PR #507, opened 2018-07-30
Recognise unicode bullets as list item signifiers

quis:
Looks like people are copy-pasting content from Word or similar into their Notify templates.

When this is happening they’re getting unicode bullet points, which our markdown parser doesn’t recognise as the start of a list item. So they get something that looks a bit like a list, but without the proper spacing or indentation.

This commit amends the regular expressions used to detect lists to also recognise bullet characters (and makes the tests a bit more comprehensive in the process).

# Markdown

image

# Output before 

image

# Output after 

image


quis:
This is the only letter template with bullet characters: https://www.notifications.service.gov.uk/services/d578c0f5-7e0a-4545-8383-eada108993ea/templates/391b9b87-c403-4676-9b37-dc83b0597213

It will change the layout slightly, but won’t cause re-pagination.

PR #515, opened 2018-08-16
Increase sms limit

klssmith:
Increased the SMS character count limit from 459 characters (3 text messages) to 612 characters (4 text messages). 

Bumped the version from 30.0.1 to 30.1.0 as a result.

[Pivotal story](https://www.pivotaltracker.com/story/show/159695634)


klssmith:
👍 Changed the version to 30.1.0

PR #518, opened 2018-08-20
Limit the number of pages previewed in a letter

quis:
At the moment the admin app will let you preview as many pages as your letter is long. This has some performance _implications_.

It’s also not very helpful because it makes it look like your 13 page letter will get printed.

So this commit:
- adds an upper limit on the number of displayed pages (10, same as we can print)
- adds an error message if your letter is longer than this (which won’t stop you making a letter that’s too long, but is better than nothing)

***

image

***

image


notifications_utils/template.py
```diff
@@ -547,10 +548,20 @@ def __init__(
         self.image_url = image_url
         self.page_count = int(page_count)
 
+    @property
+    def page_numbers(self):
+        for i in range(min(self.page_count, self.max_page_count)):
+            yield i + 1
```
CrystalPea: I don't understand this function, what does it do?

notifications_utils/jinja_templates/letter_image_template.jinja2
```diff
@@ -1,3 +1,14 @@
+{% if too_many_pages %}
+  <div class='banner-dangerous bottom-gutter'>
+    <h1 class='banner-title'>
+      This letter is too long
+    </h1>
+    <p>
+      Letters can’t have more than {{ page_numbers[-1] }} pages
```
karlchillmaid: GOV.UK style guide advises against negative contractions: https://www.gov.uk/guidance/content-design/writing-for-gov-uk#contractions
I think it's a bit more nuanced than that – context is important – but we probably should avoid them in error messages, where the instruction must be explicit.

Suggestion:
`Letters cannot be more than {{ page_numbers [-1] }} pages long`

notifications_utils/jinja_templates/letter_image_template.jinja2
```diff
@@ -1,3 +1,14 @@
+{% if too_many_pages %}
+  <div class='banner-dangerous bottom-gutter'>
+    <h1 class='banner-title'>
+      This letter is too long
```
karlchillmaid: This h1 banner title more or less repeats the description.
Suggest we use:
`There is a problem`
as described in the design system: https://design-system.service.gov.uk/components/error-summary/

I'm working with the other GaaP content designer to try and make our approach to error messages more consistent across Notify, Pay and Registers.

notifications_utils/template.py
quis: Would it be clearer like this?
```python
last_page_number = min(self.page_count, self.max_page_count)
return range(1, last_page_number)
```

notifications_utils/jinja_templates/letter_image_template.jinja2
quis: I strongly feel that being specific and direct is better here.

The design system has to use ‘There is a problem’ because it’s dealing, potentially, with multiple errors (so should really be ‘There are some problems’).

notifications_utils/template.py
CrystalPea: yes, much clearer to me, thank you! :)

notifications_utils/jinja_templates/letter_image_template.jinja2
quis: image

notifications_utils/jinja_templates/letter_image_template.jinja2
karlchillmaid: I think you're right that a specific and direct error summary is better in cases where the page can only display a single error.

If we keep `This letter is too long` as the summary we could change the description to avoid the repetition, eg:

`Letters must be 10 pages or less`

quis: image

2886e0d

PR #523, opened 2018-09-03
Add a preheader to HTML email for nicer snippets

quis:
Email clients take the first couple of sentences<sup>1</sup> of an email’s content and use it to show a preview of the email in the inbox. This is useful, but we are making it less useful because the first bit of
content in our emails is:
- GOV.UK (if the email has GOV.UK branding)
- the name of the organisation (for single identity brands)

Worse, some email clients (looking at you Outlook) display the URL of any images as part of this preview. This is URL is not useful or informative.

We can get some control over the text of this preview by making sure that the first content in the `<body>` of the message is what we’d want in the preview. So this commit insert the first 256 characters of the content of the message before any branding. Then it adds some CSS to hide this content from anyone looking at the full email.

This approach is based on the recommendations in
https://www.campaignmonitor.com/blog/email-marketing/2011/12/a-practical-guide-to-email-preheaders/

1. In Outlook on a wide monitor this can reach 250 characters, which is  how I arrived at the 256 character truncation.

# Before 

image

# After 

image


quis:
Should also stop this happening in Outlook:

image

quis:
Tested that the preheader is not showing when opening the full email in any client:
https://app.emailonacid.com/app/acidtest/RPVQoXoi91PxQRWVrWOppqyFy73WAzgviBcYsqRehSgx0/list

PR #537, opened 2018-10-03
Add antivirus scan to utils repository

CrystalPea:
This code has been copied (and adjusted in the process) from
document-download-api repo, since now we also need to use it
in notifications-admin, so we need to have it in a place that
will be accessible for both those apps.

Next step will be to make make document-download-api call
the antivirus client from here and to delete duplicate code from
there.

### Link to code in document-download-api: 
**client:** https://github.com/alphagov/document-download-api/blob/master/app/utils/antivirus.py
**tests:** https://github.com/alphagov/document-download-api/blob/master/tests/utils/test_antivirus.py

Pivotal ticket: https://www.pivotaltracker.com/story/show/159697071


servingUpAces:
We must remember to add the new config to admin app

PR #540, opened 2018-10-09
Encode full stops in tokens we put in URLs

quis:
When we send confirmation emails they have a tokenised URL in them. Sometimes this token contains full stops, for example:
> http://example.com/confirm/3nfu34h34.g34ug34ug

It looks like some email filters interpret this as a link to a file with the extension `.g34ug34ug`. Since this is an unknown file extension they filter out the email.

Our testing has shown that URL-encoding the full stop seems to get around this problem. Any browser will take this URL and decode the full stop before sending a request to our servers, so this should Just Work.


notifications_utils/formatters.py
```diff
@@ -134,6 +136,10 @@ def strip_dvla_markup(value):
     return re.sub(dvla_markup_tags, '', value)
 
 
+def url_encode_full_stops(value):
+    return full_stops.sub('%2E', value)
```
allait: You could do `value.replace('.', '%2E')` here to make it a bit simpler since we're not going to add anything else we encode to '%2E' to that regex

PR #542, opened 2018-10-12
Accessibility fixes

tombye:
Fixes for problems experienced by users of the NVDA screenreader when viewing our emails on Outlook 2016 and Mail on Windows 10. We don't have Outlook on a physical machine so I've been testing with Mail on the Windows 10 machine in the empathy lab.

## Extra spaces

When you move through emails sentence-by-sentence in NVDA, 4 spaces are announced between the banner at the top of the email and the content.

### Fixes

After a lot of testing it seems the spaces came from one of:
- `&nbsp;` characters used in table cells to ensure they retain a certain dimension
- the table we use for the blue bar below the GOV.UK banner

This removes all `&nbsp;` characters. Removing the table would require a lot more restructuring of the template so I'm leaving it out of this pull request.

## Presentational images being announced

Images marked as presentational, by giving them `alt=""`, were having their URLs announced by both NVDA and JAWS.

### Fixes

After a testing with various combinations of characters in the alt text, it seems a single space means the images are ignored in JAWS and announced in NVDA as 'graphic'.

This isn't a perfect solution but is a lot better than having a URL announced. It also gives some information about what you're reading. URLs can easily be interpreted as links.

The fix also includes the removal of all images used as invisible 'spacers' as their being announced has no value to users.

## Testing

I've checked the email template with all variations of branding, with fixes in, in all emailonacid and with Mail on Windows 10 using NVDA.

## Tickets

This contains fixes for issues raised in these tickets:
- https://www.pivotaltracker.com/story/show/160887330
- https://www.pivotaltracker.com/story/show/161183433


notifications_utils/jinja_templates/email_template.jinja2
```diff
@@ -149,19 +149,14 @@
                 </td>
               {% endif %}
               {% if brand_name %}
-                {% if brand_logo %}
-                <td width="5">{{ spacer(5) }}</td>
-                {% endif %}
-                <td width="5">{{ spacer(5) }}</td>
-                <td width="100%" bgcolor="{{brand_colour}}" valign="middle" align="left">
+                <td width="100%" bgcolor="{{brand_colour}}" valign="middle" align="left" style="padding: 14px 0 14px 10px">
```
quis: Confirm that the left-hand spacing here should always be `10px`. That the previous code had different left-hand spacing values<sup>1</sup> for with and without an image is wrong.

1. `10px` with an image, `5px` without an image

notifications_utils/jinja_templates/email_template.jinja2
```diff
@@ -199,25 +194,23 @@
...
-                    <td bgcolor="{{ brand_colour }}" width="2">{{ spacer(2) }}</td>
-                    <td width="7">{{ spacer(7) }}</td>
-                    {% endif %}
-                    <td><img src="{{ brand_logo }}" style="display: block; border: 0;" height="{% if brand_name -%} 27 {%- else -%} 54 {%- endif %}" alt=" " /></td>
-                    <td width="5">{{ spacer(5) }}</td>
-                    <td width="100%" style="font-family: Helvetica, Arial, sans-serif; font-size: 18px; line-height: 23px;" valign="center">
-                      {% if brand_name %}
-                      {{ brand_name }}
-                      {% endif %}
+                  <tr>
```
quis: Duplicate opening `<tr>` tag.

notifications_utils/jinja_templates/email_template.jinja2
```diff
@@ -199,25 +194,23 @@
...
                 <table role="presentation" width="100%" cellpadding="0" cellspacing="0" border="0" style="border-collapse: collapse">
                   <tr>
-                    {% if brand_colour %}
-                    <td height="10" width="1">{{ spacer(1, 10) }}</td>
-                    <td width="5">{{ spacer(1, 10) }}</td>
-                    {% endif %}
-                    <td height="10">{{ spacer(1, 10) }}</td>
-                    <td height="10">{{ spacer(1, 10) }}</td>
-                    <td height="10">{{ spacer(1, 10) }}</td>
+                    <td height="24" width="100%" colspan="{% if brand_colour %}3{% else %}2{% endif %}"><br /></td>
```
quis: `colspan` here should always be `2` – [this `<tr>`](https://github.com/alphagov/notifications-utils/pull/542/commits/d4a12691b3ad09c67c1db45a751e6680ff567da6#diff-7d3ef7fa9c747e8e757790bb586383e0R203) only ever has 2 columns.

tombye:
I rebased against the latest `master` branch so the history is a bit out on this, in favour of nicer history long-term.

PR #548, opened 2018-10-24
Change email validation to remove obscure whitespace from addresses

klssmith:
We were stripping all types of whitespace from the ends of email
addresses, but ignoring whitespace in the middle. We recently saw an
error where there was a zero-width whitespace in an email domain. This
commit changes the validation to now remove obscure whitespace characters
from the middle of email addresses, since users will not be able to see
that these are there. (All whitespace at the start and end of email addresses
will still be stripped.)

[Pivotal story](https://www.pivotaltracker.com/story/show/161290942)


tests/test_formatters.py
```diff
@@ -982,6 +983,15 @@ def test_strip_whitespace(value):
     assert strip_whitespace(value) == 'bar'
 
 
+@pytest.mark.parametrize('value', [
+    'notifications-email',
+    '  \tnotifications-email \x0c ',
+    '\rn\u200Coti\u200Dfi\u200Bcati\u2060ons-\u180Eemai\uFEFFl\uFEFF',
```
quis: Is it worth testing a counter example? i.e. that a word with normal spaces or newlines doesn’t have those spaces removed.

klssmith: 👍 I've added a test to check that 'normal spaces' get removed from the ends, but not the middle of a string

quis: 👍

PR #553, opened 2018-11-09
Don't log proxy header warnings if the check is disabled

allait:
### Don't log proxy header warnings if the check is disabled

Header warnings on each requests make it very hard to find actual warnings. Since the request headers are going away when we switch to CDN we're not planning to switch them on for any new apps.

### Log an error for log messages with unused formatting placeholders

If a log message has a string with placeholders (for example if the application code never called `.format`) nothing is logged since `.format` raises an unhandled exception that doesn't go through JSON formatting.

Currently, we handle the case with named placeholders but not positional ones.

### Update requests to fix CVE-2018-18074

> The Requests package through 2.19.1 before 2018-09-14 for Python sends an HTTP Authorization header to an http URI upon receiving a same-hostname https-to-http redirect, which makes it easier for remote attackers to discover credentials by sniffing the network.


sakisv:
Completely missed this, sorry for the delay 😓

PR #557, opened 2018-11-21
Add function to check if letter can be cancelled

klssmith:
We need a way of telling whether a letter can be cancelled by a user or not in both `notifications-admin` and `notifcations-api`, so this adds the a function which can be used in both places.

* In `notifications-admin` this will get used to determine whether the link to cancel should be shown on the notification page.
* In `notifications-api` this will get used as a final check before cancelling a letter.


notifications_utils/letter_timings.py
```diff
@@ -68,3 +68,36 @@ def get_next_royal_mail_working_day(date):
...
+        return False
+
+    return True
+
+
+def _is_after_half_five_local_time():
+    current_utc_datetime = datetime.utcnow()
+    bst_time = convert_utc_to_bst(current_utc_datetime).time()
+
+    return bst_time >= time(17, 30, 0, 0)
```
allait: API has a config variable for this called `LETTER_PROCESSING_DEADLINE`. If this is going to be used across admin and the API maybe we should define it as a constant in this module and use that in the API instead.

klssmith: 👍 Now updated

PR #571, opened 2019-01-16
Set links to use govuk-blue

edwardhorsford:
Links are currently not given an explicit colour, whilst text is.

This explicitly sets them to use `$govuk-blue`. 

NB this doesn't set `active` or `visited` states. These would be trickier to support, and appear to not have great support anyway:
https://www.campaignmonitor.com/css/selectors/active/
https://www.campaignmonitor.com/css/selectors/visited/


quis:
@edwardhorsford Thanks for your work on this. It was deployed to production on Friday 🎉

PR #582, opened 2019-02-05
Display letter postage in preview of template

quis:
This adds an extra HTML element our preview of letters.

It adds an extra element to the first page of the letter which has:
- a class which will allow us to style first/second class letters differently
- some text which we can hide with CSS, but which will be useful to screen reader users


quis:
@CrystalPea Updated to be a breaking change because if the admin app pulls in this code without any other changes then it will (incorrectly) display the postage as second class for every letter.

CrystalPea:
LGTM 👍

PR #584, opened 2019-02-11
allow letter logo_file_name to be None

leohemsted:
it'll just show a blank space there by keeping the spacing div and just removing the logo img tag

image


notifications_utils/template.py
```diff
@@ -440,7 +440,8 @@ def __str__(self):
         return Markup(self.jinja_template.render({
             'admin_base_url': self.admin_base_url,
             'logo_file_name': self.logo_file_name,
-            'logo_class': 'svg' if self.logo_file_name.lower().endswith('svg') else 'png',
+            # logo_class should only ever be None, svg or png
```
quis: Is it worth putting this in the `__init__`?

eg
```python
self.logo_class = logo_file_name.lower()[-3:] if logo_file_name else None
if self.logo_class not in {'png', 'svg', None}:
    raise TypeError("Unsupported file type for letter logo")

leohemsted: so i think this isn't a concern any more since the flow is more controlled now - admin validates that the file you upload is an svg, and then we generate the png ourselves

PR #602, opened 2019-04-16
 Change when RecipientCSV 'get_rows' method gets called

klssmith:
The RecipientCSV had a `rows` attribute which generated all the rows when the class was instantiated. This commit changes `rows` to be a property so that the `get_rows` generator only gets called when it is needed.

*Commits are appearing in the wrong order below (due to rebasing) but are in the right order if you check out the branch locally*


quis:
I think the idea behind this is solid, but it could be refactored a bit: c62adba

tests/test_recipient_csv.py
```diff
@@ -162,6 +162,36 @@ def test_get_rows(file_contents, template_type, expected):
...
+
+    rows = recipients.get_rows()
+    for i in range(3):
+        next(rows)
+
+    assert not has_error_mock.called
+    assert not has_bad_recipient_mock.called
+    assert not has_missing_data_mock.called
+    assert not cell_recipient_error_mock.called
+
```
quis: Probably worth testing here that we’re only instantiating 3 `Row` objects (ie we’re not iterating over the file multiple times).

tests/test_recipient_csv.py
```diff
@@ -162,6 +162,36 @@ def test_get_rows(file_contents, template_type, expected):
...
+    )
+
+    has_error_mock = mocker.patch.object(Row, 'has_error')
+    has_bad_recipient_mock = mocker.patch.object(Row, 'has_bad_recipient')
+    has_missing_data_mock = mocker.patch.object(Row, 'has_missing_data')
+    cell_recipient_error_mock = mocker.patch.object(Cell, 'recipient_error')
+
+    rows = recipients.get_rows()
+    for i in range(3):
+        next(rows)
```
quis: This would be a more realistic test of what the code in the API is doing:
```suggestion
        assert next(rows).recipient == 'a@b.com'
```

tests/test_recipient_csv.py
klssmith: I couldn't find a way to test this in the existing test, so I've added a new test for this below the current one

klssmith:
Thanks for the suggestions - I've made the changes to the tests, and cherry-picked your refactoring commit into this branch

notifications_utils/recipients.py
```diff
@@ -177,15 +187,20 @@ def get_rows(self):
...
+                row = Row(
                     output_dict,
                     index=index,
                     error_fn=self._get_error_for_field,
                     recipient_column_headers=self.recipient_column_headers,
                     placeholders=self.placeholders_as_column_keys,
                     template=self.template,
                 )
+                if rows_as_a_list_is_empty:
+                    self.rows_as_list.append(row)
```
quis: You don’t need to build up the list here (and there’s a risk that you’ll end up building a partial list).

If this method doesn’t build up the list then the API can use it as a pure generator, which should reduce memory usage.

tests/test_recipient_csv.py
```diff
@@ -162,6 +162,59 @@ def test_get_rows(file_contents, template_type, expected):
...
+        """,
+        template_type='email',
+        placeholders=['name'],
+    )
+
+    rows = recipients.get_rows()
+    for i in range(3):
+        next(rows)
+
+    assert row_mock.call_count == 3
```
quis: Should test to make sure that when you only call `get_rows()` it doesn’t build up a list in memory:
```suggestion
    assert row_mock.call_count == 3
    assert recipients.rows_as_list == None
```

klssmith:
I've tried this out with a 50,000 row CSV file to compare the timings of doing things the old way (using the code on the master branch and calling `.rows`) vs the new way (the code on this branch, using `.get_rows()`).

| | Old way | New way |
--- | --- | --- 
| Average time taken to call `.rows` / `.get_rows()`| 3.999 seconds | 0.033 seconds |
| Average time taken to iterate through 50,000 rows | 4.090 seconds | 3.880 seconds |

Iterating through lots of rows is (very) slightly faster, but the main change would be that we start processing the rows a couple of seconds sooner in notifications-api.

PR #606, opened 2019-05-07
Support welsh characters in SMS

klssmith:
**Allowed Welsh characters not already included in GSM to be used in SMS**
- Changed `SanitiseGSM` class so that it allows Welsh characters which are not supported in the default GSM character set (ÂâÊêÎîÔôÛûŴŵŶŷ).
- Renamed some code to make it clear that Welsh characters are allowed in addition to GSM characters.
- Replaced  **`** (which is not a valid GSM character) in the list of allowed GSM characters with **§** (which is a valid GSM character).

**Changed how fragment count is calculated**
When we get the SMS fragment count for an `SMSMessageTemplate` we now check if we should use the fragment count for a GSM or unicode message based on whether or not the message contains and non-GSM Welsh characters.

[Pivotal story](https://www.pivotaltracker.com/story/show/165672985)


notifications_utils/sanitise_text.py
```diff
@@ -96,9 +100,11 @@ class SanitiseGSM(SanitiseText):
     """
     ALLOWED_CHARACTERS = set(
         '@£$¥èéùìòÇ\nØø\rÅåΔ_ΦΓΛΩΠΨΣΘΞ\x1bÆæßÉ !"#¤%&\'()*+,-./0123456789:;<=>?' +
-        '¡ABCDEFGHIJKLMNOPQRSTUVWXYZÄÖÑÜ`¿abcdefghijklmnopqrstuvwxyzäöñüà' +
+        '¡ABCDEFGHIJKLMNOPQRSTUVWXYZÄÖÑÜ§¿abcdefghijklmnopqrstuvwxyzäöñüà' +
```
leohemsted: any reason ` ` ` has been replaced with `

leohemsted: any reason `` ` `` has been replaced with `§`?

notifications_utils/template.py
```diff
@@ -168,12 +169,15 @@ def content_count(self):
...
             if self._values
-            else gsm_encode(add_prefix(self.content.strip(), self.prefix))
+            else gsm_and_welsh_encode(add_prefix(self.content.strip(), self.prefix))
         ).encode(self.encoding))
 
     @property
     def fragment_count(self):
-        return get_sms_fragment_count(self.content_count)
+        content_with_placeholders = str(self)
+        unicode = is_unicode(content_with_placeholders)
```
leohemsted: `unicode` is a built-in, it's good practice not to overwrite it. i'd probably just in-line this into the `get_sms_fragment_count` call.

notifications_utils/template.py
```diff
@@ -620,8 +624,18 @@ def __init__(self, keys):
         super(NoPlaceholderForDataError, self).__init__(", ".join(keys))
 
 
-def get_sms_fragment_count(character_count):
-    return 1 if character_count <= 160 else math.ceil(float(character_count) / 153)
+def get_sms_fragment_count(character_count, unicode):
+    if unicode:
```
leohemsted: can change this to `is_unicode` for the same overwriting reason

notifications_utils/template.py
```diff
@@ -620,8 +624,18 @@ def __init__(self, keys):
...
+        return 1 if character_count <= 70 else math.ceil(float(character_count) / 67)
+    else:
+        return 1 if character_count <= 160 else math.ceil(float(character_count) / 153)
+
+
+def is_unicode(content):
+    for char in SanitiseText.WELSH_NON_GSM_CHARACTERS:
+        if char in content:
+            return True
+    return False
```
leohemsted: can use set maths to clean make this more efficient, since it'll go only have to loop through `WELSH_NON_GSM_CHARACTERS` and `content` once each to construct the set, and then once through one of the sets to check intersection. (this would reduce complexity from O(n^2) to O(n))

```python
return set(content) & set(SanitiseText.WELSH_NON_GSM_CHARACTERS)
```

tests/test_sanitise_text.py
```diff
@@ -37,14 +36,16 @@ def test_encode_chars_the_same_for_ascii_and_gsm(char, expected, cls):
     (('\n', '\n', '?'), 'ascii control char in gsm (newline)'),
     (('\r', '\r', '?'), 'ascii control char in gsm (return)'),
     # These characters are present in GSM but not in ascii
+    (('â', 'â', 'a'), 'non-ascii gsm char (a with hat)'),
     (('à', 'à', 'a'), 'non-ascii gsm char (a with accent)'),
+    (('Ŷ', 'Ŷ', 'Y'), 'non-ascii gsm char (capital y with hat)'),
```
leohemsted: can you move these two added lines to the bottom and add a separate comment to say that they're not GSM but are welsh characters

notifications_utils/sanitise_text.py
```diff
@@ -17,6 +17,8 @@ class SanitiseText:
         '\t': ' ',  # TAB
     }
 
+    WELSH_NON_GSM_CHARACTERS = 'ÂâÊêÎîÔôÛûŴŵŶŷ'
```
leohemsted: i think this should be a set (same as `ALLOWED_CHARACTERS`)

leohemsted: also, I think it should be in the `SanitiseGSMandWelsh` class rather than the base class

notifications_utils/sanitise_text.py
```diff
@@ -78,13 +80,15 @@ def encode_char(cls, c):
             return c if c is not None else '?'
 
 
-class SanitiseGSM(SanitiseText):
+class SanitiseGSMandWelsh(SanitiseText):
```
leohemsted: As discussed, SanitiseSMS or SanitiseTextMessage or something might be a bit less of a mouthful

klssmith: 👍 Changed the name to `SanitiseSMS`

notifications_utils/template.py
klssmith: I've changed this and the other place I'd used `unicode`

notifications_utils/sanitise_text.py
klssmith: All the GSM character lists I came across have `§` in instead of `` ` `` (there are no other differences between our list and the ones I found online). So I guessed that we had a small mistake in our list?

notifications_utils/sanitise_text.py
klssmith: 👍I've changed this to be a set in the `SanitiseSMS` class

tests/test_sanitise_text.py
```diff
@@ -72,16 +74,16 @@ def test_get_unicode_char_from_codepoint_rejects_bad_input(bad_input):
...
 
 @pytest.mark.parametrize('content, cls, expected', [
-    ('The quick brown fox jumps over the lazy dog', SanitiseGSM, set()),
-    ('The “quick” brown fox has some downgradable characters\xa0', SanitiseGSM, set()),
-    ('Need more 🐮🔔', SanitiseGSM, {'🐮', '🔔'}),
-    ('Lots of GSM chars that arent ascii compatible:\n\r€', SanitiseGSM, set()),
+    ('The quick brown fox jumps over the lazy dog', SanitiseSMS, set()),
+    ('The “quick” brown fox has some downgradable characters\xa0', SanitiseSMS, set()),
+    ('Need more 🐮🔔', SanitiseSMS, {'🐮', '🔔'}),
+    ('Lots of GSM chars that arent ascii compatible:\n\r€', SanitiseSMS, set()),
```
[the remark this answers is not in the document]
klssmith: Yep, I've added an extra line to this test

PR #607, opened 2019-05-08
 Check if contact detail value is not a list

CrystalPea:
If it's a list it means there are probably multiple columns for contact detail with data in one of the columns missing.
This would cause an exception if not caught.

Pivotal ticket: https://www.pivotaltracker.com/story/show/165768319

### Before:
Screen Shot 2019-05-08 at 15 23 19

### After: 
Screen Shot 2019-05-08 at 16 13 40


tests/test_recipient_csv.py
```diff
@@ -937,6 +937,28 @@ def test_multiple_sms_recipient_columns(international_sms):
     assert recipients.has_errors
 
 
+@pytest.mark.parametrize('international_sms', (True, False))
```
servingUpAces: I don't think international_sms adds anything to this test.
It would more interesting to know that the test passes if `phone number` has spaces, underscores, no spaces, etc.

CrystalPea: Thanks, that's a good suggestion, I tested for those cases and discovered some weirdness :D

CrystalPea: I added a comment to the test explaining the oddity. This is pre-existing behaviour, but we probably didn't know about it. It shouldn't affect our users.

PR #608, opened 2019-05-08
Break URLs onto a new line to stop very long lines

quis:
`word-break: break-all` tells the browser to break something wherever possible. There is also the non-standard `word-break: break-word` which says to only break a URL when it needs to (ie when it would otherwise overflow the whole line).

I think that `break-all` is better because:
- it doesn’t leave such big gaps in the text, so looks better
- is the only one that in that CSS standard, so has comprehensive browser support (see https://caniuse.com/word-break)

This only applies to links where the text of the link is a URL. Links where the template author is specifying the text of the link will continue to break in the normal way (ie like other text).

# Broken (no forced line breaks)

image

# `break-word` (not what this pull request does)

Screen Shot 2019-05-08 at 16 33 49

# `break-all` (what this pull request does)

image


quis:
Test results: https://app.emailonacid.com/app/acidtest/o3wvqRkLZBvnfTzHSYoOCURSu1OWUSG95sQlcJYx0bWPH/list

quis:
This doesn’t actually work (see test results). I’ve written up a story for the backlog so we can revisit this: https://www.pivotaltracker.com/story/show/167402427

PR #621, opened 2019-06-19
 Define letter statuses for which a letter job can be cancelled

CrystalPea:
Also add a condition that a letter / letter job cannot be cancelled
if it was created before 17:30 and it's already the next day now.


notifications_utils/letter_timings.py
```diff
@@ -84,9 +87,13 @@ def letter_can_be_cancelled(notification_status, notification_created_at):
     if notification_status not in ('created', 'pending-virus-check'):
         return False
 
-    if _after_letter_processing_deadline() and _notification_created_before_deadline(notification_created_at):
+    if _after_letter_processing_deadline() and _notification_created_before_today_deadline(notification_created_at):
         return False
 
+    if _notification_created_before_that_day_deadline(
+        notification_created_at
+    ) and notification_created_at.date() < datetime.utcnow().date():
```
servingUpAces: How can the `notification_created_at` be less than `now`?

CrystalPea: it's always less than now, past dates are smaller than present dates, no? So when date is yesterday it's smaller than today's date.

klssmith:
Do we need the extra check? If the letter was created before 5.30pm yesterday, then the notification status won't be `created` or `pending-virus-check`, so we will return `False` without checking the creation time.

The situation in the test of a letter created yesterday and still in `created` status should (hopefully!) be impossible.

CrystalPea:
@klssmith There are rare situations when that would happen. It's when our task sending letters to DVLA would not run and we need to re-run it manually the next date. In theory we could then allow our users to cancel their letters, but it could be confusing and also we would need to update our printing time messages in admin because right now those messages are not prepared for such a scenario.

notifications_utils/letter_timings.py
servingUpAces: Instead can we just do something like `notification_created_at < today + deadline time`?
It means that as long as the status is in allowable and the notification is before today's deadline then we can cancel the notification.

servingUpAces: Ok - I've finally wrapped my head around all the possibilities. This is probably for the best because we are only in a situation of having a letter that is not sent and from yesterday if the send to provider task failed. Since that task would likely be started at some point during the day, it's difficult to allow a letter to be cancelled while it is being sent. So it's best if we decline the cancellation.

PR #622, opened 2019-06-20
Keep a list of characters that cause tech failure

quis:
We know that if one of these characters appears in the hostname that SES will reject the email.

This causes a technical failure, which is something we should be keeping an eye on. If there’s noise in the technical failure logs then we won’t pay such close attention to them.

This commit changes the validation so that these email addresses will never even get to the point of being sent.


leohemsted:
i think its just a case of fixing the thing - there's an implementation bug.

hostnames have an incredibly strict list of characters - just `a-zA-Z0-9\-\.` so to give people full unicode domains, there's punycode, an encoding that takes your ünîcødę string and converts it into something that looks like `xn--ncd-vma5cu2s`. we aren't doing that when we send, and it's failing when it gets to SES. Quoth their docs:

> The email address string must be 7-bit ASCII. If you want to send to or from email addresses that contain Unicode characters in the domain part of an address, you must encode the domain using Punycode. Punycode is not permitted in the local part of the email address (the part before the @ sign) nor in the "friendly from" name.  _[ snip  ]_

we check that it's possible to encode it in punycode - but we don't actually do the conversion in `normalised_to` so that we can search on the original string. We should convert it to punycode before sending to SES.

That said, when i submit a test email through admin interface to `foo@vírgin.example.com` it converts to punycode there and then. but i can't figure out where in the code the conversion happens?

quis:
Closing as this isn’t the right solution (even if it’s not quite clear what is).

PR #633, opened 2019-07-29
AUTO: Refresh the statsd client every second

andy-paine:
- The statsd client uses UDP which means it doesn't know when the IP
  address it is sending to has disappeared
- We would like to run the statsd exporter on the PaaS so that we can
  co-locate it, scale it, monitor it easily etc. When the exporter gets redeployed on PaaS or PaaS have to re-run it, its IP address changes and the statsd clients in the apps sends the metrics into the void
- To make sure we pick up the new IP address when the exporter is redeployed, recreate the statsd client every 1 second which will refresh DNS
- I did a very basic test sending 500k stats with this client and a raw
  StatsClient and the performance was almost identical
- We are running this on the PaaS which means DNS is actually BOSH DNS
  which is much faster than normal DNS

### DIFF
--- a/notifications_utils/clients/statsd/statsd_client.py
+++ b/notifications_utils/clients/statsd/statsd_client.py
@@ -1,4 +1,5 @@
 from statsd import StatsClient
+from threading import Timer
 
 
 class StatsdClient():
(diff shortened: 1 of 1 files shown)


andy-paine:
Closing in favour of #634

PR #637, opened 2019-08-01
Allow more special characters to be used in SMS

klssmith:
This change allows any vowel, y or w with an accent, circumflex, acute
accent or grave accent to be used in text messages. Some of these
characters are GSM characters, but most are not included in GSM so will
be encoded using unicode.

[Pivotal story](https://www.pivotaltracker.com/story/show/167462423)


klssmith:
Closing, since we now plan to add support for more foreign languages than Welsh https://www.pivotaltracker.com/story/show/168031270

leohemsted:
This is coming back, as other foreign language support hasn't been scheduled yet but there's still a need for this.

PR #654, opened 2019-10-01
Remove error message for letter that is too long

CrystalPea:
This error will be instead defined in the notifications-admin app

We did this change so the CSS for this banner communicates with
styling for other elements on the page.


quis:
I think you can remove these lines as well:
https://github.com/alphagov/notifications-utils/blob/f73fbda03467e558f53224f2b5f901e12d19c858/notifications_utils/template.py#L606-L608

tests/test_pdf.py
```diff
@@ -32,6 +32,13 @@ def test_pdf_page_count_src_pdf_not_a_pdf():
         pdf_page_count(BytesIO(file_data))
 
 
+@pytest.mark.parametrize("page_count, expected_result", [
+    (None, False), (3, False), (11, True)
```
quis: Always best to test the boundary cases to make sure you haven’t made an off-by-one error:
```suggestion
    (None, False), (10, False), (11, True)
```

CrystalPea: Good idea, done ☺️

PR #668, opened 2019-11-15
don't encode sms when calculating content length

leohemsted:
we were previously encoding the utf-8 content, and measuring the number of bytes. We've been doing this since the beginning of time, and for most of time, it wasn't an issue. However, when we started accepting unicode characters (notably some welsh-only accents eg Ŵ or ŷ), they are encoded as two bytes rather than 1. So a message of 70 ŷ characters is 140 bytes long, but a message of 69 a characters and one ŷ is 71 characters long.

However, in reality, when we send a message, if _any_ character in it is non-gsm (similar but not the same as ascii), then the entire thing is encoded in UTF-16, where every single character (including basic latin text) is two bytes long. So both ŷŷŷŷ... and aaaaa....aŷ are actually 70 utf-16 characters - we already take account of the double code point width when determining billable fragments.

This means in a small amount of cases (where someone has a message that is almost at a message boundary in size, which contains a lot of welsh characters), we've over-billed that service. But this is probably an incredibly low amount of messages, and impossible to account for after 1 wk, so probably not worth worrying about.


notifications_utils/template.py
```diff
@@ -166,12 +166,12 @@ def prefix(self, value):
 
     @property
     def content_count(self):
```
idavidmcdonald: So just to spell out my understanding, this PR is changing `content_count` from instead of returning the number of bytes to return the number of characters. Originally, we've always wanted to return the number of characters and were assuming that the number of bytes was equivalent (although we were wrong). 

To make it clear what this method is purposed for, I'd like to see either the name of the method change to `content_character_count` (however realise this is more work and would mean changing things in other apps) or at the least add a doc string to explicitly say that we return the number of characters please.

notifications_utils/template.py
```diff
@@ -166,12 +166,12 @@ def prefix(self, value):
...
-        return len((
+        return len(
             # we always want to call SMSMessageTemplate.__str__ regardless of subclass, to avoid any html formatting
             SMSMessageTemplate.__str__(self)
             if self._values
             else sms_encode(add_prefix(self.content.strip(), self.prefix))
```
idavidmcdonald: Out of interest, this appears to be formatting and tidying up the content of the message in two different ways here, based on whether there is `self._values`. Could this behaviour be a contender to move into `SMSMessageTemplate.__str__` itself? It seems like it would be nice to have all message formatting be done in one place. By the way, not suggesting you change that here (if you think it seems sensible that is) - just checking how well I've understood the code

leohemsted: it doesn't quite work, unfortunately, because for placeholders that aren't filled it returns some HTML so that when they're rendered we highlight them in yellow (as seen when you send a one off message through the admin interface)
So in our tests, they fail because the count would be of `<span class='placeholder'>((placeholder))</span>` instead of `((placeholder))`. Perhaps the code could be refactored slightly but i'm a bit reluctant to prod this particular snake pit too much 😅

notifications_utils/template.py
leohemsted: yeah, your assumption is correct - bytes are irrelevant, and we made a bad assumption four years ago. in the ideal world, bytes should never have come into this, cos at no point have we or the providers ever encoded the unicode string into bytes.

It appears that the providers don't care between us sending `'Â' * 612` and sending `'a' * 612`. It's only when we're calculating billing information that we care about what characters are included, (the first is 10 fragments and the second is only 4) ie in [get_sms_fragment_count](https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/template.py#L628-L632)

i won't change the name partly because it's a relatively unnecessary breaking change, but also it's unneeded clarification as there's no other alternative way to count the content. I can add a docstring though.

leohemsted: semi-relatedly, we are using the content_count method elsewhere, but it does appear that we're checking `content_count > SMS_CHAR_COUNT_LIMIT` in three separate places in api/admin, all of which could be replaced with calls to `template.is_message_too_long`. (We still do call the property in a couple of other places)

notifications_utils/template.py
idavidmcdonald: Got ya!

PR #671, opened 2019-11-19
Content count for email and letter templates

CrystalPea:
Previously only Template subclass for sms messages had a content_count attribute. Now, as we check if notifications are not empty, we need to check content count of all types of messages.


tests/test_template_types.py
```diff
@@ -819,6 +819,35 @@ def test_subject_line_gets_replaced():
     assert template.subject == 'Jo'
 
 
+@pytest.mark.parametrize("content, values, expected_count", [
+    ("Content with ((placeholder))", {"placeholder": "something extra"}, 28),
+    ("Content with ((placeholder))", {"placeholder": ""}, 12),
+    ("((placeholder))  ", {"placeholder": "  "}, 0),
+    ("  ", {}, 0),
```
idavidmcdonald: Could be worth one additional test case for `("some string", {}, 11)` for the case of no stripping and no placeholder

tests/test_template_types.py
```diff
@@ -819,6 +819,35 @@ def test_subject_line_gets_replaced():
...
+    ("Content with ((placeholder))", {"placeholder": ""}, None, 12),
+    ("Just content", {}, None, 12),
+    ("((placeholder))  ", {"placeholder": "  "}, None, 0),
+    ("  ", {}, None, 0),
+    ("Content with ((placeholder))", {"placeholder": "something extra"}, "GDS", 33),
+    ("Just content", {}, "GDS", 12),
+    ("((placeholder))  ", {"placeholder": "  "}, "GDS", 5),
+    ("  ", {}, "GDS", 5),
+])
+def test_SMSMessageTemplate_character_count(content, values, prefix, expected_count):
```
idavidmcdonald: Turns out we already have a test for most of this (although I haven't compared the test cases themselves so there could be some that are missing).

Take a look at `test_get_character_count_of_content` in `test_base_template.py`

tests/test_template_types.py
```diff
@@ -838,9 +839,9 @@ def test_WithSubjectTemplate_character_count(content, values, expected_count):
     ("((placeholder))  ", {"placeholder": "  "}, None, 0),
     ("  ", {}, None, 0),
     ("Content with ((placeholder))", {"placeholder": "something extra"}, "GDS", 33),
-    ("Just content", {}, "GDS", 12),
-    ("((placeholder))  ", {"placeholder": "  "}, "GDS", 5),
-    ("  ", {}, "GDS", 5),
+    ("Just content", {}, "GDS", 17),
```
idavidmcdonald: This one meets my expectation of us sending the provider `GDS: Just content` and I don't see a code change in this commit that would change it from 12 characters to 17. Does that mean the test was failing before?

tests/test_template_types.py
```diff
@@ -838,9 +839,9 @@ def test_WithSubjectTemplate_character_count(content, values, expected_count):
...
+    ("((placeholder))  ", {"placeholder": "  "}, "GDS", 4),
+    ("  ", {}, "GDS", 4),
```
idavidmcdonald: Is this what we expect to send to the SMS providers. I was under the impression that we will send them
`GDS: ` which is 5 characters.

tests/test_template_types.py
```diff
@@ -838,9 +839,9 @@ def test_WithSubjectTemplate_character_count(content, values, expected_count):
     ("((placeholder))  ", {"placeholder": "  "}, None, 0),
     ("  ", {}, None, 0),
     ("Content with ((placeholder))", {"placeholder": "something extra"}, "GDS", 33),
-    ("Just content", {}, "GDS", 12),
-    ("((placeholder))  ", {"placeholder": "  "}, "GDS", 5),
-    ("  ", {}, "GDS", 5),
+    ("Just content", {}, "GDS", 17),
+    ("((placeholder))  ", {"placeholder": "  "}, "GDS", 4),
```
idavidmcdonald: Same as test case below, I would expect us to again send `GDS: `? Note, I may be incorrect, I'm not too familiar with this stuff. If you need a second opinion then Leo is a good person to speak to as he worked on this recently.

tests/test_template_types.py
CrystalPea: I think it's there at line 825

tests/test_template_types.py
CrystalPea: I'm not sure I understand... "Just content" is 12 characters, and "GDS: " is 5 characters, so 17 characters together, no?

I just added tests for this case because it was not tested before.

tests/test_template_types.py
CrystalPea: we strip the space at the end if there is nothing after it, that's why 4 characters, not 5.

But actually soon we won't send them a message like this at all :). Because the function that gets SMSMessageTemplate instance in api does not provide prefix as an argument, and without prefix that message is empty.

tests/test_template_types.py
CrystalPea: ahh, thanks, I will check 👍

PR #674, opened 2019-11-20
Preserve placeholder ordering in template change

quis:
Given some new template content like:
> Hello ((first name)) ((last name))

It’s more natural to, in the admin app, say ‘You added ((first name)) and ((last name))’ rather than ‘You added ((last name)) and ((first name))’. At the moment we can’t control which of the two messages the user will see (the order is non-deterministic).

Fixing this means preserving the order of the placeholders throughout the various classes that are used to calculate the difference in placeholders between two templates. The way we do this is by using `OrderedDict` and `OrderedSet` where we would otherwise use native `set` and `dict` objects. This is because `set` and `dict` don’t preserve insertion order, at least in the version of Python we’re using.


notifications_utils/columns.py
```diff
@@ -5,13 +5,29 @@
 
 class Columns(OrderedDict):
 
+    """
+    `Columns` behaves like an ordered dictionary, except it normalises
+    case, whitespace, hypens and underscores in keys.
+
+    In other words,
+    Columns({'FIRST_NAME': 'example'}) == Columns({'first name': 'example'})
```
idavidmcdonald: Great example

notifications_utils/recipients.py
```diff
@@ -289,15 +289,20 @@ def is_optional_address_column(self, key):
...
             if not self.is_optional_address_column(recipient_column)
-        ) <= set(
-            self.column_headers_as_column_keys
         )
 
+    @property
+    def has_recipient_columns(self):
+        # self.column_headers_as_column_keys is an OrderedSet. We need
+        # to cast it to a set because doing subset operations on
+        # OrderedSet returns `True` when you wouldn’t expected it to
```
idavidmcdonald: Optional: Maybe change to : `OrderedSet` has misleading behaviour when comparing it to `Set` objects

notifications_utils/recipients.py
```diff
@@ -289,15 +289,20 @@ def is_optional_address_column(self, key):
...
         return set(
             Columns.make_key(recipient_column)
             for recipient_column in self.recipient_column_headers
             if not self.is_optional_address_column(recipient_column)
-        ) <= set(
-            self.column_headers_as_column_keys
         )
 
+    @property
+    def has_recipient_columns(self):
```
idavidmcdonald: Optional: Maybe this method name could be better, like `contains_required_recipient_column_headers`?

quis: I’d prefer to keep this as-is, because it would be a breaking change to rename it, and the current name matches other properties like `has_errors`.

notifications_utils/recipients.py
quis: Done, and added a link to the relevant issue (squashed into 30741d9)

PR #676, opened 2019-11-26
Special characters

leohemsted:
https://github.com/alphagov/notifications-utils/pull/637 is back from the dead!

support all welsh characters, no other characters.

todo: test that it works with new mmg endpoint (specifically `Ò`)


notifications_utils/sanitise_text.py
```diff
@@ -96,15 +96,19 @@ class SanitiseSMS(SanitiseText):
...
     * any remaining unicode characters (eg chinese/cyrillic/glyphs/emoji) are replaced with ?
     """
-    # Welsh characters not already included in GSM
-    WELSH_NON_GSM_CHARACTERS = set('ÂâÊêÎîÔôÛûŴŵŶŷ')
+    WELSH_DIACRITICS = set(
+        'àèìòùẁỳ' 'ÀÈÌÒÙẀỲ'  # grave
+        'áéíóúẃý' 'ÁÉÍÓÚẂÝ'  # acute
+        'äëïöüẅÿ' 'ÄËÏÖÜẄŸ'  # diaeresis
+        'âêîôûŵŷ' 'ÂÊÎÔÛŴŶ'  # carets
+    )
```
quis: This is beautiful.

notifications_utils/template.py
```diff
@@ -645,14 +645,19 @@ def __init__(self, keys):
         super(NoPlaceholderForDataError, self).__init__(", ".join(keys))
 
 
-def get_sms_fragment_count(character_count, is_unicode):
-    if is_unicode:
+def get_sms_fragment_count(character_count, non_gsm_characters):
+    if len(non_gsm_characters) != 0:
```
idavidmcdonald: Optional: I believe that this could just be `if non_gsm_characters`, as it will always return a set (and the empty set is bool false).

Depends if you want to keep the `len` in particular for readability (or my above statement is false).

tests/test_template_types.py
```diff
@@ -861,51 +861,34 @@ def test_SMSMessageTemplate_character_count(
...
+        ('ÿ' * 71, 2),
+
+        ('ÿ' * 134, 2),
+        ('ÿ' * 135, 3),
+
+        ('ÿ' * 268, 4),
+        ('ÿ' * 269, 5),
+
+        ('à' * 70 + 'ÿ', 2),  # just one non-gsm character means it's sent at unicode
+        ('🚀' * 160, 1),  # non-welsh unicode characters are downgraded to gsm, so are only one fragment long
```
idavidmcdonald: Nice test coverage

PR #683, opened 2019-12-17
Add a docstring for the Field class

quis:


notifications_utils/field.py
```diff
@@ -49,6 +49,17 @@ def __repr__(self):
 
 
 class Field:
+
+    """
+    An instance of Field represents a string of text containing
```
idavidmcdonald: This is good. Might be worth clarifying that the placeholders may or may not be present e.g.

An instance of Field represents a string of text, which may or may not contain placeholders which can be replaced with values.

quis: Updated and squashed

PR #687, opened 2020-01-03
Lazy load letter preview images

quis:
Modern browsers now support lazy-loading of images<sup>1</sup>. Lazy loading means images won’t be loaded until they’re likely<sup>2</sup> to come into the viewport.

Doing this should:
- reduce the load on template preview
- reduce bandwidth usage
- improve page load times for our users

***

1. Currently Chrome and Android – https://caniuse.com/#feat=loading-lazy-attr
2. ‘The exact heuristics for ‘when the user scrolls near’ is left up to the browser’ – https://addyosmani.com/blog/lazy-loading/


quis:
Tested in Chrome 79 running locally with a 6 page letter:
- page 1 is still the first image to be loaded (as previously)
- page 2 and 3 load after the CSS images (not before, as previously)
- page 4 loads only once the user scrolls
- page 5 loads once the user has scrolled to page 2
- page 6 loads once the user has scrolled to page 3 

We couldn’t scroll to the bottom fast enough to see the images waiting to load (this is with the images cached in S3, so they only take ~300ms to be served)

PR #701, opened 2020-03-03
Exclude prefix in message count for SMS 

servingUpAces:
Update the is_message_too_long function on the SMSTemplate so that the message is length is calculated without the prefix.

As it turns out the message length was being validated differently for CSV uploads than for one-off messages and SMS sent via the API. Now all the validations on SMS message length should be the same.


servingUpAces:
I agree Leo, it's a bit of a tricky one. I do plan on updating the API to use this method to validate the message length. However, the existing methods already exclude the service name. Hence the decision to be more lenient. 
I'll take a look at our documentation to see if we can be more clear about how this works. We don't want to encourage long messages, but we also do not want to break existing templates.

idavidmcdonald:
Out of interest, does this mean we may charge users for a 5 fragment message when they may think they are sending a 4 fragment message? Does our billing code handle messages that are 5 fragments long?

PR #703, opened 2020-03-10
Add country normalisation and validation utilities

quis:
This pull request adds some code to:
- check that a country is real<sup>1</sup>
- normalise it to the ‘correct’ name<sup>2</sup>
- work out how we should charge for posting a letter there<sup>3</sup>

Some examples:
```python
from notifications_utils.countries import Country
```

```python
country = Country('pitcairn')
country.canonical_name
>>> 'Pitcairn, Henderson, Ducie and Oeno Islands'
country.postage_zone
>>> 'rest of world'
```

```python
country = Country('Macedonia')
country.canonical_name
>>> 'North Macedonia'
country.postage_zone
>>> 'Europe'
```

```python
country = Country('scotland')
country.canonical_name
>>> 'United Kingdom'
country.postage_zone
>>> 'United Kingdom'
```

If the country can’t be recognised a `CountryNotFound` exception will be raised:
```python
country = Country('London')
CountryNotFoundError: 'Not a known country or territory (London)'
```
***

You can try a working example that uses this code at https://postage.cloudapps.digital

***

The sources are:
- The [graph data](https://github.com/alphagov/govuk-country-and-territory-autocomplete/blob/master/dist/location-autocomplete-graph.json) from the [GOV.UK Country and Territory autocomplete](https://alphagov.github.io/govuk-country-and-territory-autocomplete/examples/) which includes the canonical names for countries and territories, as well as a bunch of useful synonyms
- [Royal Mail’s list of countries in Europe](https://www.royalmail.com/international-zones#europe), edited to use the normalised names from the country register

We are also storing some data of our own:
- `uk-islands.txt` – the following places whose addresses should not be normalised to _United Kindgom_, but should still be charged postage for the United Kingdom and allowed to choose between first and second class post:
    - Guernsey
    - Isle of Man
    - Jersey
- `synonyms.json` – a hand crafted list of synonyms not covered by the Country and Territory autocomplete graph
- `european-islands.txt` – the following places which are not countries, but are in the Royal Mail list of places you can send a letter to, and, similar to UK islands` should not be rewritten to their parent country:
    - Azores
    - Canary Islands
    - Madeira
    - Balearic Islands
    - Corsica

***

1. Real according to the [country register](https://www.registers.service.gov.uk/registers/country) or [territory register](https://www.registers.service.gov.uk/registers/territory). That is to say it’s more than just countries
2. Again, correct according to the registers
3. Postage can be _UK_, _Europe_ or _rest of world_
    - _UK_ is anything where the given country names maps to _United Kingdom_ (or _Jersey_, _Guernsey_ and _Isle of Man_)
    - Europe is places in the Royal Mail Europe postage zone
    - _rest of world_ is any other recognised country or territory


notifications_utils/countries/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+            if character not in " _-',.()"
+        )
+
+        if '?' in SanitiseASCII.encode(normalised):
+            return normalised
+
+        return SanitiseASCII.encode(normalised)
+
+    def __getitem__(self, key):
+        for option in (key, 'The {}'.format(key)):
```
idavidmcdonald: Do you have a test for the `The {}` case somewhere? Couldn't see one at a glance

notifications_utils/countries/data.py
```diff
@@ -0,0 +1,63 @@
...
+ADDITIONAL_SYNONYMS = [
+    (key, value)
+    for key, value in _load_data('synonyms.json').items()
+]
+
+_UK_ISLANDS_LIST = _load_data('uk-islands.txt')
+_EUROPEAN_ISLANDS_LIST = _load_data('european-islands.txt')
+
+UK_ISLANDS = [
+    (synonym, synonym) for synonym in _UK_ISLANDS_LIST
```
idavidmcdonald: It may not be a bad thing, but it's interesting to see that you are treating 'Isle of Man' (or similar islands) as not a synonym of the UK. That was my initial mental model. Instead you are treating `Isle of man` as it's own country and then the `UK_POSTAGE_REGIONS` decides if it's in the UK postage region. This doesn't match the pattern that you are doing for places like `Wales` which are treated as a synonym of `UK`. Any particular thoughts behind this?

notifications_utils/countries/__init__.py
quis: Added in cf6912a

notifications_utils/countries/data.py
quis: For parcels (but not letters) Royal Mail treat Jersey etc differently from the rest of the UK. Therefore they recognise it as a valid destination. The worry is that because these places are quite small, letters might only be addressed to the island, and not a town, for example this is a very deliverable letter:

> House Name
> Castle Road
> Guernsey

However if we treat _Guernsey_ as a synonym for _United Kingdom_, and then go on to strip it we get an ambiguous address:

> House Name
> Castle Road

The difference with Wales is that we want to treat it as a synonym for _United Kingdom_ then remove it, to ensure the last line is a postcode, so this:

> 1234 Castle Road
> Cardiff
> CF10 XYZ
> Wales

Becomes:
> 1234 Castle Road
> Cardiff
> CF10 XYZ

…and the we can check that _CF10 XYZ_ is a valid postcode.

But now I’m wondering after writing that why we aren’t just doing the same for Jersey etc… 🤔

tests/country_synonyms.py
```diff
@@ -0,0 +1,1895 @@
...
+    ('united states america', 'United States'),
+    ('united states of america', 'United States'),
+    ('united stats', 'United States'),
+    ('united sttes', 'United States'),
+    ('unitedarabemirates', 'United Arab Emirates'),
+    ('unitedarabrepublic', 'Egypt'),
+    ('unitedmexicanstates', 'Mexico'),
+    ('unitedrepublicoftanzania', 'Tanzania'),
+    ('unitedsat', 'United States'),
+    ('unitedstaes', 'U
```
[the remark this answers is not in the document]
quis: There was some weird character encoding stuff going on with them that I couldn’t figure out. Do you want to try uncommenting them and see if you can figure it out? I think it’s pretty unlikely we’ll actually get a letter addressed in this way, but it would be nice for completeness…

notifications_utils/countries/data.py
quis: Another issue here is that Sark, another one of the channel islands, costs the same as sending to Guernsey<sup>1</sup> but isn’t on the list of places you can send mail to. So if we got a letter addressed to Sark should we:
- let it through
- rewrite it to Guernsey, because that’s where the post office is
- remove it, and require a postcode

1. https://personal.help.royalmail.com/app/answers/detail/a_id/98/~/help-sending-mail-to-the-channel-islands-and-isle-of-man

quis: Consulted with @yahoopete and decided to treat the Channel Islands and Isle of Man the same as the rest of the UK, ie requiring a postcode. Done as part of df50dc1.

notifications_utils/countries/data.py
```diff
@@ -0,0 +1,72 @@
...
+
+# Copied from https://www.royalmail.com/international-zones#europe
+# Modified to use the canonical names for countries where incorrect
+ROYAL_MAIL_EUROPEAN = _load_data('europe.txt')
+
+
+class Postage:
+    UK = UK
+    EUROPE = 'Europe'
+    REST_OF_WORLD = 'rest of world'
```
leohemsted, line 72: can these be `united-kingdom`, `europe`, and `rest-of-world` so that it's more straightforward machine parseable? then we can do the mapping to human values on the admin app

quis: Changed in 95cb8e6

Will also need updating here: https://github.com/alphagov/notifications-api/pull/2747/files#diff-413e9aa8961126d20e2d90befa13e0f1R1355-R1356

PR #709, opened 2020-03-30
Speed up parsing of placeholders

quis:
String parsing can be a bit slow. And when we’re getting placeholders from a template that’s often unchanging, we shouldn’t be repeating this operation over and over again.

So this commit factors the placeholder-parsing into a separate function, which can then have an LRU cache[1] put on it.

# Improvements

Measured by running
```bash
python -m timeit -s "from notifications_utils.template import get_placeholders" "get_placeholders('hello ((name)) your ((thing)) is due soon')"
```

## Before
```
50000 loops, best of 5: 6800 nsec per loop
```

## After
```
5000000 loops, best of 5: 77 nsec per loop
```

i.e. almost 100× faster (for what is, admittedly, a tiny initial starting value)

***

1. https://docs.python.org/3/library/functools.html#functools.lru_cache


notifications_utils/version.py
```diff
@@ -1 +1 @@
-__version__ = '36.9.2'
+__version__ = '36.10.2'
```
idavidmcdonald: This should go to `36.10.0` I think as the patch version gets reset

quis:
Here’s some more testing where the template content can be one of 1,000 random values:
```bash
python -m timeit -s "import random;from notifications_utils.template import get_placeholders" "s = 'hello ((name)) your ((thing)) is due soon {}'.format(random.randint(0,999)); get_placeholders(s)"
```

## Before

```
100000 loops, best of 5: 2.18 usec per loop
```

## After

```
20000 loops, best of 5: 10.7 usec per loop
```

***

So that’s a 5× speedup in what’s probably a worse-than-real-case scenario.

PR #715, opened 2020-04-06
Make base template instances abstract

quis:
It’s dangerous to use the base template instances in any real code because they’re never going to be the thing that actually gets used to render the message.

So they could behave in unexpected ways, like not extracting all the placeholders for a template.

Making them abstract classes means they can never be used directly, and force us to choose the appropriate type of template in each case.


quis:
We’re not going to be able to do this because we still need to be able to use `WithSubjectTemplate` to render templates for the contents of API responses.

PR #720, opened 2020-04-14
Refactor template classes 

quis:
This commit does a refactor of our template classes with two aims:
- enable refactoring of the (currently slightly gross) API code that uses these classes
- make future work on CSV parsing to handle international letters easier by ensuring that any instance of one of these classes is guaranteed to have a correct, non-empty `template_type`

It does a lot of moving things around so that someone reading the code should be clearer about what each subclass is and isn’t responsible for.


tests/test_template_types.py
```diff
@@ -817,30 +879,48 @@ def test_subject_line_gets_applied_to_correct_template_types():
...
     ]:
-        assert not issubclass(cls, WithSubjectTemplate)
+        assert not issubclass(cls, SubjectMixin)
 
 
-def test_subject_line_gets_replaced():
-    template = WithSubjectTemplate({"content": '', 'subject': '((name))'})
+@pytest.mark.parametrize('template_class, template_type', (
+    (EmailPreviewTemplate, 'email'),
+    (LetterPreviewTemplate, 'letter'),
```
idavidmcdonald, line 892: Should we also test the other letter and email classes here? e.g. `LetterPrintTemplate`. Similar question for the other tests around this which are only testing the preview classes

tests/test_template_types.py
```diff
@@ -817,30 +879,48 @@ def test_subject_line_gets_applied_to_correct_template_types():
...
 @pytest.mark.parametrize("content, values, expected_count", [
     ("Content with ((placeholder))", {"placeholder": "something extra"}, 28),
     ("Content with ((placeholder))", {"placeholder": ""}, 12),
     ("Just content", {}, 12),
     ("((placeholder))  ", {"placeholder": "  "}, 0),
     ("  ", {}, 0),
 ])
-def test_WithSubjectTemplate_character_count(content, values, expected_count):
-    template = WithSubjectTemplate({"content": content, 'subject': 'Hi'})
+def test_WithSubjectTemplate_character_count(
```
idavidmcdonald, line 914: Should we rename this to get rid of `WithSubjectTemplate` which no longer exists

notifications_utils/template.py
```diff
@@ -158,6 +178,10 @@ def __str__(self):
             str.strip
         )
 
+    @property
+    def content_replaced(self):
+        return str(self)
```
idavidmcdonald, line 183: I think a comment above this to explain why we are overriding the base method would be good here please

notifications_utils/template.py
```diff
@@ -49,7 +50,7 @@
 ))
 
 
-class Template():
+class Template(ABC):
```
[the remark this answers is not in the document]
quis: I have:
- made `LetterImageTemplate` inherit from `BaseLetterTemplate` in 79b9fef
- made a `BaseSMSTemplate` in 6d1ebe6

I didn’t:
- make `LetterPrintTemplate` inherit from `BaseLetterTemplate` (it shares too much with `LetterPreviewTemplate`, and extracting the shared stuff out into another abstract class or mixin felt like it would be over-egging things)

notifications_utils/template.py
```diff
@@ -1,4 +1,5 @@
 import math
+from abc import ABC, abstractproperty
```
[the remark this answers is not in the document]
quis: Changed in 4168400

notifications_utils/template.py
```diff
@@ -83,6 +83,16 @@ def __repr__(self):
     def __str__(self):
         pass
 
+    @property
+    def content_replaced(self):
```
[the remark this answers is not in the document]
quis: Changed in c08d206

tests/test_template_types.py
quis: Added the extra test cases in a5eecb8

tests/test_template_types.py
quis: Renamed in c679a44

notifications_utils/template.py
quis: Comment updated in 4d3021f

notifications_utils/template.py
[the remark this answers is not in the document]
quis: So the updated list is:
```
class Template(ABC)

class BaseSMSTemplate(Template)

class SMSMessageTemplate(BaseSMSTemplate)

class SMSPreviewTemplate(BaseSMSTemplate)

class SubjectMixin()

class BaseEmailTemplate(SubjectMixin, Template)

class PlainTextEmailTemplate(BaseEmailTemplate)

class HTMLEmailTemplate(BaseEmailTemplate)

class EmailPreviewTemplate(BaseEmailTemplate)

class BaseLetterTemplate(SubjectMixin, Template)

class LetterPreviewTemplate(BaseLetterTemplate)

class LetterPrintTemplate(LetterPreviewTemplate)

class LetterImageTemplate(BaseLetterTemplate)
```

notifications_utils/template.py
```diff
@@ -164,7 +164,12 @@ def __init__(
     @property
     def content_with_placeholders_filled_in(self):
         # We always call SMSMessageTemplate.__str__ regardless of
-        # subclass, to avoid any HTML formatting
+        # subclass, to avoid any HTML formatting. SMS templates differ
+        # in that the content can include the service name as a prefix.
+        # So historically we’ve returned the fully-formatted message,
```
idavidmcdonald, line 169: To make sure I'm clear, when we say fully-formatted message is that 'service name: content' vs just 'content'? And you say 'plain text representation of the content'. Is the word 'plain-text' important, in that the fully-formatted message isn't plaintext?

notifications_utils/template.py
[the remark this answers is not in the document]
idavidmcdonald: Very optional (so optional that even I don't think I'd do it myself): Would be nice if all the email classes started with the word `Email`, like the letter and SMS classes do. Probably not worth the effort though, I just like things to look pretty ;)

notifications_utils/template.py
quis: It’s also about other normalisation, for example removing multiple newlines and spaces before punctuation.

For emails and letters the fully-formatted message is HTML, as opposed to plain text.

quis: It’s also about other normalisation, for example removing multiple newlines and spaces before punctuation.

For emails and letters the fully-formatted message is HTML, as opposed to plain text.

notifications_utils/template.py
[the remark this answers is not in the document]
quis: Yeah, `EmailHTMLTemplate` or `EmailHTMLEmailTemplate` don’t feel great either 😕

PR #721, opened 2020-04-17
Make sure letter personalisation is strings

CrystalPea:
Fix a small bug where our utils throws an unhandled exception if personalisation passed by the user is not a string.

TODO: write a test for this


notifications_utils/postal_address.py
```diff
@@ -40,7 +40,7 @@ def from_personalisation(cls, personalisation_dict):
         else:
             keys = address_lines_1_to_6_and_postcode_keys
         return cls('\n'.join(
-            personalisation_dict.get(key) or '' for key in keys
+            str(personalisation_dict.get(key)) or '' for key in keys
```
quis, line 43: This will result in `None` → `'None'`

If you do the `str` after the `or`:
```suggestion
            str(personalisation_dict.get(key) or '') for key in keys
```

`None` → `''`

CrystalPea: Good spot

PR #722, opened 2020-04-17
Account for bank holidays in delivery estimates

quis:
When a letter is delivered depends which days DVLA and the Royal Mail are working. DVLA don’t work weekends and bank holidays. Royal Mail are the same, except generally they do deliver on Saturday.

This commit:
- brings in the MoJ’s packaging<sup>1</sup> of the bank holidays from GOV.UK, as we’re now doing in the admin app
- makes the letter delivery calculations account for bank holidays

I did try to use the `get_next_work_day` method from the MoJ package, but it didn’t seem to work with the example I tried (most recent Easter weekend). So I wrote my own little method instead.

1. https://github.com/ministryofjustice/govuk-bank-holidays


notifications_utils/letter_timings.py
```diff
@@ -26,31 +36,26 @@ def get_letter_timings(upload_time, postage='second'):
     # shift anything after 5:30pm to the next day
     processing_day = utc_string_to_aware_gmt_datetime(upload_time) + timedelta(hours=6, minutes=30)
 
-    def next_monday(date):
-        """
-        If called with a monday, will still return the next monday
-        """
-        return date + timedelta(days=7 - date.weekday())
+    def get_next_work_day(date, bank_holidays):
```
idavidmcdonald, line 39: I think the argument `non_working_days` would be more accurate than `bank_holidays` here? As we are actually including bank holidays and 'weekends' in the days to be ignored, which isn't very clear just by the phrase `bank_holidays`

notifications_utils/letter_timings.py
```diff
@@ -12,6 +13,15 @@
 ]
 
 
+bank_holidays_monday_to_friday = BankHolidays(
```
idavidmcdonald, line 16: I wonder if it's better to call this `working_days_monday_to_friday` or `working_days_printers` or `working_days_dvla`

notifications_utils/letter_timings.py
```diff
@@ -12,6 +13,15 @@
...
+    use_cached_holidays=True,
+)
+bank_holidays_monday_to_saturday = BankHolidays(
```
idavidmcdonald, line 19: Maybe 'working_days_royal_mail` or similar

quis:
@idavidmcdonald good idea. I went with the naming from the package but I think this is clearer:
```python
get_next_work_day(date, non_working_days=non_working_days_dvla)
```

I’ve squashed that change into the previous commit.

PR #724, opened 2020-04-20
Let users provide any 3 address columns when sending letters by uploading a spreadsheet

quis:
Currently we require `address column 1`, `address column 2` and `postcode`, and we require that they’re all filled in. 

Similar to what we’ve done with the API and with precompiled letters, we should let you provide data for any of the 7 possible address columns, and just validate that we can make a good address from them.

This PR includes some refactoring and moving things around, so is best reviewed commit-by-commit.


tests/test_recipient_validation.py
```diff
@@ -374,34 +359,16 @@ def test_validate_address_allows_any_non_empty_value_for_non_postcode(column):
...
-
-
-def test_validate_address_raises_if_postcode_is_not_real_uk_postcode(mocker):
-    is_a_real_uk_postcode_mock = mocker.patch('notifications_utils.recipients.is_a_real_uk_postcode')
-    is_a_real_uk_postcode_mock.return_value = False
-    with pytest.raises(InvalidAddressError) as e:
-        validate_recipient('90899', 'letter', column="postcode")
-    assert is_a_real_uk_postcode_mock.called
-    assert 'Not a real UK postcode' == str(e.value)
+    assert validate_recipient(valid_address, 'letter') == valid_address
```
idavidmcdonald: We aren't using `column` here either?

tests/test_recipient_validation.py
```diff
@@ -374,34 +359,16 @@ def test_validate_address_allows_any_non_empty_value_for_non_postcode(column):
...
 
 
 def test_valid_address_line_does_not_raise_error():
```
idavidmcdonald: This test name is confusing. It says `valid_address` but then the variable we are using is called `invalid_address`? Also is this different to `test_non_ascii_address_line_is_fine_for_all_columns` in what it's testing or could they be merged?

notifications_utils/recipients.py
```diff
@@ -286,20 +286,15 @@ def is_optional_address_column(self, key):
...
     @property
     def has_recipient_columns(self):
-        # `self.column_headers_as_column_keys` is an `OrderedSet`. We need
-        # to cast it to a set because `OrderedSet` has misleading
-        # behaviour when doing comparisions with `Set` objects.
-        # See: https://github.com/simonpercivall/orderedset/issues/25
-        return self.required_recipient_column_headers <= set(self.column_headers_as_column_keys)
+        return len(
+            self.recipient_column_headers_as_column_keys
+            & self.column_headers_as_column_keys
```
idavidmcdonald, line 296: Nice. Optional - a small inline comment just to remind people that `&` in this case is set intersection to save them having to figure that out in the docs

tests/test_recipient_validation.py
```diff
@@ -323,51 +323,6 @@ def test_validate_email_address_raises_for_invalid(email_address):
...
-@pytest.mark.parametrize('column', [
-    'address_line_1',
-    'address_line_2',
-    'address_line_3',
-    'address_line_4',
-    'address_line_5',
-    'address_line_6',
-    'postcode',
-])
-def test_non_ascii_address_line_is_fine_for_all_columns(column):
```
idavidmcdonald: Is there some sort of check that we still want to ensure non ascii content in any of the address lines are fine?

notifications_utils/recipients.py
```diff
@@ -501,7 +501,6 @@ def validate_recipient(recipient, template_type, international_sms=False):
     return {
         'email': validate_email_address,
         'sms': partial(validate_phone_number, international=international_sms),
-        'letter': lambda _: recipient,
```
idavidmcdonald: Check we aren't still calling this case elsewhere

tests/test_recipient_validation.py
```diff
@@ -327,83 +323,6 @@ def test_validate_email_address_raises_for_invalid(email_address):
...
-
-@pytest.mark.parametrize('column', [
-    'address_line_1',
-    'address_line_2',
-    'address_line_3',
-    'address_line_4',
-    'address_line_5',
-    'address_line_6',
-])
-def test_non_ascii_address_line_is_fine_for_non_postcode(column):
```
idavidmcdonald, line 378: Do we wish to retain any checks that non ascii characters are fine in general in all our placeholders or is this just fine to assume?

notifications_utils/columns.py
```diff
@@ -130,6 +137,11 @@ def recipient(self):
         ]
         return columns[0] if len(columns) == 1 else columns
 
+    @property
+    def as_postal_address(self):
+        from notifications_utils.postal_address import PostalAddress
```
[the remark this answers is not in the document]
quis: `postal_address.py` imports from `countries/__init__.py` which imports from `columns.py`. Therefore we can’t import from `postal_address.py` at the top of `columns.py` without causing a circular dependency. Importing at a local level avoids the circular dependency.

tests/test_recipient_validation.py
quis: We’re not doing anything unusual with encoding or downgrading so I think we can assume that it should behave like any other string.

tests/test_recipient_validation.py
quis: I think this test has been removed in a subsequent commit.

tests/test_recipient_validation.py
quis: I think this test has been removed in a subsequent commit.

notifications_utils/recipients.py
quis: Added comment in aa2ecd2

notifications_utils/recipients.py
quis: All uses in this repo:
image

API:
image

Admin:
image

PR #725, opened 2020-04-22
Add a template type for previewing text message content

quis:
`SMSPreviewTemplate` previews the whole text message, including the sender and prefix, wrapped in some HTML that lets us style it like a speech bubble. Optionally it can redact missing personalisation, if it’s being used to render an already-sent message.

`SMSMessageTemplate` renders the actual message, including the prefix and has no option to redact missing personalisation.

This commit adds a third type of template for use on the notifications page, which just renders the content, like `SMSMessageTemplate` but:
- ignores any prefix (because this just takes up spaces that’s more usefully used by the content of the message)
- allows for redacting missing personalisation

Will fix this bug, where we’re incorrectly using `SMSPreviewTemplate`:

Screenshot 2020-04-22 at 11 58 58


CrystalPea:
It would be cool to see a "before" and "after" :)

quis:
# After 

Screenshot 2020-04-22 at 15 58 18

PR #739, opened 2020-05-21
Add estimated delivery dates for international letters

quis:
This lets us calculate the estimated delivery date for letters sent to Europe (which take 3–5 days) and the rest of the world (which take 5–7 days).

https://www.pivotaltracker.com/story/show/171622849

There’s quite a lot of refactoring in this PR so it’s best reviewed commit-by-commit.


notifications_utils/letter_timings.py
```diff
@@ -27,7 +28,55 @@ def set_gmt_hour(day, hour):
...
+    next_day = get_next_royal_mail_working_day(date)
+    if days_to_deliver == 1:
+        return next_day
+    return get_delivery_day(next_day, days_to_deliver=(days_to_deliver - 1))
+
+
+def get_min_and_max_days_in_transit(postage):
+    return {
+        # first class post is printed earlier in the day, so will
+        # actually transit on the printing day, and be posted the next
```
idavidmcdonald, line 65: Optional: Small nitpick, is it actually that first class is printed earlier in the day so will actually transit on the printing day and therefore be *delivered* the next day? The word posted to me implies the transit period

quis: Yeah, I think the original comment I copied maybe meant ‘posted through the letterbox’ but delivered is clearer
```suggestion
        # actually transit on the printing day, and be delivered the next
```

notifications_utils/letter_timings.py
```diff
@@ -27,7 +27,7 @@ def set_gmt_hour(day, hour):
     return day.astimezone(pytz.timezone('Europe/London')).replace(hour=hour, minute=0).astimezone(pytz.utc)
 
 
-def get_letter_timings(upload_time, postage='second'):
+def get_letter_timings(upload_time, *, postage):
```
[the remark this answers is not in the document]
quis: Yeah, think you’re right. You’ll get an exception if you use anything other than `first`, `second` etc for postage anyway, so should be clear enough what the problem is.
```suggestion
def get_letter_timings(upload_time, postage):
```

PR #749, opened 2020-06-12
Treat numbers in ranges belonging to crown dependency territories

CrystalPea:
of Jersey, Guernsey and Isle of Man as international sms so that they are sent by our international provider.

They are still charged the same as domestic sms.

This is done because our domestic provider was failing sms in this number range.


quis:
Will users who have international text messaging switched off in their settings still be able to send to these numbers?

CrystalPea:
@quis if the utils change is the only change to our code, then no, they wouldn't be able to. But if we add this change to the mix, it should be alright: https://github.com/alphagov/notifications-api/pull/2878

I will go and test it on my local now.

@idavidmcdonald 
> do we also need to change is_uk_phone_number to declare crown dep numbers as non UK?

then we would not validate it as a UK number, even though it follows UK pattern. Instead we would validate it according to international number validation, which is more permissive. I feel it would be a loss.

> And as Chris says, also mean that users can not send to international numbers without the permission to do so?

As per my above reply to Chris, I have a possible solution, will be testing it now.

> Further, is there any impact of having the TV phone numbers in this range. 07700900123 is in the list of TV phone numbers and is also classed as a number in the crown deps range and therefore is international?

Do you have any particular impact in mind?

CrystalPea:
@quis @idavidmcdonald ok I can now confirm [the API PR ](https://github.com/alphagov/notifications-api/pull/2878) is now tested, it works well with utils changes. 

With API changes I can confirm that:
- users without international permissions will still be able to send to crown dependency numbers
- there should be no changes for tv numbers in the range
- crown dependency numbers will have `international` flag in the db and so they will be sent with our international provider.

notifications_utils/recipients.py
```diff
@@ -405,12 +405,19 @@ def get_international_phone_info(number):
...
 
     return international_phone_info(
-        international=(prefix != uk_prefix),
+        international=(prefix != uk_prefix or _is_a_crown_dependency_number(number)),
         country_prefix=prefix,
         billable_units=get_billable_units_for_prefix(prefix)
     )
 
 
+CROWN_DEPENDENCY_RANGES = ['7781', '7839', '7911', '7509', '7797', '7937', '7700', '7829', '7624', '7524', '7924']
```
quis, line 414: I think we’re accidentally catching a small number of phone numbers which might be mainland UK here. But worst thing that happens is that they only get sent by one provider. So we should look later to see if we can make these ranges more specific. But if you’re happy that this range will catch all crown dependency numbers then I’m happy for us to ship as is.

CrystalPea: alright, lets do it! ☺️

PR #750, opened 2020-06-15
Make validating addresses in spreadsheets about 5× faster 

quis:
# Summary 

We’ve noticed some users getting `504` errors when uploading big CSV files. I think that this problem is specific to letters. Processing a file with lots of rows of address data has got a lot slower since we started validating each row as an address<sup>1</sup>.

The request timeout on our app servers is 60 seconds, so if it takes longer than that to process the CSV the user will see an error.

This pull request makes a series of optimisations to make the validation of addresses about 5× faster. 

Our aim should be to return a response in under 10 seconds, because this about the length of time a user will wait before going off and doing something else<sup>2</sup>. These optimisations get us in the region of 10 seconds, depending on machine speed and how much else is going on in the request.

# Before and after

Validating a CSV file with 50,000 unique but correct UK addresses. Run on a mid 2015 MacBook Pro with a 2.2 GHz Intel Core i7.

Profiler visualisations generated with [Snakeviz](https://jiffyclub.github.io/snakeviz/).

## Before 

66 seconds to run and profile the test. Almost all of the time is spent in the `PostalAddress` code:

Screenshot 2020-06-13 at 17 16 54

## After 

12 seconds to run and profile the test. About half of the time of the time is spent in the `PostalAddress` code:

Screenshot 2020-06-13 at 17 15 05

# How each commit contributes to the improvement

Most of the time saved comes from not re-running the same code that normalises the address. But there are some other smaller optimisations that give us another 30% improvement.

These timings are faster than those given above because they’re run without profiling.

Commit | Time to run test | Improvement
-------|------------------|------------
Add test for processing a big CSV of addresses | 38.11s | 0.00×
Compute frequently used properties in constructor of `PostalAddress` | 11.38s | 3.34×
Put LRU cache on postcode detection and formatting | 9.81s | 3.88×
Only compute Country instance for UK once | 9.64s | 3.95×
Optimise country lookup when given a postcode | 8.27s | 4.61×
Reduce number of lines in big address CSV test | 0.91s | N/A

***

1. We used to check only that there was content in 3 of the address columns – see https://github.com/alphagov/notifications-utils/pull/724
2. https://www.nngroup.com/articles/response-times-3-important-limits/


notifications_utils/postal_address.py
```diff
@@ -163,3 +163,9 @@ def format_postcode_for_printing(postcode):
     elif "BFPO" in postcode:
         return postcode[:4] + " " + postcode[4:]
     return postcode[:-3] + " " + postcode[-3:]
+
+
+@lru_cache(maxsize=8)
```
idavidmcdonald, line 169: Optional: Add to the commit message or code why maxsize 8 was chosen. You say `When validating an address we look at the postcode multiple times.` so I'm guessing we look at it up to something like 8 times so that's why it was picked.

Out of interest, when processing a CSV why do we look at the postcode multiple times?

quis:
```suggestion
# When processing an address we look at the postcode twice when
# normalising it, and once when validating it. So 8 is chosen because
# it’s 3, doubled to give some headroom then rounded up to the nearest
# power of 2
@lru_cache(maxsize=8)
```

PR #751, opened 2020-06-15
Only count characters of text message once

quis:
When we validate a spreadsheet for sending text messages we check to see whether:
- the resulting message would be empty
- the resulting message would be longer than the maximum size of a text message

This means that we count the number of characters in the message twice. Which means that we render the message twice. Which is quite costly. So this commit changes the code to only count the content once, then cache that value.

This takes the time to process 50,000 rows with a template that has a single placeholder down from about 13s to about 9s.

This is a bit fiddly because we often instantiate the template once, then change its values for each row of the spreadsheet. So we can’t just cache the value forever once it’s been computed once.


quis:
Update: counting the characters once instead of twice is some improvement. But what I didn’t notice earlier is that rendering the message in order to count the characters is a lot slower for longer templates because we are encoding the content before counting it.

Encoding the text message content just means swapping one character for another. So `len(sms_encode(content)) == len(content)`. Therefore running `sms_encode` is pointless if all you want to know is how many characters you have (rather than which characters).

And it turns out that encoding the content is expensive, especially for long templates.

So if we don’t encode the content we can save a lot of time. With 50,000 rows and a template of 600 characters this reduces the time for checking it for errors from 28s to 11s by applying 42f44dd.

quis:
So with both optimisations, for a row of 50,000 phone numbers, with a message template 600 characters in size:

Before | After | Improvement
---|---|---
51 seconds | 11 seconds | 4.6×

notifications_utils/template.py
```diff
@@ -158,8 +158,28 @@ def __init__(
...
+    @values.setter
+    def values(self, value):
+
+        # If we change the values of the template it’s possible the
+        # content count will have changed, so we need to reset the
+        # cached count.
+        if self._content_count is not None:
+            self._content_count = None
+
+        # Assigning to super().values doesn’t work here. We need to get
```
idavidmcdonald, line 177: Optional: Nice code comment but I'd also like to know why assigning to super().values doesn't work, rather than just the fact that it doesn't please?

quis: The specific error is
```
AttributeError: 'super' object has no attribute 'value'
```

There’s a bug report here: https://bugs.python.org/issue14965

Or as [someone on Stackoverflow put it](https://stackoverflow.com/a/10810545/147318):

> I was trying to find the correct language to back up why this behavior is the way it is, so as not to give you a "because it just is" answer...

PR #754, opened 2020-06-24
statsd error cleanup - reduce logging and cache dns failures

leohemsted:
### reduce error log to warning log for statsd errors

we send lots of statsd messages. If there's a problem with statsd, it's likely to be a high volume at once, which means we're likely to trip the pagerduty P1 error threshold

we see lots of statsd errors when cells roll. This is probably because there's just another moving cog - the internal DNS server, hosted on each cell. cells roll often and without our knowledge

there's nothing we can do about these errors, and they resolve themselves within seconds

there's a risk we may not see these messages at all if they're reduced to warning level, however, we can look at a tool like sentry if this is something we're concerned about

### dont send statsd data for 15 secs if DNS fails

If the DNS lookup fails (and it's not cached) we'll raise an exception, skip sending the statsd data, and then carry on with regular program flow - which may well try another statsd call almost immediately. Interestingly, if these DNS calls slow down, as we run in eventlets, we may end up with many concurrent DNS requests, compounding any problems the DNS server is experiencing.

To solve this, if the DNS lookup fails, cache that failed value for ~15 seconds. Then, if we encounter that failed value (None) when trying to send, we can just quietly and quickly skip sending statsd. After ~15 seconds the cache will expire and we'll try hitting DNS again.

### DIFF
--- a/notifications_utils/clients/statsd/statsd_client.py
+++ b/notifications_utils/clients/statsd/statsd_client.py
@@ -31,7 +31,7 @@ class NotifyStatsClient(StatsClientBase):
         try:
             self._sock.sendto(data.encode('ascii'), (self._cached_host(), self._port))
         except Exception as e:
-            current_app.logger.exception('Error sending statsd metric: {}'.format(str(e)))
+            current_app.logger.warning('Error sending statsd metric: {}'.format(str(e)))
             pass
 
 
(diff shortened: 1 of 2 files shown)


leohemsted:
Based on my incredibly scientific method of "unit tests and print statements", /v2/notifications/sms makes five statsd calls. /v2/notifications/email for a high volume service makes three statsd calls.

If the cache is empty and DNS fails, this PR will reduce the amount of (blocking) DNS lookups from five/three to one. (and really, that one will carry on for multiple requests afterwards)

PR #758, opened 2020-07-01
Add model for broadcast areas

quis:
This pull request adds a new model class which can be used by any app to interact with a broadcast area. A broadcast area is one or more polygons representing geographical areas.

It also adds some models that make browsing collections of these areas more straightforward. So the hierarchy looks like:

> **BroadcastAreaLibraries**
> Contains multiple libraries of broadcast areas
> > **BroadcastAreaLibrary**
> > A collection of geographic areas, all of the same type, for example counties or electoral wards
> > > **BroadcastArea**
> > > Contains one or more shapes that make up an area, for example _England_ or _Hackney South and Shoreditch_
> > > > **BroadcastArea.polygons[n]**
> > > > A single shape, for example the _Isle of Wight_ or _Lindisfarne_
> > > > > **BroadcastArea.polygons[n][o]**
> > > > > A single coordinate pair marking a point along a polygon

The classes support iteration and sorting, so all the areas in a library can be looped over, for example if `countries` is an instance of `BroadcastAreaLibrary` you can do:
```python
for country in countries:
    print(country.name, country.id)

>>> 'England', 'england'
>>> 'Northern Ireland', 'northern-ireland'
>>> 'Scotland', 'scotland'
>>> 'Wales', 'wales'
```

The `BroadcastAreaLibraries` class also provides some useful methods for quickly getting the polygons for an area or areas, for example to render them on a map. So if `libraries` is an instance of `BroadcastAreaLibraries` you can do:
```python
for polygon in libraries.get_polygons_for_areas_long_lat('england', 'wales'):
    print(polygon)

>>> [[-0.000, 51.123], [-0.321, 51.123], …]
>>> […]
>>> …
```

One polygons might represent the Welsh mainland, Anglesey, the English mainland or the Isle of Wight, and so on.

The models load data from GeoJSON files, which is an open standard for serialising geographic data. I’ve added a few example files taken from http://geoportal.statistics.gov.uk to show how it works.


notifications_utils/broadcast_areas/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+            with suppress(KeyError):
+                return self.feature['properties'][possible_name_key]
+
+        raise KeyError(f'No name found in {self.feature["properties"]}')
+
+    @property
+    def polygons(self):
+        if self.feature['geometry']['type'] == 'MultiPolygon':
+            return [
+                polygons[0]
```
leohemsted, line 80: does each array within `coordinates ever contain more than 1 thing? i had a quick look in the spec but wasn't clear 🤷‍♂️

notifications_utils/broadcast_areas/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+                )
+
+    def __eq__(self, other):
+        return self.id == other.id
+
+    @property
+    def name(self):
+        for possible_name_key in {
+            'rgn18nm', 'ctyua16nm', 'ctry19nm',
+        }:
```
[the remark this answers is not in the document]
quis: Yeah, each set seems to have a different key for the name. The number refers to the year of publication, eg 2018, 2016, 2019

notifications_utils/broadcast_areas/__init__.py
quis: Do you mean an array that mixes coordinate pairs and other things?

leohemsted: i mean we're saying `polygons[0]` - is there ever a `polygons[1]`? (And the same when we're accessing point [0] in Polygons.

Turns out it could hypothetically: https://tools.ietf.org/html/rfc7946#appendix-A.3

> Coordinates of a Polygon are an array of linear ring (see
>   Section 3.1.6) coordinate arrays.  The first element in the array
>   represents the exterior ring.  Any subsequent elements represent
>   interior rings (or holes).

don't think we need to worry about these until we find a data set with one in

PR #760, opened 2020-07-02
Optimise checking for empty emails and letters

quis:
At the moment we render the entire content of an email or letter for every row in a CSV file, just to check whether it has more than zero characters.

We have to do this in case all of the message is contained in a placeholder, for example a template like `((content))`.

However rendering templates is quite expensive, because it involves a lot of regex and string mutation. For any case where the template has content before or after any placeholders we can skip this check – knowing that the template has at least some content is enough.

This assumes that templates won’t have leading or trailing whitespace, which the admin app guarantees at the point of creation.

# Benchmarks

Checking a 50,000 row CSV for errors, against a template containing 5 paragraphs of [lorem ipsum](https://www.lipsum.com/) with one placeholder.

## Before (10.4s)

Screenshot 2020-07-02 at 09 32 53

## After (5.6s)

Screenshot 2020-07-02 at 09 33 51


tests/test_template_types.py
```diff
@@ -1178,6 +1182,41 @@ def test_is_message_empty_email_and_letter_templates(
...
+    })
+    mock_content = mocker.patch.object(
+        template_class,
+        'content_count',
+        create=True,
+        new_callable=mock.PropertyMock,
+        return_value=None,
+    )
+    template.values = values
+    template.is_message_empty()
```
idavidmcdonald, line 1216: Optional: Could assert that this is false as a sanity check that we've set up the rest of the test correctly

notifications_utils/template.py
```diff
@@ -137,6 +137,16 @@ def content_count(self):
         return len(self.content_with_placeholders_filled_in)
 
     def is_message_empty(self):
+
+        if not self.content:
```
idavidmcdonald, line 141: Optional: I don't see any added tests for this line of code, is there a small test that you can write?

Also, is it even possible to have this case?

quis: It’s tested by this line:
https://github.com/alphagov/notifications-utils/pull/760/files#diff-e5050dea996611c8bdf32cfe6e8214cdR1161

I don’t think it’s something that can really happen, but I didn’t want to remove an existing test case.

PR #761, opened 2020-07-06
Change BroadcastMessageTemplate to output XML, not plain text

quis:
This commit adds the minimum required† properties to the output of `BroadcastMessageTemplate.__str__` so that it validates against [the CAP XML schema](https://docs.oasis-open.org/emergency/cap/v1.2/CAP-v1.2.xsd).

At the moment it hard codes a lot of values with what seem like sensible defaults. We’ll probably need to change these or make them configurable at some point in the future.

***

Using this in an app will look something like:
```python
xml = str(BroadcastMessageTemplate(
    template={
        'template_type': 'broadcast',
        'content': 'Hello ((name))',
    },
    values={
        'name': 'Example',
    },
    polygons=broadcast_area_libraries.get_polygons_for_areas_long_lat(
        'the-south-west',
        'the-north-east',
    ),
))
```

***

† polygons and description aren’t strictly required, but it’s not much of a broadcast without them


CrystalPea:
Hey,

It looks good to me, nice tests, too. 

I have a few questions though:

1. what would you like from a review here? What would you recommend that I check?
2. at what point would we substitute hardcorded values? Wouldn't it be easier to make them into parameters with default values straight away?

quis:
@CrystalPea 

1. I think the most important thing is to check that the tests are doing what you’d expect. So if you remove something from the XML or change it, does it cause the test to fail? I added some extra tests in 78d4802 to be more confident of this
2. At the moment we don’t know which values the phone uses. So I don’t want to make things into parameters until we know that they’re something we want to vary.

PR #763, opened 2020-07-09
broadcast templates to include sender and identifier

quis:


tlwr:
So that CAP XML validates

PR #765, opened 2020-07-09
Leos fixes

leohemsted:
### ensure you can get broadcast_areas from a list as well as from *args

i'm not a fan of *args generally. They're good when someone will be calling your function with static data (like a print function, think `print('hello', name, '.')` or when the function needs to be as flexible as possible so it can silently replace something (eg: a decorator that calls `func(*args, **kwargs)` in the middle.

However, get_area (and the lat_long and long_lat functions that call it) don't fulfil either of these checks: You'll be calling it with a dynamic list (from a db or json response). And the function is very opinionated about what you pass in - you should pass in a list of area id strings, nothing else will work.

To make this a non-breaking change, allow you to pass in either

### refer to sender as notify_identifier

it's a very thematically separate concept to `BaseSMSTemplate.sender`. Also, set it to our website, based on this quote from the CAP spec.

> Identifies the originator of this alert. Guaranteed by assigner to be unique globally; e.g., may be based on an Internet domain name.

### allow you to create broadcast message template with areas 

i accidentally passed the wrong way round in, creating valid but completely wrong coordinates pointing to the middle of the ocean. There's no point allowing someone to give this option when the template knows about areas, and knows the best way to extract those areas.

make it override polygons if not defined, but still accept polygons for backwards compatibility


notifications_utils/broadcast_areas/__init__.py
```diff
@@ -147,6 +147,10 @@ def get_all_areas(self):
                 yield area
 
     def get_areas(self, *area_ids):
+        # allow people to call `get_areas('a', 'b') or get_areas(['a', 'b'])`
```
tlwr, line 150: 🤨

tlwr:
This is green in Concourse

notifications_utils/broadcast_areas/__init__.py
leohemsted: i'd rather just the second, but in the sake of keeping it non-breaking

PR #769, opened 2020-07-23
Add electoral wards to broadcast areas

tlwr:
What
----

Before:
```
len(list(broadcast_areas.broadcast_area_libraries.get_all_areas()))
187
```
After:
```
len(list(broadcast_areas.broadcast_area_libraries.get_all_areas()))
6751
```

These are put in the `Electoral Wards of the United Kingdom` group.

Process
---

In order to get the wards in a GeoJSON file that had a useful name I added a script `join-wards.py` which joins two vendored datasets:
- Wards and polygons without LA names
- Wards and LA names

I did not spend any time working around the performance implications of the current implementation:
- The vendored dataset and the generated dataset add about 80MB of the size of the package
- `from notification_utils import broadcast_areas` takes a couple of seconds instead of near instantaneous import

Problems
---

1) We are missing some wards (~2000 / 9000), because the codes have changed. We can change this when we find two datasets that work better

2) I think the strategy of loading all the areas when importing the package isn't feasible going forward, but some discussion in this PR on how to proceed would be welcome

I think we should:

- Hardcode the list of datasets we expect (instead of the `*.geojson` glob iteration)
- Hardcode a description of the dataset (instead of the autogenerated `East Midlands, East of England, London and 6 more…`)
- Do we want these in (a/the) database instead? Do we want a sharded set of GeoJSON on disk? I'm concerned about memory usage when loading a large file into memory with eventlet.


simonneb:
Hi - you can use these 3 geojson files combined with the final table to match electoral wards -> parliamentary constituency -> local authority.

**geojson**

- [Wards](https://geoportal.statistics.gov.uk/datasets/wards-may-2020-boundaries-uk-bgc?geometry=-18.984%2C53.201%2C14.392%2C57.564)

- [Parliamentary constituency
](https://geoportal.statistics.gov.uk/datasets/westminster-parliamentary-constituencies-december-2019-boundaries-uk-bgc?geometry=-8.947%2C51.405%2C7.741%2C53.741)

- [Local Authorities](https://geoportal.statistics.gov.uk/datasets/local-authority-districts-may-2020-boundaries-uk-bgc)

**matching file**

- [Ward to Parliamentary constituency to LA matching file](https://data.gov.uk/dataset/700e207d-f747-4899-a83c-44575097c694/ward-to-westminster-parliamentary-constituency-to-local-authority-district-december-2019-lookup-in-the-united-kingdom)

tlwr:
Thanks @simmoneb that is super helpful

I've resolved some of the concerns I had above with file sizing, we're using an embedded sqlite db generated from the geojson. Using the datasets you provided reduced the number of skipped wards to 219.

```
sqlite> select dataset, count(*) from broadcast_areas GROUP BY dataset;
Counties and Unitary Authorities in England and Wales|174
Countries|4
Electoral Wards of the United Kingdom|8595
Regions of England|9
```

tlwr:
Total sizes now:

```
8662544842 -rw-r--r--   1 tobylornewelchrichards  staff   1.5M 23 Jul 14:23 Counties and Unitary Authorities in England and Wales.geojson
8662544843 -rw-r--r--   1 tobylornewelchrichards  staff   324K 23 Jul 14:23 Countries.geojson
8662544844 -rw-r--r--   1 tobylornewelchrichards  staff   147K 23 Jul 14:23 Regions of England.geojson
8650865121 drwxr-xr-x  32 tobylornewelchrichards  staff   1.0K 23 Jul 14:23 ..
8663012390 -rw-r--r--   1 tobylornewelchrichards  staff   4.3K 24 Jul 16:56 repo.py
8663012391 -rwxr-xr-x   1 tobylornewelchrichards  staff   2.0K 24 Jul 16:56 create-broadcast-areas-db.py
8663012439 -rw-r--r--   1 tobylornewelchrichards  staff   3.5K 24 Jul 16:56 __init__.py
8663012459 -rw-r--r--   1 tobylornewelchrichards  staff    45M 24 Jul 16:56 Electoral Wards May 2020.geojson
8663012461 -rw-r--r--   1 tobylornewelchrichards  staff   2.1M 24 Jul 16:56 Electoral Wards and Local Authorities 2020.geojson
8658785042 drwxr-xr-x  12 tobylornewelchrichards  staff   384B 24 Jul 16:56 .
8663012462 -rw-r--r--   1 tobylornewelchrichards  staff    41M 24 Jul 16:56 broadcast-areas.sqlite3
```

The sqlite3 database is never loaded into memory, just seeked

notifications_utils/broadcast_areas/__init__.py
```diff
@@ -98,25 +77,19 @@ def unenclosed_polygons(self):
...
-    def __init__(self, filename):
-        self.name, geojson_data = load_geojson_file(filename)
-        self.items = geojson_data['features']
-
-    def get_examples(self, max_displayed=4):
+    def __init__(self, library_name):
+        self.name = library_name
 
-        truncate_at = max_displayed - 1
+    def get_examples(self):
```
servingUpAces, line 87: Where/how do we use this get_examples method? 
I understand that how it shows an example of the dataset for a given library like "countries". But when the dataset is large like it is for local authorities how will it be useful?

tlwr: Currently this is used [when selecting a library](https://github.com/alphagov/notifications-admin/blob/78bda9d2679faad1f4611866816ef154da304e54/app/templates/views/broadcast/libraries.html#L180)

ie on this page 
image

with the changes currently in this PR it looks something like

```
Choose where to broadcast to

Counties and Unitary Authorities in England and Wales
Barking and Dagenham, Barnet, Barnsley, 171 more…

Countries
England, Northern Ireland, Scotland and Wales

Regions of England
East Midlands, East of England, London, 6 more…

Electoral Wards of the United Kingdom
Eden - Ravenstonedale, Folkestone and Hythe - Sandgate & West Folkestone, East Lothian - Musselburgh, Daventry - Weedon, 8591 more…
```

tlwr:
Have done some MVP feature simplification (if @quis wants to do some testing)

Now there is an attribute `simple_polygons` on `BroadcastArea` where all polygons:
- do not contain islands
- are capped to 125 points (arbitrarily)

There is also an extremely hacky tool for visualising the DB, the below have two layers, a green layer of the real polygon, and a blue layer of the simplified polygon, blue/green where they intersect

derbyshire and derby
_Derby is enclosed by Derbyshire when simplified_

simplified bits of London
_Simplified bits of London_

England looks okay
_England looks okay_

simonneb:
Hi @tlwr - after our call yesterday,I have been experimenting with various simplification methods to lower the point count of the polygons whilst maintaining as much fidelity as possible.

Looking at the work you have done [here](notifications_utils/broadcast_areas/create-broadcast-areas-db.py) I think that is more than adequate. The point count is likely to be the issue for the CBCs, so reducing the fidelity of the polygon with the max point count per CBC in mind is the best route to take.

The most complex polygon from the [Local Authorities](https://geoportal.statistics.gov.uk/datasets/local-authority-districts-may-2020-boundaries-uk-bgc) data set was **Highlands** in Scotland. It would be interesting to see what that looks like using your simplification script.

tlwr:
If you [email me](mailto:toby.lornewelch-richards@digital.cabinet-office.gov.uk) your geojson file, or pastebin it to me, I can add it to the libraries. That would be interesting.

---

The following images were generated to ~150 polygons, with a buffer proportional to the simplification

Scotland downsampled
_Scotland downsampled_

Close up of coast
_Close up of coast_

The effect of the buffering on the Scotland/England border
_The effect of the buffering on the Scotland/England border_

notifications_utils/broadcast_areas/repo.py
```diff
@@ -0,0 +1,198 @@
...
+        description_areas_joined AS (
+            SELECT GROUP_CONCAT(name, ", ") FROM description_area_names
+        )
+        SELECT
+        CASE (SELECT * FROM subset_area_count)
+        WHEN 0 THEN
+            (SELECT * FROM description_areas_joined)
+        ELSE
+            (SELECT * FROM description_areas_joined)
+            || ", "
```
quis, line 123: To match what the admin app does at the moment:
```suggestion
            || ", and "
```

tests/test_broadcast_area.py
```diff
@@ -1,83 +1,68 @@
...
     ] == [
-        ('england', 'England'),
-        ('northern-ireland', 'Northern Ireland'),
-        ('scotland', 'Scotland'),
-        ('wales', 'Wales'),
+        ('countries-E92000001', 'England'),
+        ('countries-N92000002', 'Northern Ireland'),
+        ('countries-S92000003', 'Scotland'),
+        ('countries-W92000004', 'Wales'),
     ]
```
quis, line 48: Is it worth adding a similar test for some wards? Just to check that all the wards for a given area have the names and IDs we’d expect?

notifications_utils/broadcast_areas/__init__.py
```diff
@@ -63,98 +49,94 @@ def __init__(self, feature):
...
+    def simple_feature(self):
+        return geojson.loads(self._simple_feature)
 
-        names = [area.name for area in sorted(self)]
-        count_of_excess_names = len(names) - truncate_at
+    @property
+    def sub_areas(self):
+        return [
+            BroadcastArea(row)
+            for row in BroadcastAreasRepository().get_all_areas_for_group(self.id)
```
quis, line 102: Is there potentially some inefficiency in instantiating and connecting to the database each time we call `.sub_areas` (and others)? Would it not be better to do this once at the time of starting the app?

tlwr:
➡️ https://github.com/alphagov/notifications-admin/pull/3553

tlwr:
This needs a squash but now this PR removes broadcast areas from the app

notifications_utils/template.py
```diff
@@ -372,17 +371,15 @@ def __init__(
         self,
         template,
         values=None,
-        polygons=None,
-        areas=None,
+        areas=[],
+        polygons=[],
```
quis, line 375: You should [avoid using mutable types as default arguments](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) unless you really dislike your future self. Instead do:
```python
def __init__(areas=None):
    self.areas = areas or []
```

tlwr: a66559ec6728c49ee78fe3e0fc849afa517c207d

PR #776, opened 2020-08-05
Copy SQLite files when installing package

quis:
We’ll need this for the electoral wards stuff. By default Pip only installs Python files.


quis:
This isn’t needed any more because we’ve moved the broadcast areas stuff to the admin app.

PR #784, opened 2020-09-07
Do not break gov.uk if a '=' character precedes it

idavidmcdonald:
Currently, if we have 'gov.uk' that does not look like a URL then we add
a small non breaking space between gov. and uk so that it is not
formatted as a link.

This commit adds a new case where we should not break gov.uk when an
equals sign precedes gov.uk as this likely indicates it is part of a
url, as per the unit test.

https://www.pivotaltracker.com/story/show/174510633


leohemsted (commented):
This'll work for this case but I just have the horrible feeling that it'll break quietly the next time someone sends this stuff, but with a slightly different query string arg. We shouldn't be altering any query strings, regardless of what their content looks like. eg `?foo=bargov.uk` will fail this.

I'm not sure we can easily use a regex for this - pretty much the only thing that _can't_ come before `gov.uk` inside a query string is whitespace I think?

here's a couple of additional use cases that show things we don't want to pick up, but we currently do (and your proposed solution still will)
```
www.this-site-is-not-gov.uk
www.gov.uk?websites=bbc.co.uk;gov.uk;nsh.scot
```

and for reference here are some examples from real templates where we are currently (correctly) adding a nbwsp to gov.uk (`select id, content from templates where content ~* '.*(?<!\.|\/|\=)(GOV)\.(UK)(?!\/|\?).*';`)
```
the GOV.UK Licensing service.  # in a sentence
Regards\n\nGOV.UK # new line before, end of string after
both data.gov.uk and GOV.UK.  # with a fullstop after
now published on GOV.UK:  # with a colon after
To unsubscribe please reply to: xxxx@xxx.gov.uk # as part of email, with end of string after
```

idavidmcdonald:
Good investigation.

Yes agreed, this is a simplistic way to fix the bug and only fixes it for this use case of 'gov.uk' being used as the value of a query param.

Just to double check, you said 'I'm not sure we can easily use a regex for this' which I think I agree with. I'm also cautious about trying to come up with a solution that may be fairly complex and break things for existing users. 

I kind of want to merge this as is just as it will fix the immediate issue for the gov.uk email team but I would also be keen to think a bit about how we would put in a long term fix for this. If it wasn't with regex, how would we do this even?

leohemsted:
The other option is (at least temporarily) removing the gov.uk un-linking. The benefits of not having it as a link are:

* less confusion for users about if it's a link they need to click on to do things
* more consistent screenreader experience (the same as for sighted users but a bit more important I think)
* Consistent with branding elsewhere where we have more control over what's a link and what's not.

Those are valid reasons for implementing it but breaking URLs is definitely worse than stopping those in my eyes.

==================

I think, however, we could do something more around "identify gov.uk used as a word" rather than "identify gov.uk not used as a word". It's better to miss de-linking a niche `gov.uk` usage than accidentally de-link a niche `gov.uk` usage in a url I think. We currently say "lets try and find gov.uk where it's not part of a URL"

> Find all usages of "GOV.UK", unless they're preceeded by one of `.`,`/`,`=`

Instead we could flip the logic. "Lets try and find gov.uk where it's presentational

> Find all usages of "GOV.UK" where it has whitespace before the GOV.UK.

I _think_ we could do this with `r'(^|\s)(GOV)\.(UK)(?!\/|\?)'` (we'd probably need to tweak the `re.sub` command to make sure it keeps the leading whitespace`r'\0\1' + '\u200B' + r'\2').

Ugh, except I've just realised that we have people using `gov.uk/register-to-vote` in at least one case. And we've got people using gov.uk in a header without a space before, both as `#gov.uk notify` and `#gov.uk/register-to-vote`

I think this solves it: `(^|\s)(#|\*|-)?(GOV)\.(UK)(?!\/|\?)` as "proven" here: https://regexr.com/5bjal. But I'm worried I'm down a rabbit hole now. Dunno if @quis has any thoughts?

quis:
@leohemsted I think that’s a neat solution, and much more sensible than what we had before. 

The only edge case I could find was [`gov.uk#departments-and-policy`](https://www.gov.uk/#departments-and-policy) which is unlikely to be used in a real email but might be worth adding to the regex for completeness

idavidmcdonald:
@leohemsted @quis I've pushed up a new set of (failing) tests which I think could be test data for the behaviour we want if we go down the approach suggested (find uses we are sure are presentational).

Have a look at the test data and let me know if you think I've got the test cases with the behaviour we want. Note, the tests will need a bit of changing as things like `*` get rendered as `•` etc but I think it illustrates the behaviour we want.

If so, then I can actually drop in/tweak the regex Leo mentioned above

tests/test_formatters.py
```diff
@@ -206,6 +207,47 @@ def test_escaping_govuk_in_email_templates(template_content, expected):
...
+        ("coronavirus.data.gov.uk", "coronavirus.data.gov.uk"),
+        ("gov.uk/foo", "gov.uk/foo"),
+        ("*GOV.UK/foo", "*GOV.UK/foo"),
+        ("#GOV.UK/foo", "#GOV.UK/foo"),
+        ("^GOV.UK/foo", "^GOV.UK/foo"),
+
+        # Cases I'm not sure on
+        ("data.gov.uk", "data.gov.uk"), # depends if this is a special case like gov.uk is a brand
+        ("southwark.gov.uk", "southwark.gov.uk"), # i lean towards making this a link
+        ("#gov.uk/register-to-vote", "#gov.uk/register-to-vote") # i lean towards making this a link
```
leohemsted, line 240: we currently make all three of them links (based on the leading `.` for the first two, and the trailing `/` for the last one)

i think we keep that behaviour. Deffo last two I agree with you. data.gov.uk's a bit more interesting but I dont know enough about branding etc on that.

idavidmcdonald:
@leohemsted Great suggestion, hats off for the hard work, I just had to write it up 👍 

This approach seems to work really well. Your suggested regex `(^|\s)(#|\*|-)?(GOV)\.(UK)(?!\/|\?|#)` didn't need much tweaking. I included the `^` case for styling. The other question I have is that in yours you were finding instances of `-GOV.UK` is the `-` case intentional that we don't want to break on? The dash could be used as an informal list item but I haven't included it (`*` is the our markdown bullet point). Thoughts on whether we should or shout not break `-GOV.UK` or not?

idavidmcdonald:
Oh, I have found some edge cases that break our regex. The following cases fail the test, but I haven't added them in a commit

```
("gov.uk.com", "gov.uk.com"),
("gov.ukandi.com", "gov.ukandi.com"),
("gov.uks", "gov.uks"),
```

I can't find any easy way to change our regex to not pick these up. Is this acceptable? I'd say probably yes as the only realistic one would be if someone had `gov.uk` at the very start of a their URL which I feel is highly unlikely.

leohemsted:
@idavidmcdonald i think those three cases are fine too - i might be tempted to add them as `pytest.mark.xfail` cases maybe just record that we know and acknowledge them?

also you're right that `-gov.uk` doesnt really mean anything with our markdown.

servingUpAces (approved):

PR #786, opened 2020-09-07
Always preview postage of international letters as international

quis:
No matter what we pass in, international letters should never be shown with second or first class postage.


idavidmcdonald (approved):
Looks good.

Do you know if there is any performance impact for CSV uploading? I don't know if we access `self.postage` during processing of a CSV? Or if we do, if we are already instantiating the `PostalAddress` class for each of them anyway (in which case, very little extra processing is needed to check if the letter is international if the class has already been instantiated)

Does this mean there was somewhere we were showing an international letter as say 2nd class postage in the admin app? Got a screenshot for context for me?

Regardless I'm happy with the code though

quis:
@idavidmcdonald we don’t look at postage while processing a CSV. We do construct a `PostalAddress` elsewhere to check the count of address lines etc:
https://github.com/alphagov/notifications-utils/blob/01678e46226d0eb95d69977adff7be5e94518e77/notifications_utils/columns.py#L140-L146

Changes in this PR shouldn’t have any impact on CSV processing performance.

PR #789, opened 2020-09-15
Update letter placeholders to use new focus colour

quis:
So that we don’t have two different yellows in use.

Once template preview is deployed, this will match https://github.com/alphagov/notifications-admin/pull/3628


tombye (approved):
Looks good to me 👍

PR #790, opened 2020-09-25
Increase max CSV file size to 100,000 rows

quis:


leohemsted (approved):
oh no, even more time sat waiting for test_big_list to finish 😅

PR #803, opened 2020-11-02
Take extended GSM characters into account in fragment count

klssmith:
Extended GSM characters count as two GSM characters, but we were not
taking this into account. We now count extended GSM characters as two
characters when calculating the fragment count for SMS and broadcast
messages.

[Pivotal story](https://www.pivotaltracker.com/story/show/175457294)


notifications_utils/template.py
```diff
@@ -236,7 +236,14 @@ def content_count_without_prefix(self):
     @property
     def fragment_count(self):
         content_with_placeholders = str(self)
-        return get_sms_fragment_count(self.content_count, non_gsm_characters(content_with_placeholders))
+
+        no_of_extended_gsm_chars = sum(
+            map(content_with_placeholders.count, SanitiseSMS.EXTENDED_GSM_CHARACTERS)
+        )
+        # Extended GSM characters count as 2 characters
+        character_count = self.content_count + no_of_extended_gsm_chars
```
quis, line 244: Very tidy way of doing this 👍

quis (approved):

PR #805, opened 2020-11-09
Revert "Revert 798 validate email size"

servingUpAces:
Reverts alphagov/notifications-utils#804

This was reverted because if the payload for an email had a file for document download, the size of the file was included in the length validation. 

https://github.com/alphagov/notifications-api/pull/3027 will only check the length of the email after the document has been replaced with the link to the uploaded file.


CrystalPea (approved):

klssmith:
Since the version number is now going down you might need to delete the `43.2.0` tag from the repo, since that version won't exist any more

PR #810, opened 2020-12-02
Validate overly-long international phone numbers

quis:
We check that UK phone numbers don’t have too many digits:
https://github.com/alphagov/notifications-utils/blob/db8171f23cfc2f29d35a2412db545ad64f6aeb0a/notifications_utils/recipients.py#L442-L443

We don’t do a similar check for international phone numbers.

According to [the E.164 numbering plan](https://www.itu.int/rec/dologin_pub.asp?lang=f&id=T-REC-E.168-199905-S!!PDF-E&type=items) a phone number can be at most 15 digits, including the country code:
image

This commit adds validation to reject phone numbers longer than 15 digits, which might help our users catch an error somewhere in their system (like the missing comma it found in our test data).


idavidmcdonald (approved):
So looks good and matches existing code.

Question over the actual error message. I believe our error messages shouldn't just tell people what's wrong but also tell them how to fix it. For example should it be telling them how much shorter it needs to be? However, it is consistent with the other errors for this.

I think it's good to merge regardless but might be worth a review of the copy for these errors in the future cc @karlchillmaid

PR #812, opened 2020-12-07
Optimise key generation for `Columns` class

quis:
# The change

[`str.translate`](https://docs.python.org/3/library/stdtypes.html#str.translate) seems to be the fastest way of removing characters from a string in Python.

Part of the work `Columns.make_key` does is to remove spaces, dashes and underscores so that `first_name` and `first name` are considered equivalent. It is called _a lot_ when doing replacement of placeholders, so any optimisation we can do on it is a good win.

# Results

To test this I temporarily removed the LRU cache on the method, and then ran the following shell command:
```bash
python -m timeit -s "from notifications_utils.columns import Columns" "Columns.make_key('this is _not_ a -test')"
```

## Before
```
100000 loops, best of 5: 3.86 usec per loop
```

## After
```
200000 loops, best of 5: 1.18 usec per loop
```

3.27× faster

# Previous

Some previous optimisation in https://github.com/alphagov/notifications-utils/pull/48 where we discussed removing the `split`/`join` but ultimately didn’t.


idavidmcdonald (approved):
Nice and clear, thanks

PR #814, opened 2020-12-08
Fix bug with double semicolons appearing in sequences of special characters 

quis:
We use a library called Bleach for escaping dangerous HTML in our template content. It doesn’t seem to handle things which look like HTML entities (for example `&copy;` or `&gt;`) but contain unexpected characters (the only one I’ve found so far is a question mark, for example `&?a;`).

This results in the last character of an escape sequence being replaced with a semicolon. I haven’t dug into the Bleach source deeply enough to identify where the problem is, and which other characters might affect it.

But a fix is to use Python’s builtins to convert any escape sequences to their unicode equivalent, and then escape them again. This seems to not get confused the way Bleach does.

We can then pass the output of this to Bleach, which passes through stuff that’s already properly encoded.

This fix is described in https://github.com/mozilla/bleach/issues/453#issuecomment-499182492:

> A quick solution that can be implemented by developers is to just escape(unescape()) the input before sending it to bleach.
>
> …
>
> This solution does not utilize a HTML parser, it's all regex under the hood, so there may be some edge cases. This has passed all our use cases though.
>
> While I personally think bleach should handle stuff like this, I don't think it is necessary when invoking bleach is only a component of your user input sanitization routine.

There’s one more gotcha, which is that, for historic reasons, we allow users ([specifically GOV.UK Email](https://github.com/alphagov/notifications-utils/pull/427)) to include the non-breaking space character (encoded as `&nbsp;`) in order to create empty paragraphs. This character is ignored by the Markdown parser unless it’s encoded – just including the unicode character won’t force an extra paragraph. Python’s `html.escape` doesn’t encodes things like quotes:
```python
html.escape('"')
'&quot;'
```
But it doesn’t see a need to encoded non-breaking spaces:
```python
html.escape('\xa0')
'\xa0'
```

So to preserve the encoded non-breaking space we also have to temporarily convert it to something else, do the `unescape`/`escape` dance, and then convert it back.

### DIFF
--- a/notifications_utils/formatters.py
+++ b/notifications_utils/formatters.py
@@ -148,7 +148,7 @@ def strip_html(value):
 def escape_html(value):
     if not value:
         return value
-    value = str(value).replace('<', '&lt;')
+    value = str(value)
     # If the content contains a HTML encoded non-breaking space then we
     # need to protect it from the escape/unescape process by temporarily
     # turning it into something else
--- a/notifications_utils/formatters.py
+++ b/notifications_utils/formatters.py
@@ -4,6 +4,7 @@ import urllib
 
 import mistune
 import bleach
+from html import escape, unescape
 from itertools import count
 from flask import Markup
 from orderedset import OrderedSet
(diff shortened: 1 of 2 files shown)


quis:
This is slightly slower now, but I don’t think drastically enough to worry us.

# Just testing the escaping function

```
python -m timeit -s "from notifications_utils.formatters import escape_html; import string; s = (string.ascii_letters + string.punctuation) * 100" "escape_html(s)"
```

## Before 
```
50 loops, best of 5: 5.77 msec per loop
```

## After 
```
20 loops, best of 5: 9.24 msec per loop
```

# In the context of processing a whole email

```
python -m timeit -v -s "from notifications_utils.template import HTMLEmailTemplate; import string; t = {'content': (string.ascii_letters + string.punctuation) * 100, 'subject': 'foo', 'template_type': 'email'}" "str(HTMLEmailTemplate(t))"
```

## Before 
```
5 loops, best of 5: 38.7 msec per loop
```

## After 
```
5 loops, best of 5: 45.4 msec per loop
```

leohemsted (approved):

PR #818, opened 2020-12-14
Move polygon processing code from admin app

quis:
Code moved from the admin app here: https://github.com/alphagov/notifications-admin/blob/81b576acadaee60807dce386f8238a6e1376f098/app/broadcast_areas/polygons.py

This is so we can reuse it in the API. We want to make sure that we always pass the CBC polygons with a sensible point count, no matter what our clients give us.

There’s basically no unit testing of this code in the admin app, so this pull requests adds a bunch of new tests for the various properties.


idavidmcdonald (approved):
Nice one. Tests are a good addition as a sanity check.

This polygon code is complex (which is OK) and is an area we probably won't touch too often and is also probably only well understood by you. Therefore, If I was to suggest one improvement, adding a comment to most of the functions in here saying what it does would be useful. This would make it easier to understand what each one of them do which has taken me a while and I still don't fully understand some of them still.

For example `smooth`, I roughly get what it does but it's not explicitly written down what the function will do. A doc block type comment would be a good addition.

PR #821, opened 2020-12-15
Add more tests around bleach encoding bug

quis:
There were more characters than just the ? which triggered this bug. This commit adds some more test cases to validate that we are handling them all correctly.


quis:
Closing in favour of #823

PR #823, opened 2020-12-16
Stop using bleach for escaping HTML

quis:
When we first added sanitisation of user-inputted content<sup>1</sup> we chose a library called Bleach which both escapes and strips HTML from content.

Bleach allows things that are already encoded to be passed through. For example `G&amp;T` will be outputted as `G&amp;T` and rendered by a browser as _G&T_. `html.escape` escapes everything, so `G&amp;T` becomes `G&amp;amp;T` which gets rendered by a browser as _G\&amp;T_.

We are relying on Bleach’s behaviour so that one team can pass us non-breaking spaces as `&nbsp;` to control layout. I also suspect we will have some teams giving us URLs with `&amp;` already in them, and seeing _\&amp;_ in places you shouldn’t is pretty common (my TV does it with the programme descriptions on Amazon Prime Video). So it’s good that we pass this through rather than double-encoding it.

We recently added<sup>3</sup> `html.unescape` and `html.escape` before running Bleach to work around some buggy handling of things which looked like entities but aren’t.

This is problematic because `html.unescape` is way too greedy about what it interprets as an entity. This is because it follows the HTML specification<sup>4</sup> which, for backwards compatibility, allows some entities (like `&times;`) to be used without a trailing semicolon (as in `&times` 😱 ). This means when someone has a URL which includes `&timestamp=`, it gets rendered as `&×stamp=` (`&times;` is the entity for the multiplication symbol `×`, which is subtly different to the 24th letter of the alphabet, `x`).

---

In summary we:
- cannot use Bleach by itself
- cannot use Bleach in conjunction with `html.unescape`

So rather than trying to find another workaround for Bleach’s buggy behaviour this commit replaces it with `html.escape` alone, for escaping content. For stripping HTML from content Bleach is still good.

This commit also adds a tight list of allowed entities which we pass through by temporarily converting them to unlikely sequences of emoji. This means we can support `&nbsp;` and `&amp;`, without encountering unexpected behaviour when someone uses `&timestamp=`.

---

This change performs better than before. When tested with this script:
```bash
python -m timeit -v -s "from notifications_utils.formatters import escape_html; import string; s = (string.ascii_letters + string.punctuation) * 100" "escape_html(s)"
```

Before:
```
50 loops, best of 5: 9.11 msec per loop
```

After
```
10000 loops, best of 5: 37.1 usec per loop (0.0371 msec)
```
245\&times; faster 😉 

---

1. alphagov/notifications-api#135
2. https://docs.python.org/3/library/html.html#html.escape
3. https://github.com/alphagov/notifications-utils/pull/814
3. https://html.spec.whatwg.org/multipage/syntax.html#named-character-references


quis:
# Checks for content relying on 

I ran this query to find where teams are relying on escape sequences other than `&nbsp;` or `&amp;`:

```sql
select
    content
from
    templates
join
    services on templates.service_id = services.id
where
    content like '%&__;%'
    and templates.archived = false
    and services.restricted = false
;
```

## ✅ 2 character escape sequences  

No results

## ✅  3 character escape sequences 

25 instances of `&amp;`

2 instances of `&#39;` in one template, was a newsletter sent in May 2020 so shouldn’t be used again

## ⚠️ 4 character escape sequences  

16 instances of `&nbsp;`

18 instances of `&lpar;` (left opening bracket)

29 instances of `&rpar;` (right closing bracket)

16 instances of `&Ipar;` (misspelling of `&lpar;`)

***

Looks like people are using `&lpar;` and `&rpar;` to work around not being able to use brackets in the body of conditional placeholders, like this:

```
((proof of address not provided??* proof of address &lpar;council tax or utility bill, dated within the last 12 months or driving license&rpar; ))
```

This means that we can’t just unescape them to `(` and `)` because this would stop them being recognised as placeholders.

## ✅ 5 character escape sequences  

4 instances of `&#8226;` – bullet character (•) – in one template not used since April 2017

20 instances of `&#8200;` – punctuation space – in one template not used since May 2017 

2 instances of `&wcirc;` – w with circumflex (&wcirc;) – in one letter with test content

7 instances of `&rparr;` – misspelling of `&rpar;`

## ✅ 6 character escape sequences  

No results

## ✅ 7 character escape sequences  

2 instances of things that weren’t escape sequences, eg `Kent & Surrey;`

## ✅ 8 character escape sequences  

No results

## ✅ 9 character escape sequences

One instance of `Health & Social Care;`

## ✅ 10 character escape sequences

No results

## ✅ 11 character escape sequences

One template with stuff like `&nbsp;o&nbsp;`

## ✅ 12 character escape sequences

Similar to 11

## ✅ 13 character escape sequences

`Dependencies & Requirements;`

---

Going to stop at 13 since I don’t seem to be finding any more positive matches…

quis:
# Support arbitrary HTML entities (e7886aa)

Since we’re handling special cases like `&amp;` and `&lpar;` now we can unescape any other special characters to their unicode equivalent, thereby supporting them.

We need to be stricter than Python’s built in `html.unescape` in order to avoid the `&timestamp=` problem. So this commit re-implements part of the `html` module to be strict about requiring trailing semicolons (even though the HTML  spec<sup>1</sup>] allows some (😱) entities to be used without them).

I think this is worth doing because we don’t know what entities API users might be passing in with `((content))` and relying on, and if we don’t do this it’s a change in our behaviour.

This has a 10&times; performance hit, but that’s against the 245&times; performance improvement we’ve already made from removing Bleach.

## Test script
```bash
python -m timeit -v -s "from notifications_utils.formatters import escape_html; import string; s = (string.ascii_letters + '&lsquo;' + string.punctuation) * 100" "escape_html(s)"
```

## Before

```
5000 loops, best of 5: 50.6 usec per loop
```

## After

```
500 loops, best of 5: 453 usec per loop
```

1. https://html.spec.whatwg.org/multipage/syntax.html#named-character-references

CrystalPea (approved):
noice :)

PR #826, opened 2020-12-24
Add code to check length of broadcast templates

quis:
We currently don't validate length of broadcast messages, and will get an error back from the CBCs if we send them a message that is too long. Instead, we should add that validation when the user is saving the template on the UI.

There is a defined max length in characters for broadcasts, which is set at 1,395 character for GSM and 615 for unicode.

Broadcast messages follow the same encoding as SMS (GSM). Characters from the extended character set should as two characters. If there are any non-GSM characters we’ll send the whole thing as UCS-2 (unicode).

Adding this code to utils so we can re-use it in the admin and the API. 

I’ve also added some useful properties (like `.non_gsm_characters`) that we can use to write nice error messages.

In the future we might not expose all this complexity to our users, but it’s good to have it technically correct in this library.

***

https://www.pivotaltracker.com/story/show/176094622


notifications_utils/template.py
```diff
@@ -364,6 +361,31 @@ def __str__(self):
 class BaseBroadcastTemplate(BaseSMSTemplate):
     template_type = 'broadcast'
 
+    MAX_CONTENT_COUNT_GSM = 1_395
+    MAX_CONTENT_COUNT_USC2 = 615
```
leohemsted, line 365: Minor thing - it's [UCS2](https://en.wikipedia.org/wiki/Universal_Coded_Character_Set), not USC2

tests/test_template_types.py
```diff
@@ -2418,3 +2418,99 @@ def test_broadcast_message_from_event():
...
+@pytest.mark.parametrize('content', (
+    '^{}\\[~]|€'
+))
+def test_broadcast_message_double_counts_extended_gsm(content):
+    template = BroadcastMessageTemplate({
+        'template_type': 'broadcast',
+        'content': content,
+    })
+    assert template.encoded_content_count == 2
+    assert template.max_content_count == 1_395
```
leohemsted, line 2490: ooh didn't know python supported this. Nice trick

leohemsted (approved):
one tiny change

notifications_utils/template.py
quis: Nice one, I squashed it out

PR #837, opened 2021-01-26
Add < > to invalid character set for address validation

servingUpAces:
Add `<` and `>` symbols to the INVALID_CHARACTER regex for postal validation.

https://www.pivotaltracker.com/story/show/176610031


quis:
Do DVLA consider the address invalid if it has a `>` anywhere in it, or just at the start of the line? This change will only validate the latter.

servingUpAces:
To answer @quis: DVLA are only concerned with the character at the start of the line

tests/test_postal_address.py
```diff
@@ -239,6 +239,14 @@ def test_postcode(address, expected_postcode):
         ''',
         True,
     ),
+    (
+            r'''
+            >123 Example Straße
+            SW1A 1AA
+            Deutschland
+            ''',
+            True,
```
quis, line 248: Can you make the indentation match the other examples pls? Sorry for the pedantry…
```suggestion
        r'''
        >123 Example Straße
        SW1A 1AA
        Deutschland
        ''',
        True,
```

quis (approved):

PR #842, opened 2021-02-12
Add method to create broadcast template from content

quis:
In the API and admin app we now let users specify their own content without a template.

This commit adds a convenience method to get a `Template` instance without having to manually create a fake `dict` of template data.


benthorner (approved):
Makes sense to me. Very minor: have we thought about having a changelog for this repo? Or does the person merging the PR tend to propagate the new version themselves everywhere?

tests/test_template_types.py
```diff
@@ -2411,6 +2411,14 @@ def test_letter_preview_template_lazy_loads_images():
     ]
 
 
+def test_broadcast_message_from_content():
```
CrystalPea, line 2414: shouldn't the test be testing that we are getting a Template instance?

quis: That’s a good idea, makes the test a bit more robust, how about 26a58d8?

quis:
@benthorner generally the person merging the PR is responsible for the change, but it shouldn’t matter for a minor/patch change. I feel like there’s an unwritten rule that you should roll out your own major versions, rather than make other people deal with the breaking changes.

tests/test_template_types.py
CrystalPea: nice one 🎉

CrystalPea (approved):

PR #843, opened 2021-02-24
Check for country rules

servingUpAces:
Add method to check the attributes of international number so we can change the sender accordingly.

I still have some outstanding questions but I think if we check that the country has alpha: NO then we can use the Notify number rather than "Notify". 

Outstanding question.
What does alpha: REG mean?
Do we also need to check numeric? What does numeric: LIMITED mean?
Should we be sending from +447xxx instead of 07xxx?


CrystalPea:
> Should we be sending from +447xxx instead of 07xxx?

That does sounds reasonable.

tests/test_international_billing_rates.py
```diff
@@ -30,3 +31,11 @@ def test_international_billing_rates_are_in_correct_format(country_prefix, value
 
 def test_country_codes():
     assert len(COUNTRY_PREFIXES) == 214
+
+
+@pytest.mark.parametrize("number, expected",
+                         [('48123654789', False),
+                          ('140312345687', True),
+                          ('40123548897', False)])
```
idavidmcdonald, line 39: Optional, you could add some extra nice context to these test cases like


```
('48123654789', False), # Poland, code 48, alpha: 'YES'
('140312345687', True),
('40123548897', False), # Romania, code 40, alpha: REG
```

tests/test_international_billing_rates.py
```diff
@@ -30,3 +31,11 @@ def test_international_billing_rates_are_in_correct_format(country_prefix, value
 
 def test_country_codes():
     assert len(COUNTRY_PREFIXES) == 214
+
+
+@pytest.mark.parametrize("number, expected",
+                         [('48123654789', False),
+                          ('140312345687', True),
```
idavidmcdonald, line 38: What is this number? I can't see a country code starting '140' in our yaml file or on https://en.wikipedia.org/wiki/List_of_country_calling_codes#Zone_1:_North_American_Numbering_Plan and the same for just '14'

idavidmcdonald (commented):
So you raise some good questions that I think we need some answers to before merging this.

What does alpha: REG mean?
I have no idea but at the moment my hypothesis is that it means they only support a certain registered set of alpha sender names. If that is the case, then it's unlikely that they would therefore support alpha messages and we may need to have `use_numeric_sender` return `True` instead (which your code currently doesn't). I think Pete is one to ask on this.


Do we also need to check numeric? What does numeric: LIMITED mean?
Good question that I don't know the answer to. And what do we do when ALPHA is no and numeric is LIMITED/NO, like for example Peru? What number do we use then?

Should we be sending from +447xxx instead of 07xxx?
Unknown, how might we find out this answer?

tests/test_international_billing_rates.py
servingUpAces: Added comments and some numbers that are not normalised

notifications_utils/recipients.py
```diff
@@ -435,6 +435,11 @@ def get_billable_units_for_prefix(prefix):
     return INTERNATIONAL_BILLING_RATES[prefix]['billable_units']
 
 
+def use_numeric_sender(number):
+    prefix = get_international_prefix(normalise_phone_number(number))
+    return INTERNATIONAL_BILLING_RATES[prefix]['attributes']['alpha'] == 'NO'
```
idavidmcdonald, line 440: Optional: We may want to consider returning True for countries that have `REG` support as I don't know if they will get through or not on an alpha sender.

@yahoopete do you know please?

idavidmcdonald (approved):
I think this is good enough for now but yeah we may need to add some more fine grain controls to do with `REG` and the `numeric` field

notifications_utils/recipients.py
idavidmcdonald: I also wonder if we can get some more up to date data from our SMS providers about what is supported for each country as that file was last updated 4 years ago

PR #844, opened 2021-03-03
Check for invalid characters in entire address

servingUpAces:
As it turns out, invalid characters in any part of the address is not suitable for the software at the postal providers. This will check the
entire postal address for these characters.

### DIFF
--- a/notifications_utils/postal_address.py
+++ b/notifications_utils/postal_address.py
@@ -104,7 +104,7 @@ class PostalAddress():
     @property
     def has_invalid_characters(self):
         return any(
-            line.startswith(tuple(self.INVALID_CHARACTERS))
+            any([c in line for c in set(self.INVALID_CHARACTERS)])
             for line in self.normalised_lines
         )
 
--- a/tests/test_postal_address.py
+++ b/tests/test_postal_address.py
@@ -221,7 +221,7 @@ def test_postcode(address, expected_postcode):
         C@ity of Town
         SW1A 1AA
         ''',
-        False,
+        True,
     ),
     (
         '''


notifications_utils/postal_address.py
```diff
@@ -104,7 +104,7 @@ def has_valid_last_line(self):
     @property
     def has_invalid_characters(self):
         return any(
-            line.startswith(tuple(self.INVALID_CHARACTERS))
+            any([c in line for c in set(self.INVALID_CHARACTERS)])
```
quis, line 107: It will be a tiny bit faster to use a generator than build up a list in memory, and you can iterate over the string directly – no casting to set needed:
```suggestion
            any(c in line for c in self.INVALID_CHARACTERS)
```

leohemsted (approved):

notifications_utils/postal_address.py
quis: Alternatively you could do it with set intersection:
```suggestion
            set(line) & set(self.INVALID_CHARACTERS))
```

quis: **Original**
```
python -m timeit -v -s "s = 'abc3def'; q = '12345'" "any([c in s for c in set(q)])"
```
953nsec

***

**Generator** 
```
python -m timeit -v -s "s = 'abc3def'; q = '12345'" "any(c in s for c in q)"
```
680 nsec

***

**Set intersection**
```
python -m timeit -v -s "s = 'abc3def'; q = '12345'" "set(s) & set(q)"
```
827 nsec per loop

quis: Actually you can remove the parent loop now, because you don’t need to look at each line individually.

So you can replace the whole method with:
```python
    return any(c in self.normalised_lines for c in self.INVALID_CHARACTERS)
```

quis (changes requested):
I’m worried that we’re stopping people using `(` or `)` anywhere in the address. Fine to not have these at the start of a line, but I can imagine examples like 
> Los Angeles (CA)

quis:
Anonymised examples of ‘invalid’ characters in addresses we’ve sent letters to in the last 7 days:

Brackets:
> Example Company (UK) Ltd
> Example Company (formerly Sample Company)
> Example Company (part of Sample Group)
> Firstname Lastname (Treasurer)

Slashes:
> C/O Example LLP

At symbol
> Example Company @ Business Park

Comma
> City, County

quis:
@servingUpAces Do we still need to do something like this, or can we close this PR now?

servingUpAces:
I'm going to close this for now... we can revisit this if we get another complaint from the provider.

PR #848, opened 2021-03-12
Allow polygons to bleed by a custom distance

quis:
We want to show different amounts of bleed for different areas, because we know that the bleed is not uniform across different geographies.

This commit:
- adds a method to allow a custom bleed distance to be passed in
- refactors the existing code to use this method


idavidmcdonald:
> We want to show different amounts of bleed for different areas, because we know that the bleed is not uniform across different geographies.

Could you give some examples of this please? I wasn't aware of this. And how do we see this method being used by the admin/api? Just keen to gain a bit of context (although the code looks fine at a glance)

quis:
There’s a work-in-progress admin PR here: https://github.com/alphagov/notifications-admin/pull/3833

***

Each mast can handle a certain number of connections.

Therefore in built up areas the masts need to be closer together so that each one still has a manageable number of connections. Conversely, in rural areas the masts are further apart.

In built up areas the masts broadcast at a lower power, which travels less distance, so they don’t overlap too much.

This means that, on average, we will see more bleed and less accuracy in rural areas

Here’s the area around the office for example:

image

And here’s an area of Skye at the same zoom level:

image

notifications_utils/polygons.py
```diff
@@ -164,13 +164,16 @@ def simplify(self):
 
     @cached_property
     def bleed(self):
```
idavidmcdonald, line 166: What do you see as the use of this `bleed` property going ahead? Will we use it in the admin app ever (or should we?) or will it become redundant and instead we will use `bleed_by` always?

I'm cautious of having two similarly named methods, one that gives you a default bleed and one that gives you a custom bleed. I would prefer to have a single method that has say a default I think...

idavidmcdonald:
Thanks for the very useful context, I might need to see a bit more of the admin PR come together before making a call on this. No red flags in it but just interested in the interface between the two things (the admin app and the utils library).

YOu should share this stuff about bleeds and cell towers in the tech catch up by the way, it's interesting

notifications_utils/polygons.py
quis: Good call, I’ve removed the property in ea6393c

notifications_utils/polygons.py
```diff
@@ -82,7 +82,7 @@ def buffer_outward_in_degrees(self):
         '''
         return (
             # If two areas are close enough that the distance between
-            # them is less than the minimum bleed of a cell
+            # them is less than the typical bleed of a cell
```
idavidmcdonald, line 85: Optional: I wonder if the property name `approx_bleed_in_degrees` would now be more appropriate as something like `default_bleed_in_degrees` and if there is any updating we want to do of the description above that property? I don't feel too strongly though

notifications_utils/version.py
```diff
@@ -1 +1 @@
-__version__ = '43.9.1'
+__version__ = '43.10.0'
```
idavidmcdonald, line 1: Removing the `bleed` property makes this a breaking change.

idavidmcdonald (approved):

notifications_utils/version.py
quis:
```suggestion
__version__ = '44.0.0'
```

PR #849, opened 2021-03-12
Update isort config to match that in API

benthorner:
As discussed in [1]. This has no effect on the existing imports.

[1]: https://github.com/alphagov/notifications-api/pull/3175#issuecomment-795530323


quis (approved):

benthorner:
Thanks for speedy review @quis ⭐.

I've also got this one for the only (I think) other place we use isort: https://github.com/alphagov/notifications-admin/pull/3834

PR #851, opened 2021-03-19
Add methods to get the bounds and overlap of polygons

quis:
If I draw a polygon that covers 10% of the area of Wokingham, I can reasonably guess that it contains about 10% of the total phones in all of Wokingham.

This pull request gives us the tools to make that guess.

`Polygons.bounds` will allow us to build an [R-tree](https://en.wikipedia.org/wiki/R-tree), which is a very fast way of determining which polygons might overlap with another.

`Polygons.ratio_of_intersection_with(Polygons)` will let us calculate the exact percentage of overlap.


notifications_utils/polygons.py
```diff
@@ -233,6 +250,23 @@ def estimated_area(self):
...
         ) * self.square_degrees_to_square_miles
 
+    def ratio_of_intersection_with(self, polygons):
+        '''
+        Given another Polygons object, this works how much the two
+        overlap, as a fraction of the area of this Polygons object.
+        '''
+        if self.estimated_area == 0:
+            return 0
+        return sum(
```
benthorner, line 260: Is there a risk of double-counting here if multiple polygons overlap the same area? For example, if someone selections "England" + "Hackney Marshes", then a subsequent overlap would count their shared area twice.

It looks like the intended use of this is to calculate the overlap of a drawn polygon over bounding boxes at the same level in an R Tree (from a quick skim of the PR linked to this one), so there doesn't seem to be a risk in that scenario?

However, since this is meant to be a generic utility, it's worth making the non-multi-overlapping assumption explicit in the method description.

notifications_utils/polygons.py
```diff
@@ -233,6 +250,23 @@ def estimated_area(self):
             polygon.area for polygon in self
         ) * self.square_degrees_to_square_miles
 
+    def ratio_of_intersection_with(self, polygons):
+        '''
+        Given another Polygons object, this works how much the two
+        overlap, as a fraction of the area of this Polygons object.
```
quis, line 256:
```suggestion
        overlap, as a fraction of the area of this Polygons object.
        It assumes that neither of the objects already contain 
        overlapping polygons.
```

notifications_utils/polygons.py
quis: Yeah, this is a risk more generally so I’ve added a note in 524c45d

## For custom polygons 

At some point we should add validation for user-created polygons. At the moment we have no guards against:
- polygons which overlap each other
- polygons which self-intesect (hourglass or bow tie shaped polygons)

We can check with the Flood Warning team to see what kind of problems they check for in their data set and apply the same checks.

# For our library

Nothing stops double-counting at the moment – here’s Hackney alone:

image

And here it is with some of its wards

image

tests/test_polygons.py
```diff
@@ -246,3 +254,25 @@ def test_bounds(polygons, expected_bounds):
...
+    ([LEA_VALLEY], [ISLE_OF_DOGS], 1.187),
+    # …but as a proportion of the area of the Isle of Dogs, the overlap
+    # is larger
+    ([ISLE_OF_DOGS], [LEA_VALLEY], 7.115),
+    # Ratio is always 0 if one or both polygons are empty
+    ([ISLE_OF_DOGS], [], 0),
+    ([], [ISLE_OF_DOGS], 0),
+    ([], [], 0),
+))
+def test_intersection_ratio(parents, children, expected_intersection_percentage):
```
benthorner, line 276: Minor: these names don't accurately describe the data they represent, as some of the params swap between supersets and subsets, and overlaps.

We could just say `polygons_1` and `polygons_2` or similar.

notifications_utils/polygons.py
benthorner: Oh dear 😬. Thanks for adding the warning, and it's good to know it's a general problem - then we can work towards a general fix for it.

benthorner (approved):
Nice, looks good. Just checking, is this part of [this card](https://www.pivotaltracker.com/story/show/176090635)?

Thanks for addressing my comment - I noticed one other minor thing, which I've added a comment for in case you're interested. Also on a minor note, I realised the behaviour of the ratio is unintuitive for me: normally I expect "the thing on the right" to be the divisor, whereas we've got it backwards; it kind of comes out in the tests, where the areas in each comment are phrased in the reverse order of the constants they refer to. I'm sure I'll get used to it, though.

quis:
@benthorner it will be useful in the future for that card, but will be useful now for users of the API.

And thanks for the reviewing 😄

PR #855, opened 2021-03-29
Allow double hyphens in domain part of email addresses

CrystalPea:
We confirmed with Amazon SES that they allow sending to such domains.


notifications_utils/__init__.py
```diff
@@ -6,7 +6,7 @@
 # regexes for use in recipients.validate_email_address.
 # Valid characters taken from https://en.wikipedia.org/wiki/Email_address#Local-part
 # Note: Normal apostrophe eg `Firstname-o'surname@domain.com` is allowed.
-hostname_part = re.compile(r'^(xn-|[a-z0-9]+)(-[a-z0-9]+)*$', re.IGNORECASE)
+hostname_part = re.compile(r'^(xn-|[a-z0-9]+)(-??-[a-z0-9]+)*$', re.IGNORECASE)
```
benthorner, line 9: Could you explain what this regex is doing in the bit that's changed? My understanding is that `-?` means "an optional `-`", but I'm not sure how to interpret `-??`.

CrystalPea: Sure thing, `?? | Matches the previous element zero or one time, but as few times as possible.` ( https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-language-quick-reference )

But you are right, single question mark works, too, so I will use that instead, thanks!

benthorner: Ah, I see. Yeah, that shouldn't be an issue here, since we don't have subsequent `-` characters in the regex.

I was looking to see what the old regex matches:

- `xn--something`
- `something-else`

Slight issue with the new regex is that it will allow `xn---something`, which is invalid Punycode (apparently). One way (/hack) to avoid this would be to use the following i.e. remove the initial hyphen: `^(xn|[a-z0-9]+)(-?-[a-z0-9]+)*$`. It's a bit of a hack because, compared to the TLD regex, it's less explicit about the relationship between `xn` and `--`.

How about adding a couple of test cases to cover the valid/invalid cases around this.

CrystalPea: Good point, I tweaked the regex, added a comment for clarity and an extra test case for invalid punycode domain.

PR #869, opened 2021-05-24
Add ~ as an invalid character for a postal address to start with.

servingUpAces:
The postal provider has indicated that `~` is an invalid character for an address to start with.


idavidmcdonald (approved):
So this looks correct and continues the existing patterns.

However, I would suggest whilst we are here we may want to consider a small improvement.

By renaming `INVALID_CHARACTERS` into `INVALID_FIRST_CHARACTERS_IN_LINE` so it's clear that these characters are only checked on the very first character of an address line. As everytime I see this PR I go searching for ' ah, is this for every character, or just the first` and then need to delve into the code to double check what `INVALID_CHARACTERS` is being used for

PR #872, opened 2021-06-11
Round coordinates to 5 decimal places

quis:
According to Pete D:

> [polygons should have] precision of 1 metre (5dp for lat/lng). The 1 metre (5dp) is not contractual but based on “best practice”/experience from other systems

This will have nice side effect of reducing the amount of data we have to store in the admin app.


notifications_utils/polygons.py
```diff
@@ -217,7 +219,10 @@ def as_coordinate_pairs_long_lat(self):
         order, for example leaflet.js.
         '''
         return [
-            [[x, y] for x, y in polygon.exterior.coords]
+            [[
+                round(x, self.output_precision_in_decimal_places),
+                round(y, self.output_precision_in_decimal_places),
+            ] for x, y in polygon.exterior.coords]
             for polygon in self
         ]
```
servingUpAces, line 228: Do you also need to add the precision to the next method `as_coordinate_pairs_lat_long`?

quis: That method re-uses the output from this one, so it will get the same precision automatically.

servingUpAces: So it does - soz

servingUpAces (approved):

PR #873, opened 2021-06-15
Preserve a bit more detail in concave parts of polygons

quis:
After analysing the polygon for the Reading test we have decided that we’re being a bit too aggressive with how much we are smoothing it.

We still need to do a lot of smoothing to group together very disparate areas like the Western Isles of Scotland.

The limits we’re targeting are:
- Max 6000 points overall (we don’t need anywhere near this detail)
- Max 25 polygons (this is the one that’s harder to hit)

This change increases the maximum number of polygons in a single area from 6 to 9. That’s still comfortably within the limit of 25. We need to be comfortable because we can’t be sure what will happen when people start combining different areas.

Before | After
---|---
Screenshot 2021-06-15 at 09 14 43 | Screenshot 2021-06-15 at 09 07 17
Screenshot 2021-06-15 at 09 13 30 | Screenshot 2021-06-15 at 09 08 26

**Difference**

Red is ‘before’, blue is ‘after’.

unnamed-1


tombye (approved):
This is a pretty new part of the codebase for me, not all of which I totally understand so here are the reasons I'm using to say I think this is good to go:
1. the comments before `perimeter_to_buffer_ratio` explain why bumping it up reduces the buffering
2. using the admin app to find areas with similar perimeters to those in the tests and checking their buffer out/in metres match the test values when I change `perimeter_to_buffer_ratio`
3. I don't really understand how changing `perimeter_to_buffer_ratio` alters the maximum number of polygons in an area but the before/after numbers in the description show the change does what it says it should for (I assume) our data

👍🏻

PR #877, opened 2021-06-18
Allow initial content of Zendesk tickets to be hidden from the requester 

quis:
Sometimes we want to raise a ticket but not let the person raising it see the auto-generated content of the ticket.

We can achieve this by making the first message on the ticket be an ‘internal note’, rather than a ‘public reply’.

The example of this is where we want to add go live notes to an organisation, which need to appear in the ticket but might not be information that we want to divulge to an individual service.


benthorner (commented):
Change looks good 👍. Could you help me understand how we'll use it? If I understand correctly, this could mean the user won't have any context of the ticket, since none of the initial content will be visible. Do we want to hide everything, or just some things?

quis:
The intention would be to set `requester_sees_message_content=False` for request to go live tickets only. I’m not sure hiding everything will be totally problem-free, but mine and Ale’s intention was to try it and see how it goes.

benthorner (approved):
That sounds OK, thanks for explaining. 

It's fair to say that users generally only do one go-live request at a time, and remember which service it was for, so I agree this is unlikely to cause a problem for the majority of tickets.

PR #878, opened 2021-06-18
Detect optional placeholders with brackets in their text 

CrystalPea:
This didn't work before.

Here is the story ticket: https://www.pivotaltracker.com/story/show/178336216

And here is a screenshot:

Screenshot 2021-06-18 at 15 37 22


notifications_utils/field.py
```diff
@@ -63,8 +63,8 @@ class Field:
     """
 
     placeholder_pattern = re.compile(
-        r'\({2}'        # opening ((
-        r'([^()]+)'     # body of placeholder - potentially standard or conditional.
+        r'\({2}(?=[^(])'        # opening ((
+        r'.+?(?=\)\))'     # body of placeholder - potentially standard or conditional.
```
benthorner, line 67: Minor: we could use `\){2}` on the end, to match the opening regex.

notifications_utils/field.py
```diff
@@ -63,8 +63,8 @@ class Field:
     """
 
     placeholder_pattern = re.compile(
-        r'\({2}'        # opening ((
-        r'([^()]+)'     # body of placeholder - potentially standard or conditional.
+        r'\({2}(?=[^(])'        # opening ((
```
benthorner, line 66: It took me a while to understand what this is doing: we need to account for three opening brackets. Can we extend the comment to explain?

notifications_utils/field.py
benthorner: We can simplify this to just `.+?`. It's also worth adding a comment about the purpose of the optional / non-greedy `?` modifier.

tests/test_field.py
```diff
@@ -111,6 +111,11 @@ def test_returns_a_string_without_placeholders(content):
             {"warning": False},
             ""
         ),
+        (
+            "((warning??This is a conditional warning (with more info in brackets)))",
```
benthorner, line 115: Minor: we could also have a test for bracketed text in the middle of the conditional string - this is a great edge case, though 💯.

benthorner (commented):
Ooh, so tricky! Nice one figuring out a solution - the `(((colour)))` case threw me 🤯.

I had a play, and think we can make it a bit simpler. Either way, we should update the comments, so it's clear what the more complex regexes are doing.

notifications_utils/field.py
CrystalPea: >  Minor: we could use \){2} on the end, to match the opening regex.

Done now ☺️ 

> We can simplify this to just .+?. It's also worth adding a comment about the purpose of the optional / non-greedy ? modifier.

Sorry, I don't understand what you want me to do here.

quis:
Can you do a basic check that this doesn’t have a big performance impact, for example when processing a 100,000 row CSV. I feel like using stuff like `?` in regexes can make things a bit slower.

notifications_utils/field.py
benthorner: Haha, neither did I at first 😆. After reading through it again, I'm asking for two things:

- We can change `r'.+?(?=\)\))'` to just `r'.+?'` (all the tests still pass).
- Extend the comment to explain why we need the `?` in `.+?`.

The last point is because I found it hard to understand what the `?` is doing: the tests fail without it, but it wasn't clear why. I then found this in the docs: `Adding ? after the qualifier makes it perform the match in non-greedy or minimal fashion`. So without it we'd match `((something)))` from `text with (((something))) in brackets` instead of the correct `((something))`. It has a similar, but opposite, purpose to your change on line 66.

Does that help?

CrystalPea: Wow, you are right, the simpler regex works, too, amazing! 💖 

I added an explanation about the question mark in the commit message. I feel it is a bit too lengthy and technical for an inline comment.

benthorner (approved):
Code looks good 👍.

@quis makes a good point that the new regex could be slow. I did a quick test of this for a single placeholder, and the difference seems to be minimal.

```
>>> timeit.timeit(lambda: re.sub(new_regex, fn, '((warning?? This is a warning (with more info in brackets) and some more))'), number=100000)
0.34725685900775716

>>> timeit.timeit(lambda: re.sub(old_regex, fn, '((warning?? This is a warning (with more info in brackets) and some more))'), number=100000)
0.34456419400521554
```

PR #881, opened 2021-07-01
Ensure simplification doesn’t make invalid polygons

quis:
Recently we reduced the number of decimal places in each coordinate. This has introduced two problems:

# Bowties 

Sometimes rounding the precision of the polygons causes the points to shift  around in such a way that they make a bowtie shape, which is not a  valid geometry:

Screenshot 2021-07-01 at 11 17 15

We can fix this by increasing the precision by 1 decimal place.

# Polygons of 0 area

Very small areas now get rounded to have an area of 0, because all their points are closer together than the rounding precision (also not a valid geometry):

image

We can fix this by ignoring areas that are very small and would get rounded to an area of 0 (these are likely artefacts of the simplification process which we hadn’t spotted before)


notifications_utils/polygons.py
```diff
@@ -168,7 +168,13 @@ def smooth(self):
         flattened = list(itertools.chain(*[
             flatten_polygons(polygon) for polygon in debuffered
         ]))
-        return Polygons(flattened)
+        return Polygons([
+            polygon for polygon in flattened if (
+                # The smoothing process creates some artifacts which can
+                # be removed by ignoring polygons less than 1m² in area
+                polygon.area > (1 / self.approx_square_metres_to_square_degree)
```
CrystalPea, line 175: should we have a test for this?

quis: Removing it will cause this bit in the admin app to raise an exception: https://github.com/alphagov/notifications-admin/blob/256c840b46d4095edbd0af105e5c1a340242575d/app/broadcast_areas/create-broadcast-areas-db.py#L140

It’s a bit hard to test those lines in isolation because they are part of a larger method. But I’ve added something which hopefully covers the expected behaviour in 36682a2

CrystalPea (approved):

PR #884, opened 2021-07-09
Revert changes to placeholder regexes

CrystalPea:
Reverts: https://github.com/alphagov/notifications-utils/pull/878/files

This earlier PR introduced a bug for multiline conditional placeholders


klssmith (approved):
Good to go if this is tested locally too to be extra sure it fixes the issue

PR #885, opened 2021-07-13
Test detection and formatting of conditional placeholder with line-break in it

CrystalPea:
We recently had an incident where changes in the code broke optional placeholders with linebreaks in them.

We should protect against that in the future by having unit tests checking for those cases.


benthorner (approved):
👌 

(I tested this fails as expected with the broken version of the code.)

PR #886, opened 2021-07-23
Refactor to have a separate function that checks if too late to cancel letter

CrystalPea:
This is so we can show a better error messages to our users.

Needs to go in before:
- https://github.com/alphagov/notifications-api/pull/3296
- https://github.com/alphagov/notifications-admin/pull/3984


benthorner (approved):
Looks good 👍.

I was wondering: did you consider having the error messages directly in here? That would mean we could avoid the API app having to have its own code to figure why a letter can't be cancelled. If we did that approach, the method here would look something like:

```
def check_letter_can_be_cancelled(...):
   if [...]:
      return (False, "because X")

   if [...]:
      return (False, "because Y")

   return (True, "")
```

It's an optional suggestion. What got me thinking about it was the fact that the new method isn't called directly in the tests. That's OK if we take a whitebox test approach, but I try and avoid that as it's a bit risky as e.g. the way we call the new method could change but the tests would still pass. Having the flag + message in one place is a way to avoid that.

I also appreciate this is meant to be a quick bug fix, so happy to leave it as-is.

CrystalPea:
Hey @benthorner I like your suggestion, but also I feel I won't have time to implement it as I want to work on other stuff. So I will merge this as is, but we can think of a little refactor at some point?

PR #889, opened 2021-08-05
Convert polygons to a cartesian coordinate system before performing geometry operations on them

quis:
> Even though the Earth is not flat – and for that matter not exactly spherical – there are many analytic problems that can be approached by transforming Earth features to a Cartesian plane, applying tried and true algorithms, and then transforming the results back to geographic coordinates. This practice is as old as the tradition of accurate paper maps.
>
> Shapely does not support coordinate system transformations. All operations on two or more features presume that the features exist in the same Cartesian plane.

– https://shapely.readthedocs.io/en/stable/manual.html#coordinate-systems

***

At the moment, we are not doing any transformation of features before applying geometric algorithms to them. This is, in effect, assuming that the earth is flat.

This pull request implements the transformation of our polygons to a Cartesian plane. In other words, it converts them from being defined in spherical degrees to metres. 

This is important because spherical degrees correspond to a different linear distance depending on:
- where on the Earth’s surface you are
- whether you are measuring a north/south or east/west distance

Thus far we have got around this by saying 1˚ is approximately 111km and pretending that our coordinates are not spherical. However in the north of Scotland 1˚ measured east/west is only about 50km. So when we say ‘expand a shape by 5km in all directions’ it will only expand by ~2km in the east/west direction. You can see this distortion of the estimated bleed highlighted in the _Before_ screenshot here:

Before | After
---|---
image002 | image (7)

Shapely itself does not support coordinate transformations. So we need a new tool, Pyproj, to:
- convert the spherical coordinates to linear metres when needed
- decide which coordinate reference system to use for the transformation (our polygons cover 4 different zones)

***

Best reviewed commit-by-commit.


notifications_utils/polygons.py
```diff
@@ -91,12 +113,20 @@ def bounds(self):
...
+        )
+
         return (
-            min(all_min_x), min(all_min_y), max(all_max_x), max(all_max_y),
+            min_x_wgs84, min_y_wgs84, max_x_wgs84, max_y_wgs84,
         )
 
     @cached_property
-    def buffer_outward_in_degrees(self):
+    def buffer_outward_in_m(self):
```
CrystalPea, line 194: the docstring a line below needs updating from `degrees` to `metres`

notifications_utils/polygons.py
```diff
@@ -108,32 +138,32 @@ def buffer_outward_in_degrees(self):
...
             # small shapes like Orkney or the Isles of Scilly.
-            self.approx_bleed_in_degrees / 3
+            self.approx_bleed_in_m / 3
         ) + (
             self.perimeter_length / self.perimeter_to_buffer_ratio
         )
 
     @cached_property
-    def buffer_inward_in_degrees(self):
+    def buffer_inward_in_m(self):
```
CrystalPea, line 212: same here about docstring

notifications_utils/polygons.py
```diff
@@ -243,9 +278,7 @@ def estimated_area(self):
         approximate conversion of degrees to square miles for UK
         latitudes, rather than a projection.
         '''
-        return sum(
-            polygon.area for polygon in self
-        ) * self.square_degrees_to_square_miles
+        return sum(polygon.area for polygon in self)
```
CrystalPea: so here we were operating in miles, but now we operate in metres, is it something that could cause trouble?

notifications_utils/polygons.py
```diff
@@ -10,8 +12,20 @@
 from shapely.ops import unary_union
 from werkzeug.utils import cached_property
 
-wgs84_to_utm29n = Transformer.from_crs('EPSG:4326', 'EPSG:32629', always_xy=True)
-utm29n_to_wgs84 = Transformer.from_crs('EPSG:32629', 'EPSG:4326', always_xy=True)
+transformers = {
+    utm_code: {
+        'from_wgs84': Transformer.from_crs('EPSG:4326', utm_code, always_xy=True),
```
CrystalPea, line 17: it would be useful to add a comment explaining what wgs84 is, what do you think?

CrystalPea (commented):
Well done, that looks good and must have required some serious research 👏🏼 

I left a few comments, one of them is a question I'd like answered before this goes in, but generally I'm impressed ☺️

notifications_utils/polygons.py
quis: I think the only place which uses the value as an absolute number is the admin app here:
https://github.com/alphagov/notifications-admin/blob/9ba37249a47f08b1396e63a34bc591c781739686/app/templates/views/broadcast/macros/area-map.html#L6

We’ll need to change it to do its own conversion from square metres to square miles

notifications_utils/polygons.py
quis: Good spot, I’ve fixed this as squashed it into 2da2fa78cd570fd86a46f021853021abcbe5788b

notifications_utils/polygons.py
quis: Fixed and squashed into 2da2fa78cd570fd86a46f021853021abcbe5788b

notifications_utils/polygons.py
quis: Comment added in e2faacf

CrystalPea: Thank you so much, these are beautiful comments, they give all the missing context 👏🏼

notifications_utils/polygons.py
CrystalPea: cool stuff, do we have this written on a ticket or somewhere so we don't forget?

CrystalPea (approved):

notifications_utils/polygons.py
quis: I’ve got a work in progress PR on admin which will address the breaking changes: https://github.com/alphagov/notifications-admin/pull/3994

CrystalPea: Awesome stuff 💖

notifications_utils/polygons.py
```diff
@@ -45,15 +67,91 @@ class Polygons():
...
     output_precision_in_decimal_places = 6
 
-    def __init__(self, polygons):
-        if not polygons:
-            self.polygons = []
-        elif isinstance(polygons[0], list):
-            self.polygons = [
-                Polygon(polygon) for polygon in polygons
-            ]
+    def __init__(self, polygons, utm_crs=None):
```
benthorner, line 70: Just coming to this as part of reviewing a follow-up PR. Us in the future won't (easily) have the benefit of the PR description to work out what "utm_crs" and "utm_polygons" means. We should add a comment on this and the `utm_polygons` method to explain what they're doing:

- The input data may be the WSG84 coordinates (would really help to mention "traditional lat/lon degrees" here) based on what we store or UTM ("cartesian plane") coordinates in metres - it's hard to find [a reference for metres](https://pyproj4.github.io/pyproj/stable/api/transformer.html#pyproj-transform).

- We use the projected UTM coordinates in memory to perform calculations on polygons in terms of metres. The Shapely library also requires coordinates are in a cartesian plane to perform accurate transformations.

quis: Added in https://github.com/alphagov/notifications-utils/pull/892

PR #890, opened 2021-08-09
Refactor smoothing for reuse and fewer artefacts

quis:
Our polygon smoothing works by expanding and then contracting each shape.

Because it was using different parameters (resolution and join style) for the expansion and contraction it was introducing artefacts, i.e. it was creating entirely new shapes. This seemed to be mostly happening when the original polygons were very small, like those used by the Flood Warning Service.

This commit refactors the smoothing to re-use our existing `bleed_by` method to expand and contract the polygon. By re-using this method we have the same parameters for both the expansion and contraction. This seems to prevent the artefacts.

# `original_polygons`

Screenshot 2021-08-09 at 10 16 47

# `original_polygons.smooth.simplify` (before)

Note the artefacts around the centre of Penzance and St Michael’s Mount

Screenshot 2021-08-09 at 10 18 37

# `original_polygons.smooth.simplify` (after)

Screenshot 2021-08-09 at 10 19 33

# Side by side 

Original | Before | After
---|---|---
Screenshot 2021-08-09 at 10 16 47 | Screenshot 2021-08-09 at 10 18 37 | Screenshot 2021-08-09 at 10 19 33

# Performance

```shell
python -m timeit -s "from notifications_utils.polygons import Polygons; from tests.test_polygons import LEA_VALLEY" "Polygons([LEA_VALLEY]).smooth"
```

## Before 

1000 loops, best of 3: 189 usec per loop

## After 

1000 loops, best of 3: 210 usec per loop

Small but not significant slowdown, likely caused by `union`ing the polygons twice (once per call to `bleed_by`) rather than once. Here’s what the internals of the `bleed_by` method look like:
https://github.com/alphagov/notifications-utils/blob/83ed2955876de88e3e14a107a48ec4271aba80d9/notifications_utils/polygons.py#L190-L202

### DIFF
--- a/notifications_utils/polygons.py
+++ b/notifications_utils/polygons.py
@@ -148,33 +148,13 @@ class Polygons():
         these areas we can preserve it in places where it’s more
         relevant.
         '''
-        buffered = [
-            polygon.buffer(
-                self.buffer_outward_in_degrees,
-                resolution=4,
-                join_style=JOIN_STYLE.round,
-            )
-            for polygon in self
-        ]
-        unioned = union_polygons(buffered)
-        debuffered = [
-            polygon.buffer(
-                -1 * self.buffer_inward_in_degrees,
-                resolution=1,
-                join_style=JOIN_STYLE.bevel,
-            )
-            for polygon in unioned
-        ]
-        flattened = list(itertools.chain(*[
-            flatten_polygons(polygon) for polygon in debuffered
-        ]))
-        return Polygons([
-            polygon for polygon in flattened if (
-                # The smoothing process creates some artifacts which can
-                # be removed by ignoring polygons less than 1m² in area
-                polygon.area > (1 / self.approx_square_metres_to_square_degree)
-            )
-        ])
+        return self.bleed_by(
+            self.buffer_outward_in_degrees
+        ).bleed_by(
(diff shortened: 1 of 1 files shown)


rjbaker (approved):
```
Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
   144     31.4 MiB     31.4 MiB           1       @cached_property
   145                                             @profile
   146                                             def smooth(self):
   147                                                 '''
   148                                                 Fills in areas which aren’t covered by the polygons, but would
   149                                                 likely receive the broadcast anyway because of the bleed. This
   150                                                 includes very convex areas like harbours, and places where two
   151                                                 polygons are close to touching each other. By removing detail in
   152                                                 these areas we can preserve it in places where it’s more
   153                                                 relevant.
   154                                                 '''
   155     31.8 MiB      0.3 MiB           4           return self.bleed_by(
   156     31.4 MiB      0.0 MiB           1               self.buffer_outward_in_degrees
   157                                                 ).bleed_by(
   158     31.8 MiB      0.0 MiB           1               -1 * self.buffer_inward_in_degrees
   159                                                 ).remove_smaller_than(
   160     31.8 MiB      0.0 MiB           1               area_in_square_metres=1
   161                                                 )

PR #891, opened 2021-08-18
Only transform Polygon objects back to WGS84

quis:
Internally the `Polygons` class can either store:
- a list of `Polygon` objects, with a mandatory `utm_crs` code which  says how to convert them back to degrees
- a list of coordinate lists, with an optional `utm_crs` code which speeds up any future conversion to metres (if the code is not provided then it has to be derived)

When we output the polygons to display on a map or send to a CBC we need they must use degrees as their unit. This means that, if they are stored as `Polygon` objects, we need to first convert them back to degrees. If they are stored as lists of coordinates then we should just return the list, even if we also have a `utm_crs` code to use in future conversions to metres.

In other words the presence of a `utm_crs` code isn’t a good indicator of whether the polygons are stored in degrees or metres, but the `type` of the objects is.


CrystalPea:
what is utm and what is crs?

rjbaker:
> what is utm and what is crs?

utm_crs stands for Universal Transverse Mercator Coordinate Reference System: https://docs.qgis.org/2.18/en/docs/gentle_gis_introduction/coordinate_reference_systems.html#universal-transverse-mercator-utm-crs-in-detail

quis:
UTM is [universal transverse mercator](https://en.wikipedia.org/wiki/Universal_Transverse_Mercator_coordinate_system), a coordinate system based on the [transverse mercator projection](https://en.wikipedia.org/wiki/Transverse_Mercator_projection) which flattens the earth into a map that looks like this:

MercTranSph

CRS is a coordinate reference system, a set of functions which say how to translate coordinates expressed in degrees around the earth (in other words a 3 dimensional measure) into coordinates on the UTM projection (which is a 2 dimensional [Cartesian plane](https://en.wikipedia.org/wiki/Cartesian_coordinate_system)).

rjbaker (approved):

CrystalPea:
Thank you both!

PR #892, opened 2021-08-20
Add additional comments to explain UTM in `Polygons` class

quis:
Based on review comment here: https://github.com/alphagov/notifications-utils/pull/889#pullrequestreview-734960591

### DIFF
--- a/notifications_utils/polygons.py
+++ b/notifications_utils/polygons.py
@@ -68,6 +68,17 @@ class Polygons():
     output_precision_in_decimal_places = 6
 
     def __init__(self, polygons, utm_crs=None):
+        """
+        This class accepts either:
+        - a list of lists of coordinate pairs defined in WGS84 degrees
+        - a list of lists of coordinate pairs defined in WGS84 degrees,
+          with the name of a coordinate reference system
+        - a list of shapely.geometry.Polygon objects defined in metres,
+          with the name of a coordinate reference system
+
+        See the comments around `transformers` for an explanation of
+        WGS84 and coordinate reference systems.
+        """
 
         if not isinstance(polygons, list):
             raise TypeError(
--- a/notifications_utils/polygons.py
+++ b/notifications_utils/polygons.py
@@ -25,8 +25,10 @@ transformers = {
         # These are the names of coordinate reference systems, which
         # provide mathemtical functions for transforming WGS84
         # coordinates to linear measures of distance across the earth’s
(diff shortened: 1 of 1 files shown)


rjbaker (approved):

notifications_utils/polygons.py
```diff
@@ -68,6 +70,17 @@ class Polygons():
     output_precision_in_decimal_places = 6
 
     def __init__(self, polygons, utm_crs=None):
+        """
+        This class accepts either:
+        - a list of lists of coordinate pairs defined in WGS84 degrees
+        - a list of lists of coordinate pairs defined in WGS84 degrees,
```
benthorner, line 76: How come we need this? As I understand it, a `utm_crs` is only required when we've projected the WGS84 coordinates onto a plane.

notifications_utils/polygons.py
```diff
@@ -25,8 +25,10 @@
         # These are the names of coordinate reference systems, which
```
benthorner, line 25: We don't mention "degrees" (or lat / long) anywhere. Ideally we should avoid people have to research WSG84 to find out what the units are in order to understand these comments.

Could we mention it somewhere? Even better, could we mention it in [lots](https://github.com/alphagov/notifications-utils/blob/943ceeef60327596990af32702ae89d6bf9891a5/notifications_utils/polygons.py#L346) [of](https://github.com/alphagov/notifications-utils/blob/943ceeef60327596990af32702ae89d6bf9891a5/notifications_utils/polygons.py#L161) [places](https://github.com/alphagov/notifications-utils/blob/943ceeef60327596990af32702ae89d6bf9891a5/notifications_utils/polygons.py#L113) so it's really easy for someone to read through this class and understand what's going on? Most of the methods for the class have comments to explain what we do, so it looks like there's a precedent.

benthorner (commented):
Thanks for adding these in. I wonder if we're missing an abstraction here, as `Polygons` seems to be doing a whole mishmash of things. Maybe `UTMPolygons` and `WSG84Polygons`? I find it quite hard to remember what we mean by "polygons" at any given point, especially with all the acronyms flying around. An abstraction might help with that.

Something for the future, though. In the meantime, I think we should go further here and help ourselves as much as we can by adding some brief comments on the new methods. Although some of them aren't used externally, a brief explanation of each one terms in simple terms still has value for someone making changes to this class in future.

PR #893, opened 2021-08-31
Reflect correctly pluralised exception message

quis:
As of https://github.com/python/cpython/commit/4a12a178f4a6b9a59d97fecc727f2b6b28dfc85f the exception message for instantiating a class with abstract methods is only pluralised if there is more than one abstract method.

Our tests were expecting the exception message to always be pluralised.

This means the tests fail on Python versions greater than 3.9.0. This commit makes the tests pass with both (incorrectly) pluralised and non-pluralised versions of the message.

### DIFF
--- a/tests/test_serialised_model.py
+++ b/tests/test_serialised_model.py
@@ -1,3 +1,5 @@
+import sys
+
 import pytest
 
 from notifications_utils.serialised_model import (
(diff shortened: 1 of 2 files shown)


tests/test_serialised_model.py
```diff
@@ -19,27 +19,35 @@ class CustomCollection(SerialisedModelCollection):
...
+    ) or str(e.value) == (
+        "Can't instantiate abstract class SerialisedModel with abstract method ALLOWED_PROPERTIES"
     )
 
     with pytest.raises(TypeError) as e:
         Custom()
 
     assert str(e.value) == (
         "Can't instantiate abstract class Custom with abstract methods ALLOWED_PROPERTIES"
+    ) or str(e.value) == (
```
benthorner, line 31: Could we just check if the first part of the message is in `value`? I'm not sure what value we get from ensuring the entire message is just-so.

If we do keep both, we should have a comment on the new alternative to clarify it's for 3.9+, as it's unclear why we have both without looking at the commit.

quis: There’s value in the second part of the message as well, i.e. that it’s referencing the right property. I‘ve added comments in c08f315 which make it clear what can be removed when, while keeping whole assertions.

leohemsted:
i think the intended pytest way to test this would be to parametrise and use xfail (or skipif). Not sure if that's actually nicer - it's more verbose but definitely more explicit (and will make sure the tests aren't ambiguous if it turns out that when upgrading to a new python the error message _doesn't_ change

https://docs.pytest.org/en/6.2.x/skipping.html?highlight=sys%20version_info#skip-xfail-with-parametrize

```python
@pytest.mark.parametrize('expected_exception_msg', [
  pytest.param(
    'Can't instantiate abstract class SerialisedModel with abstract method ALLOWED_PROPERTIES', 
    marks=pytest.mark.xfail(
      sys.version_info < (3, 9),
      reason='python versions after 3.9 change the ABC exception message ')
  ),
  pytest.param(
    'Can't instantiate abstract class SerialisedModel with abstract methods ALLOWED_PROPERTIES', 
    marks=pytest.mark.xfail(
      sys.version_info >= (3, 9),
      reason='python versions before 3.9 have a pluralisation bug in the ABC exception message')
  ),
])
def test_abc_stuff(expected_exception_msg):
    ...
    assert exc.msg == expected_exception_msg
```

quis:
@leohemsted it’s very verbose but it’s at least precise and explains what’s going on in code, rather than in comments. Have changed and squashed back into one commit.

quis:
image

tests/test_template_types.py
```diff
@@ -31,24 +32,60 @@
...
-    (BaseEmailTemplate, (
-        "Can't instantiate abstract class BaseEmailTemplate with abstract methods __str__"
-    )),
-    (BaseLetterTemplate, (
-        "Can't instantiate abstract class BaseLetterTemplate with abstract methods __str__"
-    )),
-    (BaseBroadcastTemplate, (
-        "Can't instantiate abstract class BaseBroadcastTemplate with abstract methods __str__"
-    )),
+@pytest.mark.parametrize('template_class, possible_expected_errors', (
```
benthorner, line 35: We can revert this back to `expected_error` as it's no longer a list.

benthorner (approved):
Thanks @quis ⚡.

I agree it is a bit on the verbose side but I don't feel strongly about it. It's been interesting to see a general approach of testing across Python versions.

PR #898, opened 2021-09-15
hard-code the package list to just contain the src dir

leohemsted:
it's unlikely we'll add any separate packages, and this prevents us accidentally bundling the tests folder in.

this was previously resulting in a tests folder being found in the site_packages in my virtualenv, containing all the unit tests from utils in every venv that required the utils repo


setup.py
```diff
@@ -20,7 +20,7 @@
     author='Government Digital Service',
     description='Shared python code for GOV.UK Notify.',
     long_description=__doc__,
-    packages=find_packages(),
+    packages=['notifications_utils'],
```
klssmith, line 23: From trying this out, this change would also mean that the list no longer includes sub-packages such as `notifications_utils.clients` - that didn't seem to break anything for me though so I'm assuming that's fine.

The `find_packages` import at the top of this file can be removed. Not sure why `flake8` isn't picking that up - I also can't get `flake8` to run locally on this repo but haven't looked into why

leohemsted: good point! i did some digging and tweaked it, turns out you can pass include/exclude args to find_packages which i think is better or atl east closer to what setuptools people imagined you should do

klssmith: I like the idea of trying to keep things closer to the "standard way" of doing things. After the change you made, the sub-packages still don't seem to be included, but I'm not sure if that was intentional. If you did want those then it seems like we'd have to do either `find_packages(include=['notifications_utils*'])` or `find_packages(exclude=['tests'])`

klssmith (approved):

PR #902, opened 2021-10-01
Remove unused Zendesk code

klssmith:
This removes the code related to the old way of creating Zendesk tickets from the ZendeskClient. This code has been replaced and is not used anymore.

Also fixes flake8, which had only been running two types of checks.

[Pivotal story](https://www.pivotaltracker.com/story/show/179005041)


notifications_utils/version.py
```diff
@@ -1 +1 @@
-__version__ = '46.1.0'
+__version__ = '47.0.0'
```
klssmith, line 1: Although this is a breaking change to the ZendeskClient, nothing is using the code that was deleted.

setup.cfg
```diff
@@ -2,9 +2,18 @@
 xfail_strict=true
 
 [flake8]
+# B901, B902 are opinionated warnings from flake8-bugbear, so must be added to
+# the `select` in order to work. Since `select` acts as an allowlist, every
+# other check has to be specified here too.
+select=C, E, F, W, B901, B902
```
idavidmcdonald, line 8: Would be good to link to where you can see all the potential values for this and what they are so it's easy to tell what this is choosing

Also does this mean we are also making this mistake of not including everything we need in `select` in any other repos?

Would be nice to have all our repos consistent too with a flake8 file vs a setup.cfg file but that's really not a priority.

idavidmcdonald (approved):
Left a minor comment about flake8 but all looks pretty good in this repo.

setup.cfg
klssmith: I've added a link in a comment to the error codes.

I've checked the other repos and we don't use `select` there, so none of the others need changing.

We're currently inconsistent with where we keep the flake8 config. I went for `setup.cfg` in this case because we're already using it for other things, so we need the file anyway.

PR #911, opened 2021-11-01
Small tidy up - add changelog and remove unused variable

idavidmcdonald:
See commit messages for further details


quis (approved):
Think we should probably reserve the changelog for major versions only to minimise busywork. If you agree it might be worth adding a note to that effect as the first line.

idavidmcdonald:
Very good point Chris, have made that change

PR #917, opened 2021-11-15
DRY-up Celery boilerplate across our repos

benthorner:
https://www.pivotaltracker.com/story/show/180213914

This centralises the copy of this code we have across 5 of our
repos, so it's all in one place. See the commit message for
more detail about how this will be used in each repo.

Since many repos use this one, I don't want to introduce celery
as a runtime dependency. The test dependencies use the lowest
version of Celery among our apps - antivirus is on v4.4. If new
versions of celery have breaking changes, we could:

- Upgrade all our apps to be compatible.

- Regress back to having copies of this code in each repo, or
celery version-specific conditionals in this one.

Tested with:

- `notifications-api` ✅ 
- `notifications-ftp` ✅


CrystalPea (approved):
LGTM 👍🏼 . Nice one making our code more DRY 👏🏼

PR #920, opened 2021-11-17
Don’t try to upgrade Celery to a newer version

quis:
We need to keep the Celery version at the lowest version of used among any of our apps. Antivirus is on v4.4. We need to upgrade the apps before we can upgrade the common version in this repo, and we shouldn’t block other PRs to this repo in the meantime.


benthorner (approved):
Nice catch, thank you!

PR #922, opened 2021-11-18
Make broadcast message preview icon inline SVG

tombye:
This makes the icon into an inline SVG instead of a background image, to make it work in forced colour modes like Windows high contrast mode.

It also:
- fills the SVG with `currentColor` so it matches the text colour of the parent `<h2>`
- uses `aria-hidden=true` to hide it from assistive tech, like the original

These changes will require changes to the CSS in apps that use the template to:
- match the icon's position with the original
- force the SVG to adapt its colours in forced colour modes

## Before

image

image

## After

inline_svg_default_edge

inline_svg_high_contrast_black_edge

inline_svg_high_contrast_white_edge

## Info' on forced colour modes

- [Microsoft article on forced colour modes and Windows high contrast mode](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/)
- [Presentation on high contrast modes](https://docs.google.com/presentation/d/1wxufJr0HmkThd91Ddhr-jhHWUi3wG6CU5RGSfDBKyFQ/edit?usp=sharing)


benthorner (approved):
Nice, works for me when testing locally in Admin using a couple of forced colour schemes ✅. I haven't checked it with a screen reader, but I expect it's all fine.

As a side note, I remembered govuk-alerts has its own fancy triangle, which seems to take [a different approach](https://github.com/alphagov/notifications-govuk-alerts/blob/23bc81536cc177e19704106a1825b30dcf2a4698/app/templates/components/alerts_icon.html#L14): it's always red with a white exclamation mark. I guess the approach here wouldn't work, as we deliberately want it to be a custom colour that's distinct from the rest of the page - hopefully no one needs to have a red canvas!

tombye:
>As a side note, I remembered govuk-alerts has its own fancy triangle, which seems to take a different approach: it's always red with a white exclamation mark. I guess the approach here wouldn't work, as we deliberately want it to be a custom colour that's distinct from the rest of the page - hopefully no one needs to have a red canvas!

I'd guess that it being the same colour as the text here is us reducing the amount of visual information, compared to the gov.uk pages. @quis would have a proper answer.

PR #923, opened 2021-11-23
Let users of request cache specify a custom TTL

quis:
We have a case where data in our cache is becoming stale. Normally we would update the cached value as soon as it becomes stale but this is not possible because the data is populated asynchronously.

Instead we want to reduce the amount of time that the stale data is sitting around for.

At the moment RequestCache keeps data for 7 days (in line with our default data retention policy) and this is hard coded.

This pull request adds a keyword argument to RequestCache.set, allowing any of its user to specify their own cache expiry time.

Units in seconds for consistency with the underlying library.

Note that Redis requires integer values for its TTL argument.


tests/test_request_cache.py
```diff
@@ -57,6 +57,39 @@ def foo(a, b, c, x=None, y=None, z=None):
     )
 
 
+@pytest.mark.parametrize('cache_set_call, expected_redis_call', (
```
idavidmcdonald, line 60: Optional: I feel the param variable names might be a bit more readable as something like `requested_cache_ttl, expected_redis_cache_ttl`. I guess I was expecting to see the word `ttl` in the variable names

tests/test_request_cache.py
```diff
@@ -57,6 +57,39 @@ def foo(a, b, c, x=None, y=None, z=None):
...
+    (0, 0),
+    (1, 1),
+    (1.111, 1),
```
idavidmcdonald, line 63: Optional, could also add in a string case to this like `'2000'` becomes `2000`

idavidmcdonald (approved):

tests/test_request_cache.py
quis:
```suggestion
    (1.111, 1),
    ('2000', 2_000),    
```

PR #925, opened 2021-11-26
Fix docs for international SMS to Lesotho

benthorner:
See email chain "International SMS rules" for more details.

I've also added a header to clarify that the original "attributes"
shouldn't be modified, as discussed with the team.


quis (approved):
> Lesotho- only alpha (no numeric’s) 3-11 characters, no spaces.

> You will receive delivery notifications yes

> would need a registration to go over direct connection [to use our own sender]

Looks like it all matches up to me 👍🏻

PR #927, opened 2021-11-29
Add `make` commands for creating new versions

quis:
This pull request attempts to fix two pain points:

1. Releasing a new version requires manually editing a file, making sure you’ve not made a typo and remembering what the different parts of the version number mean
2. When two people work on two branches and both bump the version, their changes will merge cleanly but the second person to merge won’t get a new version released because their changes no longer include a bump to the version file.

This commit adds make commands which automate and standardise the process of incrementing the version number, making 1. obsolete and 2. much less likely to happen.

So now if you want to increment the version number you can do one of:

- `make version-major` for breaking changes
- `make version-minor` for new features
- `make version-patch` for bug fixes


CrystalPea (approved):
Nice one 👏🏼 . I like @benthorner 's idea to prompt people to add a line or two to changelog, too.

quis:
# Add some helpful print statements 

We want to encourage people to update the changelog when releasing a new version.

This commit not only adds a prompt telling them to do so, it makes the process easier by giving them a template to copy.

image

image

image

scripts/bump_version.py
```diff
@@ -0,0 +1,91 @@
...
+from notifications_utils.version import __version__
+
+version_parts = ('major', 'minor', 'patch')
+
+parser = argparse.ArgumentParser()
+parser.add_argument('version_part', choices=version_parts)
+version_part = parser.parse_args().version_part
+
+new_major, new_minor, new_patch = (
+    int(current) + (part == version_part)
```
benthorner, line 16: Cunning use of a `bool` here 👏.

notifications_utils/version.py
```diff
@@ -1 +1,8 @@
-__version__ = '50.0.0'
+# This file is autogenerated.
```
benthorner, line 1: Nice idea putting this in here to make it self-sustaining.

scripts/bump_version.py
```diff
@@ -0,0 +1,91 @@
...
+    YELLOW = '\033[93m'
+    RED = '\033[91m'
+    BOLD = '\033[1m'
+    UNDERLINE = '\033[4m'
+    END = '\033[0m'
+
+
+print('')
+print(
+    f'{color.BOLD}{color.GREEN}'
```
benthorner, line 62: Looks like one of the packages in this repo pulls in [colorama](https://pypi.org/project/colorama/), so you could use that instead. Equally I appreciate it's good to make the script run standalone.

```
from colorama import Style, Fore

...

f'{Style.BRIGHT}{Fore.GREEN}'
```

benthorner (approved):
Nice one!

I haven't experienced this problem myself, but having a bit more structure / process to how we version this repo can only be a good thing - for such a lightweight script.

PR #930, opened 2021-12-03
Make processing empty cells faster

quis:
# Make strip_and_remove_obscure_whitespace more efficient 

Doing `str.replace` is relatively expensive, and it has no effect when performed on an empty string. So we can make this function a bit more efficient by short circuiting when given an empty input.

### Before

`37.0s`

### After

`15.8s`

# Don’t repeatedly store values against unnamed columns from spreadsheets

If a spreadsheet has many empty columns we dutifully store those empty values against an empty column header, only to never look at them again. This takes extra time because the process of storing them is a bit more involved that just a `dict.update`.

By skipping storing these values we can make processing a spreadsheet a bit more efficient.

### Before

`15.8s`

### After

`5.3s`

# Overall performance improvement

7&times; faster

### DIFF
--- a/notifications_utils/formatters.py
+++ b/notifications_utils/formatters.py
@@ -338,6 +338,11 @@ def strip_all_whitespace(value, extra_characters=''):
 
 
 def strip_and_remove_obscure_whitespace(value):
+    if value == '':
+        # Return early to avoid making multiple, slow calls to
+        # str.replace on an empty string
+        return ''
+
     for character in OBSCURE_ZERO_WIDTH_WHITESPACE + OBSCURE_FULL_WIDTH_WHITESPACE:
         value = value.replace(character, '')
 
--- a/tests/test_recipient_csv.py
+++ b/tests/test_recipient_csv.py
@@ -448,6 +448,20 @@ def test_overly_big_list_stops_processing_rows_beyond_max(mocker):
     ) == 10
 
 
+def test_empty_column_names():
+    recipient_csv = RecipientCSV(
+        """
+            phone_number,,,name
+            07900900123,foo,bar,baz
+        """,
+        template=_sample_template('sms'),
+    )
+
+    assert recipient_csv[0]['phone_number'].data == '07900900123'
+    assert recipient_csv[0][''].data == ['foo', 'bar']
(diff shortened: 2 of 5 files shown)


tests/test_recipient_csv.py
```diff
@@ -430,6 +430,22 @@ def test_overly_big_list():
     assert list(big_csv.rows_with_message_too_long) == []
 
 
+def test_big_empty_list():
+    process = Mock()
+
+    lots_of_commas = ',' * 100_000
```
benthorner, line 436: Can we take a similar approach to 5a730be here? This test takes about 6s to run, which I'd like to avoid if possible.

notifications_utils/formatters.py
```diff
@@ -338,6 +338,9 @@ def strip_all_whitespace(value, extra_characters=''):
 
 
 def strip_and_remove_obscure_whitespace(value):
+    if value == '':
```
benthorner, line 341: I did wonder about whether using `re.sub` might be faster here, instead of doing a separate `replace` for each character. It's no match for an equality check, and it turns out it's actually worse than simple iteration:

- iterated `.replace`: 17.9s
- `re.sub`: 18.5s

benthorner (commented):
Another great speedup! I make it 2.6x, by the way 😄.

We should keep the tests speedy here, as in your other PR.

Optional: you could add a comment to each conditional to explain why they're necessary. I don't think we strictly need tests, but they do help document these things. Since we're lacking a test for the dict update, there's a stronger argument for a comment there.

tests/test_recipient_csv.py
quis: I‘ve changed the test in 2435dca

quis:
I’ve added comments to both conditionals and squashed into the existing commits

tests/test_recipient_csv.py
benthorner: Thanks. You could go smaller still e.g. just 2 commas would be sufficient for the new assertion. However, testing locally indicates the difference is now minimal.

benthorner (approved):
🚀

PR #931, opened 2021-12-03
Skip rows past 100,000 sooner

quis:
Skip processing rows past 100,000 sooner 

In 26443fa I added an efficiency improvement which avoids trying to process the contents of spreadsheet rows once we’ve iterated past the maximum number of rows we allow.

This is good, because it means we don’t waste time trying to validate a million phone numbers, for example.

However it could be even better because we’re still doing some processing of these rows to put them into a data structure, which we then throw away.

So this commit adds the row count check earlier, which means it’s as fast to process a million rows as it is to process 100,000 rows.

***

## Time to run `tests/test_recipient_csv.py::test_overly_big_list`

Row count | Before | After
---|---|---
100,001 | 8.37s | 8.34s
1,000,000 | 12.14s | 8.96s

### DIFF
--- a/notifications_utils/recipients.py
+++ b/notifications_utils/recipients.py
@@ -170,6 +170,10 @@ class RecipientCSV():
 
         for index, row in enumerate(rows_as_lists_of_columns):
 
+            if index >= self.max_rows:
+                yield None
+                continue
+
             output_dict = OrderedDict()
 
             for column_name, column_value in zip(column_headers, row):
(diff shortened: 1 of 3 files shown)


benthorner (commented):
Nice speedup 🚀.

I appreciate this isn't the main performance improvement (https://github.com/alphagov/notifications-utils/pull/930), which should be a straightforward review. For this change I had to think a bit more:

- The new test is really a benchmarking test, which will slow down all subsequent test runs by  around 16 seconds. Although the test suite isn't that slow at the moment, 16 seconds is _a lot_ and since we're not actually checking the benchmark in the test, we should keep it fast. You could keep the test, but we should flag it e.g. `pytest.mark.skip`.

- The `max_rows` behaviour is surprising. Arguably it should raise an if we're passing it an invalid argument, instead of shedding data. I realise that changes the interface, but that's a positive thing: we can make the check a lot faster if we do it even further back - even in Admin. On the other hand, this is just a refactoring, so good as-is.

Happy to approve this once the tests run quick(er) again.

quis:
@benthorner I had an idea about how the test could be fast but also more useful in 5a730be – let me know what you think.

benthorner (approved):
Much better, thanks!

This appears to have reduced the test runtime by ~10s. Would be nice if we could do something similar with:

- `test_big_list` - could just have e.g. 101 rows
- `test_processing_a_big_list` - could maybe delete this test one?

Both optional, though.

quis:
@benthorner I’ll look into those as a separate PR to keep the scope of this one tight

PR #932, opened 2021-12-07
Fix calculation of minor and patch version numbers

quis:
# Before

Previous version | Command | New version | Correct
---|---|---|---
50.0.1 | `make version-patch` | 50.0.2 | ✅
50.0.1 | `make version-minor` | 50.1.1 | ❌
50.1.1 | `make version-major` | 51.1.1 | ❌

# After

Previous version | Command | New version | Correct
---|---|---|---
50.0.1 | `make version-patch` | 50.0.2 | ✅
50.0.1 | `make version-minor` | 50.1.0 | ✅
50.1.1 | `make version-major` | 51.0.0 | ✅

### DIFF
--- a/scripts/bump_version.py
+++ b/scripts/bump_version.py
@@ -12,12 +12,13 @@ parser = argparse.ArgumentParser()
 parser.add_argument('version_part', choices=version_parts)
 version_part = parser.parse_args().version_part
 
-new_major, new_minor, new_patch = (
-    int(current) + (part == version_part)
-    for part, current in zip(
-        version_parts, __version__.split('.')
-    )
-)
+current_major, current_minor, current_patch = map(int, __version__.split('.'))
+
+new_major, new_minor, new_patch = {
+    'major': (current_major + 1, 0, 0),
+    'minor': (current_major, current_minor + 1, 0),
+    'patch': (current_major, current_minor, current_patch + 1),
+}[version_part]
 
 package_contents = subprocess.run(
     ('tar', 'cf', '-', 'notifications_utils'),
--- a/Makefile
+++ b/Makefile
@@ -29,13 +29,13 @@ reset-version:
 	git checkout origin/master -- notifications_utils/version.py
 
 .PHONY: version-major
-version-major: reset-version
+version-major: reset-version ## Update the major version number
 	./scripts/bump_version.py major
 
 .PHONY: version-minor
-version-minor: reset-version
+version-minor: reset-version ## Update the minor version number
 	./scripts/bump_version.py minor
 
 .PHONY: version-patch
-version-patch: reset-version
+version-patch: reset-version ## Update the patch version number
 	./scripts/bump_version.py patch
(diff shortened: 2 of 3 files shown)


scripts/bump_version.py
```diff
@@ -13,7 +13,8 @@
 version_part = parser.parse_args().version_part
 
 new_major, new_minor, new_patch = (
-    int(current) + (part == version_part)
+    (int(current) + (part == version_part))
+    if part <= version_part else 0
```
idavidmcdonald, line 17: I'm really really struggling to figure this code out about how it works. I've pulled it down locally and put in a bunch of print statements and I'm a bit closer but still not quite there. I'm finding it very complex by being done mostly a one liner. I would suggest a potential rewrite that really focusses on readability. I came with some rough pseudocode as an example


```
if bump == 'patch':
    new_patch = current_patch + 1
    new_minor = current_minor
    new_major = current_major

if bump == 'minor':
    new_patch = 0
    new_minor = current_minor + 1
    new_major = current_major

if bump == 'major':
    new_patch = 0
    new_minor = 0 
    new_major = current_major + 1
```

Thoughts? This isn't necessarily going to be a blocker if you feel strongly that the current approach is OK in terms of complexity

quis: The secret is that `'major'`, `'minor'` and `'patch'` happen to stay order of precedence when you sort them alphabetically 😉 

How about something like this as somewhere in between the two?

```python
new_major, new_minor, new_patch = {
    'major': (current_major + 1, 0, 0),
    'minor': (current_major, current_minor + 1, 0),
    'patch': (current_major, current_minor, current_patch + 1),
}[bump]
```

idavidmcdonald: Love it!

quis: 🥳  I’ve changed it to that and squashed

scripts/bump_version.py
```diff
@@ -12,12 +12,13 @@
 parser.add_argument('version_part', choices=version_parts)
 version_part = parser.parse_args().version_part
 
-new_major, new_minor, new_patch = (
-    int(current) + (part == version_part)
-    for part, current in zip(
-        version_parts, __version__.split('.')
-    )
-)
+current_major, current_minor, current_patch = __version__.split('.')
```
idavidmcdonald, line 15: These need to all be ints. Currently this will fail

idavidmcdonald (changes requested):
Left a comment about a problem with the code.

Also, any chance you'd like to add comments to the makefile next to the `version-x` statements as at the moment if I run `make` all I get is `bootstrap` and `test` which then forces me to go look in the makefile to find out what the commands are.

quis:
Oops, I’ve fixed the `str`/`int` problem now. Also added those comments to the `Makefile`.

idavidmcdonald (approved):
Looks great, thanks!

PR #933, opened 2021-12-07
Speed up tests and code for RecipientCSV

benthorner:
https://www.pivotaltracker.com/story/show/177535141

This PR does two things:

- Introduces timeout functionality to avoid RecipientCSV
getting stuck processing massive CSVs.

- Rewrites / removes some the tests for this class, partly
to avoid the timeout but also just make it faster!

Please review commit by commit.

## Test timing stats

```
# before
time pytest
pytest  49.66s user 2.62s system 97% cpu 53.573 total

# after
time pytest
pytest  28.59s user 2.46s system 96% cpu 32.184 total
```

## Request timings

These are for [previewing an uploaded CSV in the Admin app](https://github.com/alphagov/notifications-admin/blob/925f86aa705583e920ea4f0300c3f85fd8546c90/app/main/views/send.py#L645).

Note: these might be a bit dodgy because there's no way to filter for the "/check/" URLs in Kibana: you can't search / filter for non-alphanumeric characters in strings. The query in the screenshot seems to filter out everything we don't care about. I might try and get the same data out of Athena, but I doubt it will change the default.

Screenshot 2021-12-07 at 16 41 36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 51.2.0
+
+* Timeout processing CSVs to avoid timing out downstream requests
+
+This only affects the Admin app, which should ideally rescue the exception,
+but just letting it propagate is also acceptable as it's similar to the
+current behaviour where we timeout in CloudFront.
+
 ## 51.1.0
 
 * Make processing of spreadsheets with many empty columns more efficient
--- a/notifications_utils/recipients.py
+++ b/notifications_utils/recipients.py
@@ -6,6 +6,7 @@ from contextlib import suppress
 from functools import lru_cache, partial
 from io import StringIO
 from itertools import islice
+from threading import Thread
 
 import phonenumbers
 from flask import current_app
@@ -58,6 +59,7 @@ class RecipientCSV():
         remaining_messages=sys.maxsize,
         allow_international_sms=False,
         allow_international_letters=False,
+        processing_timeout=20,
     ):
         self.file_data = strip_all_whitespace(file_data, extra_characters=',')
         self.max_errors_shown = max_errors_shown
@@ -68,6 +70,7 @@ class RecipientCSV():
         self.allow_international_letters = allow_international_letters
         self.remaining_messages = remaining_messages
         self.rows_as_list = None
+        self.processing_timeout = processing_timeout
 
     def __len__(self):
         if not hasattr(self, '_len'):
@@ -147,8 +150,22 @@ class RecipientCSV():
 
     @property
     def rows(self):
-        if self.rows_as_list is None:
+        return self.try_get_rows()
+
+    def try_get_rows(self):
+        if self.rows_as_list is not None:
+            return self.rows_as_list
+
+        def cache_rows():
             self.rows_as_list = list(self.get_rows())
+
+        thread = Thread(target=cache_rows, daemon=True)
+        thread.start()
+        thread.join(timeout=self.processing_timeout)
+
+        if thread.is_alive():
+            raise CSVTooBigError()
+
         return self.rows_as_list
 
     @property
@@ -169,7 +186,6 @@ class RecipientCSV():
         next(rows_as_lists_of_columns, None)  # skip the header row
 
         for index, row in enumerate(rows_as_lists_of_columns):
-
             if index >= self.max_rows:
                 yield None
                 continue
@@ -359,6 +375,11 @@ class InvalidPhoneError(InvalidEmailError):
 
 class InvalidAddressError(InvalidEmailError):
     pass
+
+
+class CSVTooBigError(Exception):
+    def __init__(self, message=None):
+        super().__init__(message or 'CSV is too big to process')
 
 
 def normalise_phone_number(number):
--- a/tests/test_recipient_csv.py
+++ b/tests/test_recipient_csv.py
@@ -808,15 +808,20 @@ def test_international_recipients(file_contents, rows_with_bad_recipients):
 
 def test_errors_when_too_many_rows():
     recipients = RecipientCSV(
-        "email address\n" + ("a@b.com\n" * (RecipientCSV.max_rows + 1)),
+        "email address\n" + ("a@b.com\n" * 101),
         template=_sample_template('email'),
     )
-    assert RecipientCSV.max_rows == 100_000
+
+    # Confirm the normal max_row limit
+    assert recipients.max_rows == 100_000
+    # Override to make this test faster
+    recipients.max_rows = 100
+
     assert recipients.too_many_rows is True
     assert recipients.has_errors is True
-    assert recipients.rows[99_000]['email_address'].data == 'a@b.com'
+    assert recipients.rows[99]['email_address'].data == 'a@b.com'
     # We stop processing subsequent rows
-    assert recipients.rows[100_000] is None
+    assert recipients.rows[100] is None
 
 
 @pytest.mark.parametrize(
--- a/tests/test_recipient_csv.py
+++ b/tests/test_recipient_csv.py
@@ -391,20 +391,6 @@ def test_check_if_message_too_long_for_sms_but_not_email_in_CSV(
         is_message_too_long.called
 
 
-def test_big_list():
-    big_csv = RecipientCSV(
-        "email address,name\n" + ("a@b.com\n" * RecipientCSV.max_rows),
-        template=_sample_template('email', 'hello ((name))'),
-        max_errors_shown=100,
-        max_initial_rows_shown=3,
-        guestlist=["a@b.com"]
-    )
-    assert len(list(big_csv.initial_rows)) == 3
-    assert len(list(big_csv.initial_rows_with_errors)) == 100
-    assert len(list(big_csv.rows)) == RecipientCSV.max_rows
-    assert big_csv.has_errors
-
-
 def test_overly_big_list_stops_processing_rows_beyond_max(mocker):
     mock_strip_and_remove_obscure_whitespace = mocker.patch(
         'notifications_utils.recipients.strip_and_remove_obscure_whitespace'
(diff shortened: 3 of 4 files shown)


benthorner:
Data from Athena for the past month shows the same trend, with only a handful of requests taking over 10s:

Screenshot 2021-12-08 at 10 24 46

(X-axis: request time in seconds, Y-axis: count)

notifications_utils/recipients.py
```diff
@@ -58,6 +59,7 @@ def __init__(
         remaining_messages=sys.maxsize,
         allow_international_sms=False,
         allow_international_letters=False,
+        processing_timeout=10,
```
idavidmcdonald, line 62: Could you explain a bit more about the choice of processing_timeout as 10 seconds please? I can see from your graphs that almost all are uploaded within 10 seconds but equally I can see that some take longer than 10 seconds but less than 20. What is the argument for not upping this to 20 or some other number higher than 10? I guess we don't want it at 30 seconds because that is the http request timeout and so we probably need to put it somewhere a little under 30 seconds to account for any other activity in the http request that isn't spend running `recipients.rows`.

I acknowledge your commit message about how long do we think it is acceptable for a user to wait though so I am open to keeping it at 10. It might be good to get a second opinion on the default timeout just to make sure.

notifications_utils/recipients.py
```diff
@@ -147,8 +150,22 @@ def allowed_to_send_to(self):
...
+
+        def cache_rows():
             self.rows_as_list = list(self.get_rows())
+
+        thread = Thread(target=cache_rows, daemon=True)
+        thread.start()
+        thread.join(timeout=self.processing_timeout)
+
+        if thread.is_alive():
+            raise RuntimeError("CSV is too big to process.")
```
idavidmcdonald, line 167: I'm guessing this won't be the content shown to the user and instead we will need to put into the admin app some handling of this and showing a relevant error message written by Karl?

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
...
 More minor changes may optionally be recorded here too.
 
+## 51.2.0
+
+* Timeout processing CSVs to avoid timing out downstream requests
+
+This only affects the Admin app, which should ideally rescue the exception,
+but just letting it propagate is also acceptable as it's similar to the
+current behaviour where we timeout in CloudFront.
+
```
idavidmcdonald, line 13: Thanks, useful addition

idavidmcdonald (commented):
First three commits are good.

A few questions about the last commit which may or may not require changes.

notifications_utils/recipients.py
quis: If the intention is to catch this exception in another application I think it’s good practice to make a custom exception so you don’t accidentally catch something else which might be raising a `RuntimeError`. For example an invalid phone number doesn’t raise `ValueError`, it raises `InvalidPhoneError`:
https://github.com/alphagov/notifications-utils/blob/72a5f62d474823e1eda0c0ef47825f2d9deabf2f/notifications_utils/recipients.py#L447-L448

`InvalidPhoneError` is defined as:
https://github.com/alphagov/notifications-utils/blob/72a5f62d474823e1eda0c0ef47825f2d9deabf2f/notifications_utils/recipients.py#L356-L357

tests/test_recipient_csv.py
```diff
@@ -422,6 +422,21 @@ def test_overly_big_list_stops_processing_rows_beyond_max(mocker):
     ) == 10
 
 
+def test_rows_processing_timeout():
+    recipients = RecipientCSV(
+        # Keep the number of rows small to avoid DoSing other tests
+        'phone_number,07900900900\n' * 1000,
+        template=_sample_template('sms'),
+        # Use "0" to avoid flakey behaviour based on the no. of rows
+        processing_timeout=0,
```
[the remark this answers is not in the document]
benthorner: I did think of 0.1s, but this raises the ugly question of "how big do I need to go to ensure it takes >0.1s?" and the secondary question of "how much do I want to slow down the test suite because of that?". 

I think we are checking the two most important things here:

- The code runs normally most of the time (with the default timeout).
- The code errors if the timeout is low (0 being the extreme case).

Any other tests would be veering into checking the Thread class works as expected or benchmarking the code, which should be done elsewhere to avoid flakiness or slowing down the test suite.

Can you think of something I'm missing here?

tests/test_recipient_csv.py
```diff
@@ -422,6 +422,21 @@ def test_overly_big_list_stops_processing_rows_beyond_max(mocker):
...
+    )
+
+    with pytest.raises(RuntimeError) as excinfo:
+        assert not recipients.rows
```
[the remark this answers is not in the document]
benthorner: Good point. In theory it's fine because `recipients.rows` should never run to completion due to the timeout, so the `assert not` would correctly fail if it did. However, you're right that this is covered by `pytest.raises`.

I've removed it.

notifications_utils/recipients.py
benthorner: @idavidmcdonald yes, that's the idea. However, it's also not a disaster if we don't have time, since the behaviour is (largely) unchanged from the 504 users get at the moment.

@quis good point, I've introduced a new exception class for this.

benthorner: How does it look now?

quis: Looks good to me 👍🏻

notifications_utils/recipients.py
benthorner: My reasoning is just that: a website shouldn't take 30s to do something, or anything close to that.

Backing away from that extreme, slightly, I get that we want to give people a happy journey as much as possible, so let's look at the numbers for that graph. Worth noting some of these will be because of me testing.

- 0-10s: 99.86%
- 10-15s: 0.02%
- 15-20s: 0.01%
- 20-25s: 0.00%

Now we could argue: is there any harm in supporting an extra 0.03% of requests? Maybe not, but equally it's likely these spreadsheets / CSVs have something wrong with them, so showing an error a little sooner may actually be more helpful than making the user wait to see they can't have e.g. 10 million rows.

I am open to changing it. Given the tiny number of requests it would affect (some of which were probably mine), I'm also open to just trying it and seeing if it actually causes a problem for anyone.

@quis do you want to add an opinion here?

quis: Hmm, I guess it’s slightly arbitrary, in other words it’s hard to say why it should be 10 seconds (99.86%) or 15 seconds (99.88%).

My concern is that, let’s say 99.8% of sheets have a lot less than 100,000 rows. So hypothetically the difference between 10 and 15 seconds could be the difference between being able to process most 100,000 files and rejecting most of them. It wouldn’t be great if this change meant that we were unable to process most of the files that we say are within our row limit.

Could we do a quick look at how long it takes to process a file of 100,000 phone numbers (for example) on production, just to check we’re not getting close to timing out stuff we should be able to process?

I agree that having to wait 30 seconds is bad, but making people split their file in two so that we can process it is also an inconvenience.

tests/test_recipient_csv.py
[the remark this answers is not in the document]
idavidmcdonald: Good point about the other tests will make sure it works with the default timeout. That was something I'd missed.

Happy to stick with this! Thanks for explaining.

notifications_utils/recipients.py
benthorner: Another good point! I've done some tests in Production with 100K rows:

- Email CSV (1 column): average of 6.5s
- Email CSV (4 columns): average of 10s
- Phone CSV (4 columns): average of 13s
- Phone CSV (4 columns): average of 15s (with 32K errors for invalid numbers)
- Phone CSV (8 columns): average of 20s (with 32K errors)

I'm not sure how many columns we think is reasonable for a template. 4 seems reasonable to me, so this pushes us towards 20s, which means we may or may not timeout for the last scenario.

How do we feel about 20s?

quis: 20 seconds feels good based on that.

It also matches what we tell people in tickets at the moment, eg ‘if your file has a lot of columns you may need to split it into multiple files’

benthorner: Cool, I've bumped it to 20s. The Admin app can also easily override it if it needs to, but it's handy to have a sensible default so we don't have to.

@idavidmcdonald that work for you?

quis: One more non-blocking comment: The API app will need to override this value. Otherwise you could have a situation where a CSV which the admin app happily processes in 19 seconds takes 21 seconds to process in the API app, causing the task to fail.

benthorner: I had a look into this when I was investigating this class. As far as I can tell, the `.rows` property isn't used in API, which [calls `.get_rows()` directly](https://github.com/alphagov/notifications-api/blob/19739945163793db79c89d89da39fc01c3c31b68/app/celery/tasks.py#L93). I'm not sure why it's written that way, but it does mean the change here should have no impact there.

I do realise that `.rows` is used by various other properties / methods in the class, but again I can't see anywhere in API where they're used - the class just isn't used that much in that repo.

Also agree it's non-blocking - I don't think we need to change the default in there though. I guess this does highlight that `processing_timeout` is a bit ambiguous, but I think that's more an issue with the class' implementation e.g. it _could_ be more like a factory that does all the processing upfront.

Short summary: I agree it's non-blocking, and don't think it's an issue.

idavidmcdonald: All good

idavidmcdonald (approved):
I'll approve this as my only ask is trivial.

Please could you updated the commit message that brings in the 20 second process_timeout for two reasons

1. It refers to 10 seconds, rather than 20
2. It would be great to include the reason for our choice of 20 either in the code, or easier in the commit message. People often look at git blame to see why something was a certain way and the conversation we've had in the PR is really valuable and would be good to have a summary for why the choice of 20 in the commit message to make it easier for those who use git blame to find why.

benthorner:
Good spot, thanks @idavidmcdonald. Updated ✅.

PR #936, opened 2021-12-10
Revert "Timeout processing RecipientCSV rows"

benthorner:
https://www.pivotaltracker.com/story/show/177535141

This reverts commit 4fdf28412c4a74824f06607a19e9893b9e730326.

It turns out showing a contextual message for this error is hard,
as we would need to either:

- Show the error on an existing preview page, but be very careful
to structure the code so that none of the other checks run, since
these would trigger more timeout errors.

- Introduce a new error page. This is unprecedented for previewing
a CSV, where we always try to show a preview - this isn't possible
if the CSV is taking too long to process.

Taking a step back, we shouldn't really be adding timeout wrappers
around our code. A more future-proof approach would be:

- To monitor timings for all requests to our apps.
- Take action to improve performance where necessary.

While it's still possible to get this code to run for a long time,
the improvements we've already made should reduce the likelihood
of this being an issue in practice.

### DIFF
--- a/notifications_utils/version.py
+++ b/notifications_utils/version.py
@@ -5,4 +5,4 @@
 # - `make version-minor` for new features
 # - `make version-patch` for bug fixes
 
-__version__ = '51.2.0'  # 1e1f0a68e40b7c2b164567e4d7e1d373
+__version__ = '51.2.1'  # 15d2163e4f04672521d384b17b1fecb2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 51.2.1
+
+* Revert 51.2.0.
+
 ## 51.2.0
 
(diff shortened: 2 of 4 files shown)


notifications_utils/version.py
```diff
@@ -5,4 +5,4 @@
 # - `make version-minor` for new features
 # - `make version-patch` for bug fixes
 
-__version__ = '51.2.0'  # 1e1f0a68e40b7c2b164567e4d7e1d373
+__version__ = '51.1.1'  # 5fd0bbc81c3653cee9fa2ada1b78c8c5
```
idavidmcdonald, line 8: Do we actually need to release this as something above 51.2.0. I don't know if our pipeline will have an issue with trying to republish lower than the current version?

idavidmcdonald (approved):

notifications_utils/version.py
benthorner: Oops, thanks for spotting. You're right, we shouldn't do this: there's already a tag for 51.1.1. 

I've amended the commit to a higher, patch version.

PR #937, opened 2021-12-14
Purge unused redis functions

leohemsted:
We don't use them, we haven't for a while since we simplified our redis usage.

* Deprecate the following unused `redis_client` functions:
  - `redis_client.increment_hash_value`
  - `redis_client.decrement_hash_value`
  - `redis_client.get_all_from_hash`
  - `redis_client.set_hash_and_expire`
  - `redis_client.expire`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 52.0.0
+
+* Deprecate the following unused `redis_client` functions:
+  - `redis_client.increment_hash_value`
+  - `redis_client.decrement_hash_value`
+  - `redis_client.get_all_from_hash`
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 52.0.0
+
+* Deprecate the following unused `redis_client` functions:
+  - `redis_client.incr`
```
klssmith, line 9: Looks like `incr` is being used [in api](https://github.com/alphagov/notifications-api/blob/81a79e56ced891ad58901ea61ce707e907de3363/app/notifications/process_notifications.py#L160)

leohemsted: ah good spot - will change some search terms and look again

klssmith (approved):

PR #938, opened 2021-12-20
Log stacktrace when a task fails

benthorner:
https://www.pivotaltracker.com/story/show/180693991

This should help debug the failures we're seeing for one of the
nightly tasks - create-nightly-notification-status-for-day - even
though the task gets to the point of logging a success.

It's worth noting Celery is probably emitting useful logs, but we
supress them [1]. It looks like we used to log an exception [2], 
but somehow this changed to 'error' in [3].

Logging an exception means we automatically get a stacktrace [4].

The call to "super()" is unnecessary:

- We don't do it for "on_success".
- It wasn't tested.
- The overridden method does nothing [5].
- None of the docs say to call "super()".

[1]: https://github.com/alphagov/notifications-api/blob/master/scripts/paas_app_wrapper.sh#L9
[2]: https://github.com/alphagov/notifications-api/commit/2622866622cb331f79e208dcee389bca480012f1#diff-76936416943346b5f691dac57a64acebc6a1227293820d1d9af4791087c9fb9eR10
[3]: https://github.com/alphagov/notifications-utils/commit/8765f84a613e5bc0036daf2f036cc950308810f3#diff-9f2919d5016ba611c744a546be5484fc79b037248a6a692dc14befe2534912f8R57
[4]: https://docs.python.org/3/library/logging.html#logging.Logger.debug
[5]: https://github.com/celery/celery/blob/5b698151d5e8da10f6706df42fb99fb3105ac025/celery/app/task.py#L1024

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 51.3.0
+
+* Log exception and stacktrace when Celery tasks fail.
+
 ## 51.2.1
 
 * Revert 51.2.0.
--- a/notifications_utils/celery.py
+++ b/notifications_utils/celery.py
@@ -52,7 +52,7 @@ def make_task(app):
         def on_failure(self, exc, task_id, args, kwargs, einfo):
             # enables request id tracing for these logs
             with self.app_context():
-                app.logger.error(
+                app.logger.exception(
                     "Celery task {task_name} (queue: {queue_name}) failed".format(
                         task_name=self.name,
                         queue_name=self.queue_name,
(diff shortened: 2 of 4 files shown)


klssmith:
> It looks like the stacktrace info was just an omission from when we added our custom log originally

It was originally using `logger.exception` (instead of `logger.error`) so the stacktrace would have been output automatically. We could just switch back to `logger.exception`, to save passing the argument in and to keep it (very slightly) simpler?

benthorner:
😱 oh no, thanks for spotting @klssmith. I looked through the old classes and they're all doing `exception`, so I've no idea how `error` crept in. Anyway, it seems reasonable to change it back e.g. so we get alerted about the issues I'm trying to debug. I've amended the PR as you suggested.

klssmith (approved):

klssmith:
I've approved and then realised the tests are failing, but I think this is good to go once the tests are fixed

benthorner:
Thanks Katie. I had an argument with Git - way too many amends to modify a few lines. Oh well.

PR #939, opened 2021-12-30
Bump govuk-bank-holidays to latest version

benthorner:
While the package can always fetch new holidays via the GOV.UK API,
the latest version of the packages also caches ones for next year,
which means we can avoid unnecessary web requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 51.3.1
+
+* Bump govuk-bank-holidays to cache holidays for next year.
+
 ## 51.3.0
 
 * Log exception and stacktrace when Celery tasks fail.
--- a/notifications_utils/version.py
+++ b/notifications_utils/version.py
@@ -5,4 +5,4 @@
 # - `make version-minor` for new features
 # - `make version-patch` for bug fixes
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 51.3.1
+
+* Bump govuk-bank-holidays to cache holidays for next year.
+
```
idavidmcdonald, line 9: Optional: I personally wouldn't have added this, it feels of little importance and unlikely to cause any issues to any consumers of the utils whilst also setting the precedent of including patch version bumps for others on the team which may also not be needed.

But I really don't mind to be honest, happy to keep it if you prefer it in

idavidmcdonald (approved):
Nice work

CHANGELOG.md
benthorner: I do find them useful: having bumped this package many times on many out-of-date repos, it's annoying to have to go and manually check the commits / PRs to find out what is actually changing.

Example: https://github.com/alphagov/notifications-template-preview/pull/610

Arguably patch versions should never impact anything, but if I'm doing a thorough job of reviewing an upgrade PR then I'd want to double check. The CHANGELOG does have a comment at the top to explain it's optional.

PR #942, opened 2022-01-18
Bump celery requirement to version 5

idavidmcdonald:
No reason for it to be pinned, nothing interesting in the git
history. Brings it up to parralel with our other apps.

Also remove the pyup ignore as again no reason for this to be
ignored.

### DIFF
--- a/requirements_for_test.txt
+++ b/requirements_for_test.txt
@@ -1,6 +1,6 @@
 -r requirements.txt
 
-celery==4.4.0 # pyup: ignore
+celery==5.2.2
 beautifulsoup4==4.10.0
 pytest==7.0.0
 pytest-mock==3.7.0


benthorner:
Are you sure?

[This](https://github.com/alphagov/notifications-utils/commit/1e4b5e1f6fa91ff34cc18b71ee09348b85c87369) is still [true](https://github.com/alphagov/notifications-antivirus/blob/master/requirements.in#L3).

leohemsted:
Antivirus is now on a high enough version of celery.

Bringing up celery to 5.2.2 solves a security warning about celery 5.2.1, which while doesn't affect us as we dont use a results backend, does get in the way when trying to see if we have any outstanding security advisories

https://github.com/advisories/GHSA-q4xr-rc97-m4xx

PR #943, opened 2022-01-21
Rename `Columns` to `InsensitiveDict`

quis:
Originally the `Columns` class was just used for representing the columns in a spreadsheet row.

However it’s now used more generally when we need a dictionary whose keys are not sensitive to case or whitespace. So this pull request renames it to `InsensitiveDict`. 

I’ve gone with the name `InsensitiveDict` because it follows the pattern set by `OrderedDict` from the `collections` standard library.

***

Best review commit-by-commit.


klssmith (approved):
The last commit message could be amended to avoid confusion because it still mentions version 52.0.0, but otherwise looks good

PR #946, opened 2022-02-08
Remove `column` argument from recipient validation functions

quis:
This argument was added<sup>1</sup> because:
- we used to have a function called `validate_address` which needed to know which column of the postal address was being checked (for example the `postcode` column had separate rules)
- we have a function called `validate_recipient` which dynamically  calls `validate_email_address` or `validate_phone_number` (and used to call `validate_address`) based on the recipient type, meaning these  three functions needed to have the same signature

Now that there is no function called `validate_address` the signatures of `validate_phone_number` and `validate_email_address` don’t need to include a `column` argument.

***

1. https://github.com/alphagov/notifications-utils/commit/4dce0ecf085f818c2293d7751804d4789b9d1414#diff-a8a994bf655634f89dc7439880708b4ff0d780ac1bd8033827d8aaa2692a8e0fR423-R428


idavidmcdonald (approved):
Lovely, good to see this code become less complex

PR #947, opened 2022-02-14
add should_validate flag to RecipientCSV

leohemsted:
RecipientCSV isn't just used when uploading spreadsheets - it's also used when rendering spreadsheets that we've already accepted into the system (for example when downloading the job report). We don't need to validate that those jobs are valid - this includes checking the template length, validating phone numbers, making sure all placeholders are represented, etc.

By removing these checks, we can bring down the time it takes to parse and load a real world 77,000 line CSV into memory from 39s down to 12s.

It still takes a little longer

nb: i'm not 100% sold on the name `should_validate` and am open to suggestions.

### Before
image

### After
image


notifications_utils/recipients.py
```diff
@@ -58,6 +58,7 @@ def __init__(
         remaining_messages=sys.maxsize,
         allow_international_sms=False,
         allow_international_letters=False,
+        should_validate=True,
```
idavidmcdonald, line 61: Yep, good use of `True` here :)

notifications_utils/recipients.py
```diff
@@ -362,7 +365,13 @@ def __init__(
         placeholders,
         template,
         allow_international_letters,
+        validate_row=True,
     ):
+        # if we aren't validating the row, then no need to store template (used to check message length) and the cell
+        # level stuff (checks phone nums are valid, checks placeholders are present, etc)
```
idavidmcdonald, line 371: Very optional, could be a little bit clearer, in particular around 'cell level stuff' which might be better as `Cell` to show it is the class you are referring to and removing 'stuff' which is vague

If we don't need to validate, then:
- by setting no template we avoid the template level validation (used to check message length)
- by setting no `error_fn`, we avoid the `Cell` level validation (used to check phone nums are valid, placeholders are present, etc)

Does make me wonder if `validation_fn` would be a better name instead of `error_fn` for `Cell.__init__` but that is probably one for another day.

notifications_utils/recipients.py
idavidmcdonald, line 61: I don't have any particular objections to the naming you chose of `should_validate`, other options off the top of my head are

`validate`
`validate_rows_and_cells`

but I don't mind really. Would be happy to stick with what you've got

tests/test_recipient_csv.py
```diff
@@ -1302,3 +1302,24 @@ def test_email_validation_speed():
     )
     for row in recipients:
         assert not row.has_error
+
+
+@pytest.mark.parametrize('should_validate', [True, False])
```
idavidmcdonald, line 1307: Very optional, you could also add an explicit test case in here for the default, ie when `should_validate` has not been set. Similar to https://github.com/alphagov/notifications-utils/pull/946/files#diff-936d1cd89027c14eeb7a43e17df32fa0234db7b2313af63b9b1a8e69e689bdf3R252

However, I don't mind if we don't as I think the rest of the tests in this file give us enough confidence that validation is happening when we don't include the `should_validate` argument

idavidmcdonald (approved):
Nice one

PR #950, opened 2022-03-01
Add "delete_by_pattern" as RequestCache decorator

benthorner:
https://www.pivotaltracker.com/story/show/180663519

In response to: [^1].

This will replace [^2] with direct calls to:

      @cache.delete_by_pattern

I've included a bunch of small refactorings based on my
experience trying to make this change.

[^1]: https://github.com/alphagov/notifications-admin/pull/4162#pullrequestreview-896208092
[^2]: https://github.com/alphagov/notifications-admin/pull/4162/commits/316b987680f6f56d3896545e8064a21923755c4a


notifications_utils/clients/redis/__init__.py
```diff
@@ -1,8 +1,6 @@
-import json
-from contextlib import suppress
-from datetime import datetime, timedelta
-from functools import wraps
-from inspect import signature
+from datetime import datetime
+
+from .request_cache import RequestCache  # noqa: F401 (unused import)
```
quis, line 3: Nice way of re-organising things without a breaking change

quis (approved):
Looks good to me

notifications_utils/version.py
```diff
@@ -5,4 +5,4 @@
 # - `make version-minor` for new features
 # - `make version-patch` for bug fixes
 
-__version__ = '55.0.0'  # 5568605a0a138ed2231ae2ae2a13beae
+__version__ = '55.1.0'  # aaa1b86c488c4e25c842f0a443f9a26b
```
servingUpAces, line 8: I would have expected this to be a major upgrade since it's a breaking change to admin and api. Shall we make this version 6.0.0?

benthorner: There shouldn't be any breaking changes in here:

- The decorator is new and will only be used in Admin.
- The redis module changes should be backwards compatible.
- The rest of the changes are just to the tests.

The module change is a big one and not something I've had much practice at. Is there something I missed there?

PR #961, opened 2022-04-06
Unpin pyproj

klssmith:
There's nothing in the latest version (3.3.0) that looks dangerous, so
by loosening the requirements here we can keep pyproj up to date in
notifications-admin.


benthorner (approved):
Admin tests still pass with 3.3.0 ✅.

PR #963, opened 2022-04-08
Unpin most dependencies and remove redundant ones

benthorner:
https://www.pivotaltracker.com/story/show/181588331

We decided we prefer to use unpinned dependencies in this repo [^1]
so this makes all of them consistent.

I've also removed two dependencies which were added in response to
PyUp generating to many update PRs [^2]. Since we now have a strategy
I think we should put them back and solve the issue more explicitly
in the apps that need them.

[^1]: https://docs.google.com/document/d/1DO6nFV9E7Qr94_NnUfYsMKFrKTvcjn0ESjjDHodfQUk/edit#heading=h.pohfahyz07ro
[^2]: https://github.com/alphagov/notifications-utils/commit/d0fa02026d525ea537785b91255db182b01e6caf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 55.1.3
+
+* Unpin most dependencies and remove redundant ones (no action required).
+
 ## 55.1.1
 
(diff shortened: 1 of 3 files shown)


leohemsted (approved):

benthorner:
FYI I added an extra commit to bump the version, just to make sure we incorporate this change in other repos and check there aren't any surprises while it's still fresh.

PR #966, opened 2022-04-20
Bump pypdf2 to 1.27.5

leohemsted:
fixes https://github.com/py-pdf/PyPDF2/security/advisories/GHSA-xcjx-m2pj-8g79


quis (approved):
Good to go once it has a version bump

PR #970, opened 2022-05-23
Move markdown-related classes to own module

quis:
`formatters.py` is a big file containing a mixture of:
- short functions which take mostly taker a single string argument and transform it somehow
- big class definitions which inherit from Mistune (the library we use to parse Markdown)

This commit splits the file in two, putting the latter into a new file called `markdown.py`. It feels like the code is easier to navigate this way.

By keeping the functions which take a string and apply the Markdown formatting in `formatters.py` this should be non-disruptive change.


notifications_utils/__init__.py
```diff
@@ -17,3 +17,8 @@
     r'(^|\s)\S+@\S+(\s|$)',
     flags=re.MULTILINE,
 )
+
+# The magic sequence is a ‘unique’ series of characters which we temporarily insert
+# and then later remove when performing tricky formatting operations
+MAGIC_SEQUENCE = "🇬🇧🐦✉️"
```
benthorner, line 23: Could this go in the new markdown module?

While the above code sets a precedent, this repo does lots of different things and yet these two variables are only used for a relatively small subset of them; they don't strike me as particularly "global" and are actually specific to the markdown transformations we perform.

notifications_utils/formatters.py
```diff
@@ -361,239 +299,6 @@ def strip_unsupported_characters(value):
...
-        return ''
-
-    def link(self, link, title, content):
-        return ''.join((
-            content,
-            ' ({})'.format(title) if title else '',
-        ))
-
-
 notify_email_markdown = mistune.Markdown(
```
benthorner, line 302: Should these go in the markdown module as well?

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 56.0.0
+
+* `create_sanitised_html_for_url` must be imported from
+  `notifications_utils.markdown`, not `notifications_utils.formatters`
+* `MAGIC_SEQUENCE` and `magic_sequence_regex` must be imported from
```
benthorner, line 10: Does anything actually use any of these three values?

I can't find any references to them in Admin or API, so this seems like a minor change to me.

benthorner (commented):
This is a positive change for me 👍. What remains is - apart from my comment below - totally separate from the new markdown module. I've found this file hard to navigate before due to the sheer volume of "stuff" mixed in with the already-complex mistune classes; it was hard to know what was relevant. This makes that easier at a glance 🚀.

"formatters" is quite a vague name and I don't think it accurately describes what's left in the file - some of the functions are really just "template utils" e.g. `add_prefix`. I can see many of the functions _are_ used elsewhere, so I don't think it's a major issue, except for those remaining `_markdown` attributes - I'm not sure why they're still there.

Short version: a good change, but a few things I want to check before approving it.

CHANGELOG.md
quis: I checked and yes, we don’t import any of these directly. Changed to a patch version bump instead but kept a CHANGELOG entry.

benthorner (approved):
Thanks for going through my comments - good to go 👍.

PR #980, opened 2022-07-26
Ampersand placeholder fix

BlessedDev:
Presently the use of ampersands "&" in placeholders does not work as expected. This is because Jinja converts it to its special entity "&amp;". 

I have made adjustments to the Placeholder class so that this is done before hand.

### DIFF
--- a/notifications_utils/field.py
+++ b/notifications_utils/field.py
@@ -108,7 +108,7 @@ class Field:
 
     @values.setter
     def values(self, value):
-        self._values = InsensitiveDict(value) if value else {}
+        self._values = InsensitiveDict({self.sanitizer(k): value[k] for k in value}) if value else {}
 
     def format_match(self, match):
         return self.format_placeholder(Placeholder.from_match(match))
--- a/notifications_utils/field.py
+++ b/notifications_utils/field.py
@@ -111,8 +111,9 @@ class Field:
         self._values = InsensitiveDict(value) if value else {}
 
     def format_match(self, match):
-        placeholder = Placeholder.from_match(match)
+        return self.format_placeholder(Placeholder.from_match(match))
 
+    def format_placeholder(self, placeholder):
         if self.redact_missing_personalisation:
             return self.placeholder_tag_redacted
 
@@ -131,7 +132,7 @@ class Field:
         replacement = self.get_replacement(placeholder)
 
         if replacement is None:
-            return self.format_match(match)
+            return self.format_placeholder(placeholder)
 
         if placeholder.is_conditional():
             return placeholder.get_conditional_body(replacement)
--- a/notifications_utils/field.py
+++ b/notifications_utils/field.py
@@ -128,16 +128,15 @@ class Field:
 
     def replace_match(self, match):
         placeholder = Placeholder.from_match(match)
-        replacement = self.values.get(placeholder.name)
+        replacement = self.get_replacement(placeholder)
 
-        if placeholder.is_conditional() and replacement is not None:
+        if replacement is None:
+            return self.format_match(match)
+
+        if placeholder.is_conditional():
             return placeholder.get_conditional_body(replacement)
 
-        replaced_value = self.get_replacement(placeholder)
-        if replaced_value is not None:
-            return self.get_replacement(placeholder)
-
-        return self.format_match(match)
+        return replacement
 
     def get_replacement(self, placeholder):
         replacement = self.values.get(placeholder.name)
--- a/notifications_utils/field.py
+++ b/notifications_utils/field.py
@@ -80,15 +80,15 @@ class Field:
         markdown_lists=False,
         redact_missing_personalisation=False,
     ):
+        self.sanitizer = {
+            'escape': escape_html,
+            'passthrough': str,
+        }[html]
         self.content = content
         self.values = values
         self.markdown_lists = markdown_lists
         if not with_brackets:
             self.placeholder_tag = self.placeholder_tag_no_brackets
-        self.sanitizer = {
-            'escape': escape_html,
-            'passthrough': str,
-        }[html]
         self.redact_missing_personalisation = redact_missing_personalisation
 
     def __str__(self):
--- a/tests/test_field.py
+++ b/tests/test_field.py
@@ -116,6 +116,26 @@ def test_returns_a_string_without_placeholders(content):
             {"warning": False},
             ""
         ),
+        (
+            "Please report to the ((>location)) office at ((&time)) on ((<day)).",
+            {">location": "London", "&time": "09:00", "<day": "Monday"},
+            "Please report to the London office at 09:00 on Monday."
+        ),
+        (
+            "Please report to the ((&gt;location)) office at ((&amp;time)) on ((&lt;day)).",
+            {"&gt;location": "Manchester", "&amp;time": "08:00", "&lt;day": "Thursday"},
+            "Please report to the Manchester office at 08:00 on Thursday."
+        ),
+        (
+            "Dear ((\name)), your passport is now ready for collection from ((/collection_point)).",
+            {"\name": "Jane Doe", "/collection_point": "Point A"},
+            "Dear Jane Doe, your passport is now ready for collection from Point A."
+        ),
+        (
+            "Dear ((/\name)), your passport is now ready for collection from ((//collection_point)).",
+            {"/\name": "John Doe", "//collection_point": "Point B"},
+            "Dear John Doe, your passport is now ready for collection from Point B."
+        ),
     ]
 )
 def test_replacement_of_placeholders(template_content, data, expected):
--- a/tests/test_template_types.py
+++ b/tests/test_template_types.py
@@ -1779,6 +1779,21 @@ def test_templates_extract_placeholders(
     assert template_instance.placeholders == OrderedSet(expected_placeholders)
 
 
+def test_html_template_can_inject_personalisation_with_special_characters():
+    template_content = "This is something text with (( this&that )) HTML special character personalisation <>."
+    personalisation = {"this&that": "some very lovely &"}
+
+    result = str(
+        HTMLEmailTemplate(
+            {'content': template_content, 'subject': '', 'template_type': 'email'},
+            personalisation
+        )
+    )
+    assert (
+        "This is something text with some very lovely &amp; HTML special character personalisation &lt;&gt;." in result
+    )
+
+
 @pytest.mark.parametrize('extra_args', [
     {
         'from_name': 'Example service'
--- a/notifications_utils/version.py
+++ b/notifications_utils/version.py
@@ -5,4 +5,4 @@
 # - `make version-minor` for new features
 # - `make version-patch` for bug fixes
 
-__version__ = '57.0.0'  # 72dbfa4fbbd779a50a862674ee0d0873
+__version__ = '57.0.1'  # 578d19dbd49ee204e0ef3f6828fb713d


quis:
Hey, a few things to look at here.

Main one is that I don’t think this fixes the bug fully. It will make the ampersand render correctly when previewing the template. But it’s still not possible for someone to use a placeholder with an ampersand in the name.

Start with a text message template as a whole, rather than just a single placeholder:
```python
template = SMSPreviewTemplate({
    'content': '((date&time))',
    'template_type': 'sms'
})
```

Then here are a few test cases:

***

```python
assert "<span class='placeholder'>((date&amp;time))</span>" in str(template)
```
The ampersand is escaped correctly when previewing the template as HTML. This `assert` now passes with your change, which is good.

***

```python
assert template.placeholders == {'((date&time))'}
```
This doesn’t pass because `template.placeholders` now contains one item: `date&amp;time`.

***

```python
template.values = {'date&time': '1 January, midnight'}
assert '1 January, midnight' in str(template)
```
This doesn’t pass for the same reason as the previous test: the dictionary key `date&time` doesn’t match the placeholder `date&amp;time`.

***

The other things that are worth considering:
1. `if '&' in body and ';' not in body` feels quite a fragile check (i.e. it would be easy to think of a case like `date&time ;` which allows the unescaped ampersand through). Would it be better to use a library here?
2. Are there other characters which might be escaped by Jinja, for example `<`? Could we have test cases for those?

BlessedDev:
I will look into the test cases you raised.

I also take your point about the possible use of a library and the fact that other characters will trigger the same problem. 

I will look into a more robust solution.

samuelhwilliams:
Hey @BlessedDev - any update on this? Can we get this completed?

BlessedDev:
@samuelhwilliams 
This was put on hold in order to focus on email branding work. I will have a look at it today.

samuelhwilliams:
Ah ok. Any email branding story you're already working on should probably take priority - but perhaps when you finish it, it would be good to come back to this before starting anything else?

BlessedDev:
I have finished the email branding ticket I was working on, so I will have a look at this one.

BlessedDev:
@quis @samuelhwilliams 

I have had another go at this. I have done a bit of adjusting and refactoring to the get_replacement and replace_match methods of the Field class. The functionality now works and the solution is a light touch that would hopefully not break anything else!

notifications_utils/field.py
```diff
@@ -130,19 +131,22 @@ def format_match(self, match):
 
     def replace_match(self, match):
```
samuelhwilliams, line 132: I think it's going to be very valuable to add some decent commenting about what's happening to the values and content as we're going through here.

Specifically for this one, that if we're replacing a match, then the content will have been sanitized, including placeholder names. Which is why we need to sanitize the placeholder names in the values dict.

notifications_utils/field.py
```diff
@@ -130,19 +131,22 @@ def format_match(self, match):
...
 
         return self.format_match(match)
 
     def get_replacement(self, placeholder):
-        replacement = self.values.get(placeholder.name)
+        replacement_values_dict = {
+            self.sanitizer(k): v for k, v in self.values.items()
+        }
+        replacement_values = InsensitiveDict(replacement_values_dict)
+        replacement = replacement_values.get(placeholder.name)
```
samuelhwilliams, line 148: This feels slightly weird and wasteful - we're sanitizing all of the values every time we try to retrieve one of them.

I think we might want to have another attribute similar to `self.values` such as`self.sanitized_values`. We can populate that in the `__init__` method once and then just grab from there as needed. Saves a bit of wasted computation, and also provides another clearer place to add some comments/documentation around this slightly weird workaround.

We definitely want to have some good clear comments describing what's happening here because it's not intuitive or ideal.

tests/test_template_types.py
```diff
@@ -2636,3 +2636,19 @@ def test_broadcast_message_single_counts_diacritics_in_gsm(
...
+@pytest.mark.parametrize('template_class, value_1, value_2', [
+    (EmailPreviewTemplate({'content': "((date&time))", 'subject': 'subject', 'template_type': 'email',
+                          'redact_missing_personalisation': False}), '((date&amp;time))', 'date&time'),
+    (SMSPreviewTemplate({"content": "((>than))", "subject": "((subject))",
+                        "template_type": "sms"}), '((&gt;than))', '>than')
+])
+def test_special_html_characters_used_in_placeholders_are_handled_appropriately(
+        template_class,
+        value_1,
+        value_2
```
samuelhwilliams, line 2650: Can we think of a better name for these? `sanitized_content` and `placeholder_name` maybe?

I think this test would more clearly demonstrate what it's trying to if we also render the replaced text, to show that placeholders go in unescaped but correctly replace the escaped placeholder names.

BlessedDev: @samuelhwilliams I have removed this test. There is a test in `test_field.py`, `test_replacement_of_placeholders`, that is more appropriate. I have updated it to include tests for html special entities.

notifications_utils/field.py
BlessedDev: @samuelhwilliams 
I agree it is not ideal. The challenge of this ticket has been implementing a solution that does not break everything else. I had tried the changes you suggested and about 246 tests failed, the Field class is crucial and changes to it's `__init__` method are consequential.
I have come up with another work around, this time I have given `self.values` the ability to create a `sanitized_content`  dict that can be accessed through it(`self.values`) and this is then used  by the `get_replacement` method. This preserved the current workings of the` Field` class, doesn't break hundreds of tests and solves the bug. I have made comments in the code which I hope are clear.

notifications_utils/formatters.py
```diff
@@ -194,6 +194,13 @@ def escape_html(value):
     return value
 
 
+def unescape_html(value):
```
samuelhwilliams, line 197: I don't think we need this - it isn't used anywhere. Can we drop it now? :)

tests/test_field.py
```diff
@@ -116,6 +116,26 @@ def test_returns_a_string_without_placeholders(content):
...
+        (
+            "Please report to the ((>location)) office at ((&time)) on ((<day)).",
+            {">location": "London", "&time": "09:00", "<day": "Monday"},
+            "Please report to the London office at 09:00 on Monday."
+        ),
+        (
+            "Please report to the ((&gt;location)) office at ((&amp;time)) on ((&lt;day)).",
+            {"&gt;location": "Manchester", "&amp;time": "08:00", "&lt;day": "Thursday"},
+            "Please report to the Manchester office at 08:00 on Thursday."
+        ),
```
samuelhwilliams, line 128: Can we get a test case with these as unescaped values - which I think is actually the case we're more interested in?

eg just `date&time` in both places rather than `date&amp;time`?

notifications_utils/formatters.py
BlessedDev: Yes. I will take it out

tests/test_field.py
samuelhwilliams: 👀 you've pointed out the above test, oops

notifications_utils/field.py
```diff
@@ -91,6 +91,17 @@ def __init__(
...
+
+        # html special characters in placeholders are escaped in certain cases(depending on the
+        # sanitizer configuration) which results in a divergence from the un-sanitized
+        # placeholder stored in self.values.
+        # sanitized_values is being introduced to address this issue. It will now be used by
+        # get_replacement method to get the value to replace the placeholder.
+        if self.values:
+            self.sanitized_values = InsensitiveDict({
+                self.sanitizer(k): v for k, v in self.values.items()
+            })
```
quis, line 103: You probably don’t need the `if` here. If `values` is falsey then it gets coerced to `{}` here:
https://github.com/alphagov/notifications-utils/blob/79cc8da3e1f871d269b6c6f85d11ac20c3858e98/notifications_utils/field.py#L113

I suspect that `self.sanitized_values = InsensitiveDict({})` would be optimised to take negligible time.

That would look like:
```suggestion
        self.sanitized_values = InsensitiveDict({
            self.sanitizer(k): v for k, v in self.values.items()
        })
```

***

You could also rewrite it as a cached property, meaning its only computed when its needed, not on every `__init__`:
```python
class Field:

    […]

    @cached_property
    def sanitized_values(self):
        return InsensitiveDict({
            self.sanitizer(k): v for k, v in self.values.items()
        })
```

***

And if you wanted to go one step further you could subclass `InsensitiveDict` so that you don’t have to sanitise every key:

```python
class SanitisedKeyInsensitiveDict(InsensitiveDict):

    def __init__(self, row_dict, sanitiser):
        self._sanitiser = sanitiser

    @staticmethod
    def make_key(original_key):
        return self.sanitiser(super().make_key(original_key))

class Field:

    […]

    @cached_property
    def sanitized_values(self):
        return SanitisedKeyDict(self.values, sanitiser=self.sanitiser)
```

***

I can show you how to performance test this code if you like, and work out if any of those optimisations are worth doing.

samuelhwilliams:
@quis @BlessedDev 

## Methodology
```python
from timeit import timeit

setup = """
from notifications_utils.field import Field
template_content, data = (
    "(( condition ?? some text )) Lorem Ipsum is simply dummy text of the printing and typesetting industry. ((warning)) Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. ((warning)) It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. ((warning)) It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum. ((warning))",
    {"warning": "some text", "condition": "yes"},
)
def do_stringification():
    return str(Field(template_content, data))
"""

print('1,000 do_stringification() took: %.3fs' % timeit(stmt='do_stringification()', setup=setup, number=1_000))
print('1,000 do_stringification() took: %.2fs' % timeit(stmt='do_stringification()', setup=setup, number=1_000))
print('1,000 do_stringification() took: %.2fs' % timeit(stmt='do_stringification()', setup=setup, number=1_000))
print('1,000 do_stringification() took: %.2fs' % timeit(stmt='do_stringification()', setup=setup, number=1_000))

```

## Test case 1: baseline, current behaviour
### Implementation
```python
    def replace_match(self, match):
        placeholder = Placeholder.from_match(match)
        replacement = self.values.get(placeholder.name)

        if placeholder.is_conditional() and replacement is not None:
            return placeholder.get_conditional_body(replacement)

        replaced_value = self.get_replacement(placeholder)
        if replaced_value is not None:
            return self.get_replacement(placeholder)

        return self.format_match(match)

    def get_replacement(self, placeholder):
        replacement = self.values.get(placeholder.name)
        ...
```

### Result
```
1,000 do_stringification() took: 0.54s
1,000 do_stringification() took: 0.53s
1,000 do_stringification() took: 0.54s
1,000 do_stringification() took: 0.54s
```

## Test case 2: minimal switch
### Implementation
```python
    def replace_match(self, match):
        placeholder = Placeholder.from_match(match)
        replacement = self.values_with_sanitized_keys.get(placeholder.name)

        if placeholder.is_conditional() and replacement is not None:
            return placeholder.get_conditional_body(replacement)

        replaced_value = self.get_replacement(placeholder)
        if replaced_value is not None:
            return self.get_replacement(placeholder)

        return self.format_match(match)

    def get_replacement(self, placeholder):
        replacement = self.values_with_sanitized_keys.get(placeholder.name)
        ...
```

### Result: ~10% slower than baseline
```
1,000 do_stringification() took: 0.668s
1,000 do_stringification() took: 0.65s
1,000 do_stringification() took: 0.65s
1,000 do_stringification() took: 0.66s
```

## Test case 3: optimised switch
### Implementation
```python


    def replace_match(self, match):
        placeholder = Placeholder.from_match(match)
        replaced_value = self.get_replacement(placeholder)

        if placeholder.is_conditional() and replaced_value is not None:
            return placeholder.get_conditional_body(replaced_value)

        if replaced_value is not None:
            return replaced_value

        return self.format_match(match)

    def get_replacement(self, placeholder):
        replacement = self.values_with_sanitized_keys.get(placeholder.name)
        ...
```

### Result: ~10% faster than baseline
```
1,000 do_stringification() took: 0.498s
1,000 do_stringification() took: 0.49s
1,000 do_stringification() took: 0.48s
1,000 do_stringification() took: 0.49s
```

### Test case 3a: optimised switch with HTML escaping
Same as above, but switching out the personalisation key `condition` in the methodology with `cond&ition`, which will force the new implementation to actually sanitize the key.

### Result 3a: ~5% faster than baseline
```
1,000 do_stringification() took: 0.520s
1,000 do_stringification() took: 0.51s
1,000 do_stringification() took: 0.51s
1,000 do_stringification() took: 0.51s
```

quis:
@samuelhwilliams I’ve rebased onto the `master` branch, meaning a smaller change is needed to fix the bug. I’ve also broken the refactoring out into separate commits.

samuelhwilliams (approved):

PR #992, opened 2022-10-04
Remove option to strip HTML from `Field` instances

quis:
When rendering a field we currently give 3 options for how HTML should be treated:
- `html='strip'` (default) removes HTML tags, so `<p>example</p>` becomes `example`
- `html='escape'` replaces characters used in HTML (like `&` and `<`) with their escaped equivalents (like `&amp;` and `&lt;`)
- `html='passthrough'` does nothing, leaving the HTML intact

In almost all cases we use `html='escape'` in order to sanitise untrusted input and prevent cross-site scripting attacks. This is preferable because it leaves non-malicious, HTML-like input looking the same as how the user entered it, rather than removing it.

This means that `html='strip'` is a bad default, because hardly ever used. Where it is used I believe it is because it’s the default, and not because it’s the explicitly desired behaviour:
https://github.com/alphagov/notifications-utils/blob/364c0ed5cbbf621015cd6df79c7ea1d5e0fe32f2/notifications_utils/template.py#L923-L925

It also happens that `strip` is more than 10× slower than `escape` or `passthrough`:
Iterations | Code | Result 1 | Result 2 | Result 3
---|---|---|---|---
1,000 | `str(Field(…, html='passthrough'))` | 0.023s | 0.021s | 0.021s
1,000 | `str(Field(…, html='escape'))` | 0.031s | 0.031s | 0.031s
1,000 | `str(Field(…, html='strip'))` | 0.559s | 0.560s | 0.551s

<details>
<summary>Code used to performance test</summary>

```python
import timeit

setup = """
from notifications_utils.field import Field
template_content, data = (
    "(( cond&ition ?? some text )) Lorem Ipsum is simply dummy text of the printing and typesetting industry. ((warning)) Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. ((warning)) It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. ((warning)) It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum. ((warning))",
    {"warning": "some text", "cond&ition": "yes"},
)
def do_stringification(html):
    return str(Field(template_content, data, html=html))
"""
n = 1_000

for html in ('passthrough', 'escape', 'strip'):
    result = timeit.Timer(
        stmt=f"do_stringification('{html}')",
        setup=setup
    ).repeat(3, number=n)
    formatted_result = ', '.join(map(lambda x: f'{x:.3f}s', result))
    print(f"{n:,}× Field(…, html={html}) – {formatted_result}")
```
</details>

***

So this pull request:
- removes the `html='strip'` option
- makes `html='escape'` the default
- removes `formatters.strip_html` which was doing the underlying work
- removes `bleach` as a dependency, since it was only being used by `formatters.strip_html` 

Technically this is a breaking change and I’ve versionned it as such. There are two places in the admin or API codebases where we use `Field` directly. Both specify `html='escape'` so this shouldn’t require code changes:
- https://github.com/alphagov/notifications-admin/blob/58b3584f9cc7918b38704b48015933a4156ce98d/app/utils/__init__.py#L86
- https://github.com/alphagov/notifications-admin/blob/58b3584f9cc7918b38704b48015933a4156ce98d/app/formatters.py#L277

`formatters.strip_html` is not being used outside this codebase as far as I can tell.


samuelhwilliams (approved):
Nice optimisation 👍

PR #993, opened 2022-10-04
No logfile ecs

sakisv:
### What

This introduces a backwards-compatible check of whether or not we're running in ECS and, if we are, do not configure logging to file. I also included some `TODO` comments to help up cleanup when we're fully migrated.

### Why

On PaaS we used logging to file so that we can ship logs to cloudwatch for long-term storage. When we move to ECS, the logs will automatically go there. (We _will_ need to find what to do with kibana and short-term storage, but that's way out of scope of this PR/repo).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 57.1.0
+
+* Do not log to file when `NOTIFY_RUNTIME_PLATFORM` is set to `ecs`
 
 ## 57.0.0
 
--- a/notifications_utils/version.py
+++ b/notifications_utils/version.py
@@ -5,4 +5,4 @@
 # - `make version-minor` for new features
 # - `make version-patch` for bug fixes
 
-__version__ = '57.0.1'  # 578d19dbd49ee204e0ef3f6828fb713d
+__version__ = '57.1.0'  # 65ac448c9dd1cef13a61edebb78680f2


notifications_utils/logging.py
```diff
@@ -20,12 +20,16 @@ def init_app(app, statsd_client=None):
     app.config.setdefault('NOTIFY_LOG_LEVEL', 'INFO')
     app.config.setdefault('NOTIFY_APP_NAME', 'none')
     app.config.setdefault('NOTIFY_LOG_PATH', './log/application.log')
+    app.config.setdefault('NOTIFY_RUNTIME_PLATFORM', 'local')
```
idavidmcdonald, line 23: Optional

I'm interested in whether we should even have a default value. Could/should the default just be `None`? 

I guess it feels a bit weird that when this will be run shortly in production in the PaaS that the runtime platform will be set as `local`. Will we go and set it to be `paas` for all the apps (which feels unnecessary)?

Also `local` to me says a developer machine. Is that what you meant local to mean or do you mean local to mean running locally ie not in a docker container?

idavidmcdonald (approved):
Looks good.

I left a small optional one about semantics but it is pretty unimportant and this importantly this code will do what we want it to do (and is quite clear that you need to set the environment variable as `ecs`)

notifications_utils/logging.py
sakisv: Yeah, I wasn't particularly happy with any of the options tbh.

I think your suggestion makes sense, since we will only care about this option on ECS, we may as well set it to `None` by default, to avoid having to go and set it to `paas` for all the apps.

PR #994, opened 2022-10-18
Stop mangling pre-escaped URL hrefs

samuelhwilliams:
Our quote/unquote dance when trying to sanitise URLs going into the `href` attribute of `<a>` tags is destructive and can break the semantics of the URL we've been provided.

An example of this is when we receive a link with a redirect query parameter. This parameter is likely to already be URL encoded - turning eg `/` into %2F. When we quote then unquote this, we turn it into a native `/` which isn't what we originally received. When a recipient clicks this link, the server might not expect it to have been unencoded. Or, if we unencoded `&` (%26) values - these could fundamentally change what the URL means.

### DIFF
--- a/notifications_utils/formatters.py
+++ b/notifications_utils/formatters.py
@@ -122,11 +122,15 @@ def autolink_urls(value, *, classes=''):
 
 def create_sanitised_html_for_url(link, *, classes='', style=''):
     """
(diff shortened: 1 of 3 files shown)


CrystalPea (approved):
Looks good to me. 

Since we don't allow any brackets, I don't think this patch could increase any threats?

PR #1000, opened 2022-11-21
Add black

leohemsted:
Exact same steps as https://github.com/alphagov/notifications-admin/pull/4367

except:

* black didn't quite fix some strings. it made them too long and then flake8 complained 😞. strings fixed manually in 780583015fea471b898cdaab44d6490594f57415.
* bump_version.py produced non-black-compliant python code. fixed in b5d2690fe085c82d723f307297efebeb33baf520

Remember to run the following two lines in your local utils repo:

```bash
pre-commit install --install-hooks
git config --local blame.ignoreRevsFile .git-blame-ignore-revs
```


setup.cfg
```diff
@@ -9,8 +9,9 @@ xfail_strict=true
 select=C, E, F, W, B, B901, B902
 
 # W503: line break before binary operator
-# W504 line break after binary operator
-ignore = W503, W504
+# W504: line break after binary operator
+# E203: space before/after :
+ignore = W503, W504, E203
```
klssmith, line 14: I don't think you need to ignore `E203` because it only checks for whitespace before ':' (not after) and there is nothing that fails with it ignored.

klssmith (approved):

PR #1001, opened 2022-11-21
make HTMLEmailTemplate take brand_alt_text instead of brand_name

leohemsted:
- [x] https://github.com/alphagov/notifications-utils/pull/1000

this is a breaking change, but is just a variable rename. It felt good to do just to avoid any confusion with what each field is for though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 58.0.0
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 57.8.0
```
klssmith, line 6: Was this meant to be `58.0.0` to match version.py?

leohemsted: oops yes, typo. fixed.

klssmith (approved):

PR #1013, opened 2022-12-07
Support notification type in rate limit cache key

samuelhwilliams:
We want to move from having a single message limit per day to having daily limits based on the notification type. Let's support both for now so we can roll out backwards-compatible changes.

### DIFF
--- a/notifications_utils/clients/redis/__init__.py
+++ b/notifications_utils/clients/redis/__init__.py
@@ -3,8 +3,13 @@ from datetime import datetime
 from .request_cache import RequestCache  # noqa: F401 (unused import)
 
 
-def daily_limit_cache_key(service_id):
-    return f"{service_id}-{datetime.utcnow().strftime('%Y-%m-%d')}-count"
+def daily_limit_cache_key(service_id, notification_type=None):
+    yyyy_mm_dd = datetime.utcnow().strftime("%Y-%m-%d")
+
+    if not notification_type:
+        return f"{service_id}-{yyyy_mm_dd}-count"
+
+    return f"{service_id}-{notification_type}-{yyyy_mm_dd}-count"
 
 
 def rate_limit_cache_key(service_id, api_key_type):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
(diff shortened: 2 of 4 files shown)


idavidmcdonald:
Does this mean we will need to introduce limits for how many text messages and how many emails services can send?

What happens if they want to send a split, say 250k, which varys day by day?

I didn't go to your kick off for this work but I think I envisioned a new key specific for letters but then keeping the remaining key that would be an overall count of notifications so we could still keep an overall notification type agnostic rate limit?

notifications_utils/clients/redis/__init__.py
```diff
@@ -3,8 +3,11 @@
 from .request_cache import RequestCache  # noqa: F401 (unused import)
 
 
-def daily_limit_cache_key(service_id):
-    return f"{service_id}-{datetime.utcnow().strftime('%Y-%m-%d')}-count"
+def daily_limit_cache_key(service_id, notification_type=None):
+    if not notification_type:
+        return f"{service_id}-{datetime.utcnow().strftime('%Y-%m-%d')}-count"
+
+    return f"{service_id}-{notification_type}-{datetime.utcnow().strftime('%Y-%m-%d')}-count"
```
quis, line 10: I was tempted to do this refactor when I moved this to using f-strings. Now that we’re repeating it I think it’s definitely worth refactoring into something a bit more readable:
```suggestion
    yyyy_mm_dd = datetime.utcnow().strftime('%Y-%m-%d')
    
    if not notification_type:
        return f"{service_id}-{yyyy_mm_dd}-count"

    return f"{service_id}-{notification_type}-{yyyy_mm_dd}-count"
```

samuelhwilliams:
@idavidmcdonald At the kick-off we decided to move to having a full split by notification type, so limits for each kind, and move away from the single central limit.

For emails it will probably be 'barely' limited at all, for text messages we are going to try to work out a fair level based on historical data, and for similarly for letters.

idavidmcdonald:
Interesting, it feels like that is a lot more work that way (comms to users, picking appropriate limits, code changes in more places, more user interface changes potentially) but I'm not quite sure what the user need is or what the benefits are to the user or us of having a different limit for emails vs sms...

I'll leave the features team to figure it out and happy to back your decision regardless but feel free to give it more thought/discussion if you think appropriate

quis (approved):

PR #1015, opened 2022-12-15
Update code for version 3.4.1 onwards of pyproj dependency

klssmith:
Version 3.4.1 of pyproj changed the ESPG codes to be upper case, which means that our tests now fail (https://github.com/pyproj4/pyproj/pull/1162).

This change updates our code to also use upper case codes, and bumps the minimum version of pyproj to ensure that we're always using a version which also uses the upper case codes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 60.0.0
+
+* Bump pyproj to be version 3.4.1  or greater. This changes the ESPG codes to be upper case, which affects the how
+  `Polygons` class transforms data.
+
 ## 59.3.0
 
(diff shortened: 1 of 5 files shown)


quis (approved):

samuelhwilliams:
We should probably add a version bump to this PR and check that it can be installed into our other apps without causing issues there (it probably shouldn't 🤞)

klssmith:
I've bumped the version and tested it out - we'll need to rebuild the broadcast area library in admin once admin is using this version of utils

samuelhwilliams:
Cool - thanks for investigating. Let's do this as a major (breaking) version bump then, I think. If you update that and add a changelog entry/etc I think it's good to go in :)

klssmith:
This is now version `60.0.0` 👍

samuelhwilliams (approved):
Nice work 👏

PR #1017, opened 2022-12-30
Use locally-cached bank holidays data

samuelhwilliams:
govuk_bank_holidays is a package maintained by the Ministry of Justice which provides functionality to detect working/non-working days, including weekends and bank holidays. They have cached bank holidays data as part of their package. This means that any apps pulling in notifications_utils and checking bank holidays data need to manually keep the indirect dependency govuk_bank_holidays up-to-date, which we don't have a good process for.

Let's instead cache the bank holidays ourselves and use our own data. We can add a test to notifications_utils to check that bank holiday data is up-to-date, which gives us greater assurance that our non-working-day checks will remain valid into the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 61.0.0
+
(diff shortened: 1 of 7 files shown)


quis (approved):
75uzk9

PR #1021, opened 2023-02-20
Downgrade errors when a suspended user tries to create a Zendesk ticket

klssmith:
We're seeing increasing numbers of errors in our logs because the Zendesk client has returned a `422` status code. This happens when a user has been suspended due to submitting spam tickets.

When this happens, we now log a warning and not an error.

### DIFF
--- a/notifications_utils/clients/zendesk/zendesk_client.py
+++ b/notifications_utils/clients/zendesk/zendesk_client.py
@@ -62,6 +62,10 @@ class ZendeskClient:
         )
 
         if response.status_code != 201:
+            if response.status_code == 422 and self._is_user_suspended(response.json()):
+                error_message = response.json()["details"]
+                current_app.logger.warning(f"Zendesk create ticket failed because user is suspended '{error_message}'")
+                return
             current_app.logger.error(
                 f"Zendesk create ticket request failed with {response.status_code} '{response.json()}'"
             )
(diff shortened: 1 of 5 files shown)


notifications_utils/clients/zendesk/zendesk_client.py
```diff
@@ -71,6 +75,10 @@ def send_ticket_to_zendesk(self, ticket):
 
         current_app.logger.info(f"Zendesk create ticket {ticket_id} succeeded")
 
+    def _is_user_suspended(self, response):
+        requester_error = response["details"].get("requester")
+        return requester_error and ("suspended" in requester_error[0]["description"])
```
samuelhwilliams, line 80: Should this check all of the requester_errors rather than just the first?

Or potentially this should only suppress the error if the only error is the user being suspended and we'd be interested to know about other errors?

klssmith: I couldn't find documentation about possible errors and what they would look like so wanted to only downgrade errors due to the user being suspended and not change what happens with anything else (we could then decide what to do with others if we came across them).

I have only ever seen errors due to someone being suspended take this format `{'requester': [{'description': 'Requester: Name is suspended.'}]}`. I think for anything else it might be useful to see the error.

samuelhwilliams (approved):

samuelhwilliams:
Do we still want to get this in? :)

klssmith:
I can't actually remember the last time I saw these errors now - they seem to have stopped once I opened the PR! I'll close this for now since it's out of date anyway, then might re-open if it starts happening again.

PR #1025, opened 2023-04-06
Add odd/even class to all letter pages

tombye:
[This pull request on admin](https://github.com/alphagov/notifications-admin/pull/4648) introduced a visual bug due to its use of `nth-of-type` assuming `.letter:nth-of-type(even)` would affect even-numbered elements matching this selector: `div.letter`.

It actually matches even-numbered `div` elements, filtered by whether they have the `letter` class or not.

You will be able to match even-numbered pages in CSS in future (see https://css-tricks.com/css-nth-of-class-selector/) but until then, we can use Jinja to add ~~a specific class~~ classes to indicate if a page is odd or even this instead.

Uses BEM syntax to indicate a state of page. There is currently no CSS in admin targeting this class so this can be merged with no impact.

Edit: I've updated this so it adds either `page--odd` or `page--even` to all pages as this seemed useful (taken from https://jinja.palletsprojects.com/en/3.0.x/tricks/#alternating-rows). My branch name now sounds silly.

### DIFF
--- a/notifications_utils/jinja_templates/letter_image_template.jinja2
+++ b/notifications_utils/jinja_templates/letter_image_template.jinja2
@@ -1,5 +1,5 @@
 {% for page_number in page_numbers %}
-  <div class="letter">
+  <div class="letter page--{{ loop.cycle('odd', 'even') }}">
     {% if loop.first and show_postage %}
       <p class="letter-postage {{ postage_class_value }}">
         Postage: {{ postage_description }}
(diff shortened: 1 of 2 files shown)


quis (approved):
You’ll need to run `make version-minor` but the detail of this change is good. Also CSS: 🤯

tombye:
Yeah, I didn't quite realise it worked like that too. I also tested the newer syntax with admin and, while it's supported by Chrome, Edge and latest Firefox, our Sass setup is messing up the expression in the brackets so it's not even an easy win (`nth-child(even of .letter)` becomes `nth-child(evenof.letter)`). 😞

PR #1031, opened 2023-05-09
Add `class="page--first"` and `class="page--last"`

quis:
…to the first and last pages (respectively) of letters rendered with `LetterImageTemplate`.

This will let us target them with CSS which we can’t always do at the moment because of the limitations of selectors like `:first-child` and `:first-of-type`

Extends the approach taken in https://github.com/alphagov/notifications-utils/pull/1025

This will let us fix a bug where at the moment all letter pages have `0` bottom margin if the attach pages button is visible, whereas this was supposed to apply to the last page only:

www notifications service gov uk_services_ea3b5c2d-d954-4907-a3ac-5bb9cef05005_templates_2ae6f848-ab56-487b-92cd-21e352cb05c0(iPad-Air)-(1)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 62.4.0
+
+* Add `class="page--first"` and `class="page--last"` to the first and last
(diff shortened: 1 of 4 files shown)


tombye (approved):
Seems like a sensible addition to the existing classes 👍🏻

PR #1032, opened 2023-05-09
Fix bug with `page--first` and `page--last` classes

quis:
Didn’t account for a one-pager, where the single page is both the first and last page of the letter.


tombye (approved):
I missed that too in my last review. This looks like a good fix 👍🏻

PR #1033, opened 2023-05-11
Remove auto-log formatting and standardise format

samuelhwilliams:
### ✅  ~Waiting for Ruff release including https://github.com/charliermarsh/ruff/pull/4253~

Removes the custom functionality to automatically try to inject variables into the log string. A whole host of our log messages were erroring out on this functionality because the variable didn't exist, or because we were logging out dictionaries which "".format() was trying, and failing, to handle.

Now that we've integrated Sentry, let's standardise on the "old-style" logging format, which looks something like this:

logger.info("Response %s took %.4f to generate", response.body, duration)

Python's logging system will automatically interpolate the message with the provided arguments, and sentry will group any error-level log messages as a single issue (the way we used to do things - like `logger.error(f"something went wrong: {exc}") would generate a unique sentry issue if exc generated a unique string for each instance of the error).

This is a breaking change: when pulling into any of the apps we should go through and systematically update all of the log messages to the "old-style" %s format. We can enforce this with the "G" rule in Ruff.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,7 @@ repos:
   - id: check-yaml
   - id: debug-statements
 - repo: https://github.com/charliermarsh/ruff-pre-commit
-  rev: 'v0.0.261'
+  rev: 'v0.0.272'
   hooks:
     - id: ruff
       args: [--fix, --exit-non-zero-on-fix]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+# 65.0.0
+
+* Remove automatic formatting from JSONFormatter. Any log messages using `{}` to inject strings should be converted
+  to "old-style" log messages using %s and passing variables as arguments to the log function. Do not eagerly
+  interpolate the string (eg "log: {}" % ("string") - let Python's logging module do this itself. This is to provide
+  compatability with Sentry. Add the "G" rule to Ruff's checks to enforce this.
+* Removes `CustomLogFormatter` altogether, as its only purpose was the auto-formatting as above.
+
 ## 64.2.0
 
 * `LetterImageTemplate` now adds a hidden element marking the first page of any attachment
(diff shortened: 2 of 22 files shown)


whpearson (approved):
Changes make sense.

samuelhwilliams:
@leohemsted you had thoughts on this when I posted it originally - any conclusions?

leohemsted:
ugh i tried to write a big long comment last week but it got lost by a chrome restart before i hit save.

thoughts:

* i'm in favour of using ruff to remove any instances of `log(f"f-string with {my_obj}")`. (the sentry argument is more compelling than the "can save expensive computation if logging is disabled" since we don't really play with conditionally logging and differing log levels like that, but both are fair reasons i guess)
* i really dislike % formatting, i think it's obtuse and hard to read, and is against progression of how formatting strings generally look in multiple languages, which in my experience generally tend towards named variables in curly braces. i would _consider_ keeping the custom formatter, but us actually using it and calling `log("some normal {object}", object=my_obj)` (or maybe `log("some normal {object}", extras={"object": my_obj})`. That would solve the issue of passing formatted print content that contains curly braces in to our outputs. that would however mean our code is special snowflake, which is always a bit frustrating.

keen to hear your thoughts. i know you said that most places you've worked have just accepted the inconsistency in formatting across their codebase, but i'm not 100% convinced by that - you've spent plenty of time on notify working to make sure we don't just accept "this is the way someone who is no longer on the team chose to do things when trying to quickly build a prototype seven years ago". i've also checked a couple of open source projects and found plenty of repos doing things incredibly inconsistently (and just look at that unhinged django one). shoutout to numpy for as far as i can tell being the only place i found that has a decent amount of log messages that are all formatted consistently with %

||image|
|--|--|
| image | image |


that said, my first bullet point is a bigger desire than my dislike in the second bullet point, and if you're a hard no on the custom formatter then i might be able to compromise though i'll probably still whinge a bunch 😉. 

thoughts?

leohemsted (approved):
lets go with #1 then. 

im happy with your point about the extras being designed for decorating the rest of the log payload (i assume by this you mean the extras is metadata that'll get attached in separate fields such as timestamps, app names, request IDs, etc), so happy to avoid any risk of clashing with those fields and causing unexpected values in either the log message or that additional metadata

PR #1035, opened 2023-05-23
Move `postage` argument to base class

quis:
At the moment we are forced to use `LetterImageTemplate` in a few places in the admin app because it’s the only one that has a `postage` attribute. The code would be less complex if we could use `LetterPreviewTemplate` everywhere.

Moving the handling of the `postage` argument to the base class gets the classes a bit closer to being polymorphic. This means that any subclass (including `LetterImageTemplate` and `LetterPreviewTemplate`) will always have a `postage` attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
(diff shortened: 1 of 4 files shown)


quis:
Don’t think we need this actually

PR #1037, opened 2023-05-31
Add some useful properties to letter template classes

quis:
This means that it can be overidden by subclasses.

Previously if a subclass tried to implement a `page_count` property then it would cause an exception where the `__init__` method is trying to assign to `self.page_count`.

Letting subclasses define their own `page_count` will mean there’s a nice place to put logic about how many pages are in a templated letter with an attachment, and on what page the attachment starts.

***

For convenience we can then add the following useful attributes and properties to all letter template classes:
- `max_page_count`
- `max_sheet_count`
- `too_many_pages` (requires subclasses to implement `page_count`)

### DIFF
--- a/notifications_utils/template.py
+++ b/notifications_utils/template.py
@@ -813,8 +813,12 @@ class LetterImageTemplate(BaseLetterTemplate):
                 )
             )
         self.image_url = image_url
-        self.page_count = int(page_count)
+        self._page_count = int(page_count)
         self._postage = postage
+
+    @property
+    def page_count(self):
+        return self._page_count
 
     @property
     def postage(self):
--- a/notifications_utils/template.py
+++ b/notifications_utils/template.py
@@ -782,6 +782,7 @@ class LetterImageTemplate(BaseLetterTemplate):
 
     jinja_template = template_env.get_template("letter_image_template.jinja2")
     first_page_number = 1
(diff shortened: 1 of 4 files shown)


notifications_utils/template.py
```diff
@@ -655,6 +655,8 @@ def subject(self):
 class BaseLetterTemplate(SubjectMixin, Template):
 
     template_type = "letter"
+    max_page_count = LETTER_MAX_PAGE_COUNT
+    max_sheet_count = max_page_count // 2
```
samuelhwilliams, line 659:
```suggestion
    max_sheet_count = LETTER_MAX_PAGE_COUNT // 2
```

This reads a bit funny as-is; feels like it implies if you override `max_page_count` in a subclass then `max_sheet_count` will respect that change, but it won't. So probably better to derive from the external constant or turn it into a `@property` - don't mind which.

samuelhwilliams (approved):

notifications_utils/template.py
quis: Changed in 0d50d5e

>    **Derive sheet count from constant**
>
>    This reduces the implication that overriding `max_page_count` on an
>    instance will also update the sheet count.
>
>    Prefer this to implementing an `@property` because this way it’s still
>    accessible on both the class and any instances.

PR #1039, opened 2023-06-02
Allow subheadings via ## markdown

samuelhwilliams:
Previously we collapsed all level-2 headings (with ##) down to basic paragraphs. Our users want more formatting options and this is something we can offer. We don't want to offer bold because we know that it can be overused and lead to worse quality content, but subheadings may allow services to better structure content and still draw attention to important things.

image
image
image


notifications_utils/markdown.py
```diff
@@ -201,6 +210,15 @@ def header(self, text, level, raw=None):
                     "-" * self.COLUMN_WIDTH,
                 )
             )
+        elif level == 2:
+            return "".join(
+                (
+                    self.linebreak() * 3,
+                    text,
+                    self.linebreak(),
+                    "-" * (self.COLUMN_WIDTH // 2),
```
leohemsted, line 219: github allows `======` for h1s and `-------` for h2s, would it be worth us adopting that pattern in our plaintext?

this is an h1
========

this is an h2
-----------

```
this is an h1
========

this is an h2
-----------
```

samuelhwilliams: Will tweak 👍

samuelhwilliams: Ah, we already use `========` for hrules - is that still going to be fine?

samuelhwilliams: Provisionally added in 99a110aafb6db6c6b9417d19dc76cf511b7c37cb

samuelhwilliams: I can get on board with option 2 as well 👍

samuelhwilliams: Tweaked in 076f9744ad49d2fcd3a3cbf3783f343d76e22db6

quis:
Rationale for the `19px` type size for subheadings:
1. We don’t want to change the existing 27px size for `<h1>`s because this could be a breaking change for some of our users emails
2. The next size down in the scale (`24px`) is not visually different enough from `27px`
3. It’s ok that our large/medium (`27px`/`19px`) is smaller than the Design System because our sizes are also used on mobile. There’s no responsive typography in email templates, so the sizes we use for headings are somewhere in between the sizes the Design System uses on desktop (`36px`/`24px`) and the sizes they use on mobile (`24px`/`18px`)

Rationale for the extra spacing:
- `19px` bold type by its own doesn’t look enough like a heading
- by adding some extra space before level one and two headings we can make sure they visually group with the text they precede.

How it looks:

image

notifications_utils/markdown.py
```diff
@@ -67,6 +68,8 @@ def block_quote(self, text):
     def header(self, text, level, raw=None):
         if level == 1:
             return super().header(text, 2)
+        elif level == 2:
+            return super().header(text, 3)
```
quis, line 72: I don’t think we want to tackle headings in letters yet because there’s some additional complication there (the [story](https://trello.com/c/OLMQ3ZSy/389-add-subheadings-to-available-markdown) does say ‘for emails only’)
```suggestion
```

samuelhwilliams: I'll remove it for letters 👍  One of the open questions was also `do we want to do this for letters?` so I did it in case. But easily removed 👍 )

quis (approved):

PR #1042, opened 2023-06-07
Invisibly mark the first page of a letter attachment

quis:
So that we can jump directly to it using an anchor link.

Separate element instead of adding an `id` to the image so that we can control the exact positioning without changing the position of the visible element (i.e. the preview of the page itself).

Also adds `loading=eager` to the first page of the attachment as a hint to browsers. If the user is jumped straight to the first page of the attachment then we should try to load that as quickly as possible, rather than loading all the intermediary templated pages first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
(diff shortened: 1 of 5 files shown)


whpearson (approved):
Changes look good, nice eager loading of the first page of the attachment.

PR #1045, opened 2023-06-30
update international sms multipliers

leohemsted:
note: some countries were not specified in the source spreadsheet so were defaulted to four billable units

code used to generate this was:
```
new_rates = {"1": 1, "20": 3, ......}
billing_yml_path = Path(__file__) / "international_billing_rates.yml"
with open(billing_yml_path) as yml_file:
    data = yaml.safe_load(yml_file)

for country_code, metadata in data.items():
    if country_code not in new_rates:
        metadata["notes"] = "# not specified in MMG multiplier sheet so set to default of 4"
    metadata["billable_units"] = new_rates.get(country_code, 4)

with open(billing_yml_path, "w") as yml_file:
    yml_file.write(yaml.dump(data, default_flow_style=False, sort_keys=False))
```

then afterwards, i manually went through the international_billing_rates file to convert the little "notes" into a normal yaml comment, and restore existing comments in the file

### DIFF
--- a/notifications_utils/international_billing_rates.yml
+++ b/notifications_utils/international_billing_rates.yml
@@ -1381,8 +1381,7 @@
     sc: null
     sender_and_registration_info: null
     text_restrictions: null
-  # not specified in MMG multiplier sheet so set to default of 4
-  billable_units: 4
+  billable_units: 3
   names:
   - Aruba
 '298':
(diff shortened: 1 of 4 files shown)


CrystalPea (approved):
Would be good to have some information/link in the PR message about why this is happening.

Like, is it because MMG changed the multipliers, or are those general multipliers set on some higher level?

leohemsted:
Good point! In this case it's just mmg updating their prices. Some multipliers have gone up (eg russia), some have gone down (eg rwanda). many were 1 previously but weren't specified in the most recent spreadsheet. I wouldn't mind clarifying if those that weren't specified in the sheet have always been 3 or 4 fragments or if thats just a new policy for them, but that will have to wait until next week.

the sheet can be found here: https://trello.com/c/UjEH9sS8/398-update-international-sms-prices-with-latest-figures

PR #1048, opened 2023-07-18
Render a QR code if user provides us a QR code link

CrystalPea:
Here you can see how it looks like:

Screenshot 2023-07-19 at 11 33 12

Story ticket: https://trello.com/c/LTHaNofS/411-convert-markdown-links-in-letters-to-qr-codes

To test it, check out template-preview branch to `update-weasyprint` (and change version of utils, too).


tests/test_markdown.py
```diff
@@ -614,6 +614,16 @@ def test_link_with_title(markdown_function, expected):
...
 
 
+def test_letter_qr_code():
+    expected_qr_code_start = "<p><div class='qrcode'><svg viewBox=\"0 0 25 25\">"
+    assert notify_letter_preview_markdown('[qr](http://example.com")').startswith(expected_qr_code_start)
+
+
+def test_letter_qr_code_works_with_extra_whitespace():
+    expected_qr_code_start = "<p><div class='qrcode'><svg viewBox=\"0 0 25 25\">"
+    assert notify_letter_preview_markdown('[ qr ](http://example.com")').startswith(expected_qr_code_start)
```
samuelhwilliams, line 624: I think I'd quite like to test the entire QR code SVG remains the same as I think it should be a static value. Or does it change every time you generate it?

Gives us a bit more confidence+warning if (for example) we update the library and the implementation changes under the hood.

In the mid/long-term I'd like a functional test that generates a letter with a QR code, and then scans the QR code to validate the URL is expected. This would give us end-to-end confirmation that the QR is generated correctly and readable.

The reason I'm thinking we need some changes like this are because this test (as it stands) wouldn't detect that  segno QR code was broken by the markdown renderer replacing hyphens with designer hyphens.

CrystalPea: Okee I updated the first test 😁

notifications_utils/jinja_templates/letter_pdf/_main_css.jinja2
```diff
@@ -193,5 +194,24 @@
     word-wrap: break-word;
     border-radius: 1.05em;
   }
+  .qrcode,
+  .qrcode‑placeholder {
+      display: block;
+      width: {{ line_height_raw * 4.25 }}pt;
+      height: {{ line_height_raw * 4.25 }}pt;
```
quis, line 201: This should be
```suggestion
      width: {{ line_height_raw * 5.25 }}pt;
      height: {{ line_height_raw * 5.25 }}pt;
```
as per https://trello.com/c/cMj9CTop/408-finalise-size-and-spacing-of-qr-codes#comment-6493182d313cc01d3f3ee18f

PR #1052, opened 2023-08-08
Style the QR code placeholder so it looks better when the placeholder text is long

quis:
When a QR code contains a placeholder we render a ‘placeholder’ box instead, which has a dashed border.

If the text of the placeholder or its prefix/suffix is long, then this causes the text to wrap inside the placeholder box. Because the placeholder box is a relatively narrow fixed width this:
- happens easily
- does not look very good
- makes the text harder to read

This commit adds some extra HTML elements which allow us to style the box and its contents separately.

This way the contents can overflow the box without looking messy. The width of the text is therefore limited to the width of the column, as it is for other URLs and placeholders.

# Before 

image

# After 

image

***

By colour-coding the elements you can see how this is constructed:

image

***

Ideally this would use `vertical-align: middle` to centre the text, even when it spans multiple lines. But Weasyprint doesn’t seem to support this, so using a negative margin to vertically centre one line of text will have to do. It looks like this:

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 66.0.2
+
+* Style the QR code placeholder so it looks better when the placeholder text is long
+
 ## 66.0.1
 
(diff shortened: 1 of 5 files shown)


samuelhwilliams (approved):
This does look better for me 👍  Nice

image

PR #1056, opened 2023-08-14
Validate QR code length for letter templates

samuelhwilliams:
Adds functionality to letter templates and `RecipientCSV` to validate the length of QR codes that are present in templates and notifications.

---

Testing a small template locally with a CSV upload and wrapping https://github.com/alphagov/notifications-admin/blob/4d86205431275c24079605d5cc058308b49ee2c0/app/main/views/send.py#L604-L617 in a timing block:

Before: Took 0.7184392499999994 for 9987
After: Took 1.4068145829999992 for 9987

So it takes about twice as long to process on my speedy uncontended macbook. We consider this to be fairly acceptable at the moment for QR codes because:

- We generally have fewer bulk sends of huge numbers of letters at once.
- QR codes are going into private beta for now, not public release.


notifications_utils/template.py
```diff
@@ -704,6 +706,15 @@ def too_many_pages(self):
     def postal_address(self):
         return PostalAddress.from_personalisation(InsensitiveDict(self.values))
 
+    def has_qr_code_with_too_much_data(self):
+        content = self._personalised_content if self.values else self.content
+        try:
+            Take(content).then(notify_letter_qrcode_validator)
```
quis, line 712: `Take` is only needed if you’re passing the content through multiple functions:
```suggestion
            notify_letter_qrcode_validator(content)
```

notifications_utils/recipients.py
```diff
@@ -381,6 +387,11 @@ def __init__(
                 self.message_too_long = template.is_message_too_long()
             self.message_empty = template.is_message_empty()
 
+            if self.template_type == "letter":
+                self.qr_code_too_long = self._has_qr_code_with_too_much_data()
+            else:
+                self.qr_code_too_long = False
```
quis, line 393: The `if` statement is redundant since you perform the same check inside the method
```suggestion
            self.qr_code_too_long = self._has_qr_code_with_too_much_data()
```

notifications_utils/exceptions.py
```diff
@@ -0,0 +1,11 @@
...
+# on medium error correction settings. Beyond this amount of data we find the QR code can be more difficult to scan.
+QR_CODE_MAX_BYTES = 504
+
+
+class QrCodeTooLong(ValueError):
+    def __init__(self, num_bytes, data):
+        super().__init__(f"Too much data for QR code (num_bytes={num_bytes}, max_bytes={QR_CODE_MAX_BYTES})")
+        self.num_bytes = num_bytes
+        self.max_bytes = QR_CODE_MAX_BYTES
+        self.data = data
```
quis, line 11: Would this make more sense in a file called `qr_code.py`, along with the `qr_code_as_svg` and `qr_code_placeholder` functions?

notifications_utils/template.py
samuelhwilliams: Yeah, it's redundant; I didn't mind that so much if only for the visual similarity that you end up. I don't think it matters too much either way but I can remove it 🤷

notifications_utils/recipients.py
samuelhwilliams: Good spot yeah

notifications_utils/exceptions.py
samuelhwilliams: Fair call, will do 👍

notifications_utils/template.py
samuelhwilliams: (also you end up needing to do `notify_letter_qrcode_validator(str(content))`)

quis (approved):

PR #1057, opened 2023-08-15
Return QR code validation error from check

samuelhwilliams:
The exception raised by the contains useful information about the problem at hand, such as the data passed into the QR code, and the actual & max length of QR codes.

Let's pass this back instead of lose the information by only returning True/False. Then we can potentially display the bad QR code data to the user if it's helpful.


quis (approved):
Returning something other than `True/False` doesn’t feel like a great fit with the existing code here. But it works, and the alternative (separate properties for the number of bytes and the data) feel worse.

PR #1058, opened 2023-08-16
Remove `<strong>` tags from links in QR codes

quis:
We automatically make URLs in letters bold by adding a `<strong>` tag around them. We also remove the protocol (for example `https://`) from the URL.

If we want to reconstruct the full URL later down the line, we need to store the original protocol somewhere.

HTML allows for arbitrary attributes to be created as long as they are prefixed with `data-`. So this PR introduces a new attribute on the `<strong>` tag which stores the original protocol for later use.

***

Adding the `<strong>` tag happens before we look for our QR code syntax.

This means that URLs going into QR codes end up with a small bit of markup in them. This prevents them from going to the original URL when scanned.

This PR:
- removes the `<strong>` tag
- restores the protocol
- preserving any placeholders

Ideally we wouldn’t have to do and then undo this work, but I can’t see a way to either:
- selectively not add the `<strong>` tag if the URL is inside a QR code
- run the QR code detection before running the autolink code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 This is only used for recording changes for major version bumps.
 More minor changes may optionally be recorded here too.
 
+## 68.0.1
+
+* Fix a bug with some HTML getting injected into QR codes
+
 ## 68.0.0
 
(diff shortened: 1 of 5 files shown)


samuelhwilliams (approved):
👍

PR #1061, opened 2023-08-29
Remove empty cell for state of branding

tombye:
For some reason the cell used to hold the branding text causes an error with the JAWS screen reader
when empty (which it is when the 'only branding'
brand has no text).

In this state, and when the content of the email
has a certain form, JAWS announces all links in
the email the same.

This therefore removes that cell when it is empty.

It includes the addition of a max-width to stop
the Outlook app on Android from stretching the
logo to twice the size. It seems strange to keep
the height attribute on the  tag but that is
needed to keep the image working on Outlook on
Windows.

Also note that we only apply the max-width when
there is no text. Setting it on the cell with the
logo image makes it overlap the text in several
email clients.

### DIFF
--- a/notifications_utils/jinja_templates/email_template.jinja2
+++ b/notifications_utils/jinja_templates/email_template.jinja2
@@ -197,15 +197,14 @@
                     <td>
                       <table role="presentation" cellpadding="0" cellspacing="0" border="0" style="border-collapse: collapse">
                         <tr>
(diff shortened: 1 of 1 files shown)


whpearson (approved):
Changes make sense, although the bug doesn't. I've seen the new variant work with the screen readers, so it gets the thumbs up.

PR #1062, opened 2023-08-31
Add a note on versioning to the README

tombye:


README.md
```diff
@@ -23,3 +23,13 @@ make bootstrap
...
+
+## Publishing a new version
+
+Versioning should be done by running the `make version-[type of change]` command, following [semantic versioning](https://semver.org/). For example
+
+```
+make version-patch
+```
+
+Major changes must be recorded in the `CHANGELOG.md` file, but notes on any other types of change are also welcome.
```
samuelhwilliams, line 35: I think we generally do leave comments about all version bumps, even minor and patch, so maybe nice to suggest that?

```suggestion
Include a short summary (sentence or two) about the changes you've made in `CHANGELOG.md`. Please do this even if you're only making a minor or patch version change.
```

PR #1063, opened 2023-08-31
Add a space before `##` header in `make bump-utils`

samuelhwilliams:
We auto-generate a commit message of all the relevant changes added by a utils bump, but we prefix the headers with `##`. Some editors (eg vi/m) will treat these as comments and drop them out.

If we insert a preceding space, they won't be treated as comments but markdown renderers will still treat them as headings.


tombye (approved):
Makes sense to me 👍🏻

PR #1065, opened 2023-09-01
Fix QR code rendering with old [QR]() syntax

samuelhwilliams:
We have switched to using the syntax `QR: link/data` in letter template content for rendering QR codes. However, we use mistune to parse the letter content into HTML, and that just so happens to take our "old" QR code syntax of `[QR](link/data)` and turn it into `QR: link/data`, which then can get picked up as the new-style syntax.

I had a quick look to see if we could force the old-style syntax to only render as links, not QR codes, but the solution felt slightly convoluted (need to feed some indicator/argument between `link`/`autolink` and `paragraph` methods) and I don't think it's really worth doing. We can technically support both syntaxes, but only advertise the simpler one. Note also that the "old-style" `[QR]()` syntax only resolves to an actual QR code if it's the only thing in the paragraph block. If it's mid-paragraph, it will remain as a normal link.

**Note**: I've edited the QR code renderer to show the QR code data as a string below the code for example purposes only.

# Before

| Letter content | Before | After |
|--------|--------|--------|
| `qr: http://www.google.com`<br><br>`[qr](http://www.google.com)` | image | image |
| `qr: just some text`<br><br>`[qr](just some text)` | image | image |
| `qr: some text and then a link https://www.google.com`<br><br>`[qr](some text and then a link https://www.google.com)` | image | image |

### DIFF
--- a/notifications_utils/markdown.py
+++ b/notifications_utils/markdown.py
@@ -101,7 +101,7 @@ class NotifyLetterMarkdownPreviewRenderer(mistune.Renderer):
             # Restore http:// or https:// and strip out the <strong> tag that gets injected by
             # the `link`/`autolink` methods
             text = self._render_qr_data(
-                re.sub(r"<strong data-original-protocol='(https?://)'>(.*?)</strong>", r"\1\2", qr_code_contents)
+                re.sub(r"<strong data-original-protocol='(https?://|)'>(.*?)</strong>", r"\1\2", qr_code_contents)
             )
 
         return f"<p>{text}</p>"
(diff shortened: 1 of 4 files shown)


notifications_utils/markdown.py
```diff
@@ -101,7 +101,7 @@ def paragraph(self, text):
             # Restore http:// or https:// and strip out the <strong> tag that gets injected by
             # the `link`/`autolink` methods
             text = self._render_qr_data(
-                re.sub(r"<strong data-original-protocol='(https?://)'>(.*?)</strong>", r"\1\2", qr_code_contents)
+                re.sub(r"<strong data-original-protocol='(https?://|)'>(.*?)</strong>", r"\1\2", qr_code_contents)
```
quis, line 104: What is the pipe (`|`) doing here?

samuelhwilliams: It's saying that we might sometimes not get a protocol (eg in the case of `[]()` links), and when that happens we still need to strip the strong tags.

PR #1071, opened 2023-11-08
Charge order of invalid address line characters

quis:
So they match the order we display them in the error message. Doing this means we can reference this constant when writing the error message, rather than hard-coding the set of characters.

### DIFF
--- a/notifications_utils/postal_address.py
+++ b/notifications_utils/postal_address.py
@@ -27,7 +27,7 @@ country_UK = Country(UK)
 class PostalAddress:
     MIN_LINES = 3
     MAX_LINES = 7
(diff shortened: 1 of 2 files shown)


quis:
```python
OLD_INVALID_CHARACTERS_AT_START_OF_ADDRESS_LINE = r'[\/()@]<>",=~'
NEW_INVALID_CHARACTERS_AT_START_OF_ADDRESS_LINE = r'@()=[]"\/,<>~'
assert set(OLD_INVALID_CHARACTERS_AT_START_OF_ADDRESS_LINE) == set(NEW_INVALID_CHARACTERS_AT_START_OF_ADDRESS_LINE)
```
✅

samuelhwilliams (approved):

PR #1075, opened 2023-11-23
allow redis_client to take in optional kwargs

leohemsted:
these get passed through to the redis constructor - potentially useful for fields like `db` to define the db to connect to separate from the url


leohemsted:
i'm gonna get rid of this - rather than the app setting redis db number, i think it's more useful for the db url to include the number and be defined in terraform so this is no longer needed

PR #1077, opened 2023-11-29
logging: add request logging to flask app

risicle:
https://trello.com/c/Q5S5m4ge/446-add-http-logging-to-all-of-our-ecs-apps

Though both werkzeug and gunicorn have logging functionality, adding it here allows us to get nice fully structured logs that also include some flask parameters, like the `endpoint` name.

This also imports the `RestrictedAny` family of test utilities I used in digitalmarketplace's `digitalmarketplace-test-utils`.

(Why? The logging hooks themselves are based on those in `digitalmarketplace-utils`, and the tests were adapted from those. Also I like them and nobody told me not to :innocent: )


notifications_utils/logging.py
```diff
@@ -17,11 +19,63 @@
...
+    }
+
+
 def init_app(app, statsd_client=None, extra_filters: Sequence[logging.Filter] = tuple()):
     app.config.setdefault("NOTIFY_LOG_LEVEL", "INFO")
     app.config.setdefault("NOTIFY_APP_NAME", "none")
     app.config.setdefault("NOTIFY_LOG_PATH", "./log/application.log")
     app.config.setdefault("NOTIFY_RUNTIME_PLATFORM", None)
+    app.config.setdefault(
+        "NOTIFY_REQUEST_LOG_LEVEL",  # noqa
```
samuelhwilliams, line 42: What's this `# noqa` for?

risicle: Black was going to put it on one line, which makes the ternary statement look more complicated than it is.

notifications_utils/testing/comparisons.py
```diff
@@ -0,0 +1,93 @@
...
+    dict. e.g.
+
+    >>> [{"a": 123, "b": 456, "less": "predictabananas"}, 789] == [AnySupersetOf({"a": 123, "b": 456}), 789]
+    True
+    """
+
+    def __init__(self, subset_dict):
+        # take an immutable dict copy of supplied dict-like object
+        self._subset_dict = MappingProxyType(dict(subset_dict))
+        super().__init__(lambda other: self._subset_dict == {k: v for k, v in other.items() if k in self._subset_dict})
```
quis, line 47: What do you think about this to minimise use of lambdas?

```suggestion

    def _condition(self, other):
        return self._subset_dict == {k: v for k, v in other.items() if k in self._subset_dict}
```

notifications_utils/logging.py
```diff
@@ -17,11 +19,63 @@
...
+            "status": response.status_code,
+            "duration_real": (
+                (time.perf_counter() - request.before_request_real_time)
+                if hasattr(request, "before_request_real_time")
+                else None
+            ),
+            **_common_request_extra_log_context(),
+        }
+        current_app.logger.getChild("request").log(
+            logging.ERROR if response.status_code // 100 == 5 else logging.INFO,
```
samuelhwilliams, line 73: 🤔 unsure on this. It feels right, but it will probably trigger an error in sentry for each 5xx request, when we'll already be getting (more useful) error reports from the exceptions themselves.

samuelhwilliams: also this log line isn't actually an error itself, it's working correctly, so I might say stick with info.

could be swayed either way

but the sentry thing may need thinking about

risicle: Would it be better to be a warning?

I'd like it if we didn't have to rely 100% on sentry to spot errors.

notifications_utils/testing/comparisons.py
risicle: Yes that works doesn't it

risicle:
Only draft-ing it to indicate I want to do a squash-tidyup before merge.

notifications_utils/logging.py
samuelhwilliams: Trailing comma fixes that. I think the `# noqa` isn't needed.

At least my black and ruff ignore this with the comment removed because of the trailing comma on the following line.

notifications_utils/logging.py
samuelhwilliams: Warning is fine with me.

I'd rather we only have to use and configure one place for central error reporting.

risicle: As for "is it an error at all?", I think I'm trying to catch cases where we end up returning a 5xx which weren't because of an exception being raised - e.g. business logic? Would that get surfaced by anything currently?

Anyway, this whole feature is basically a temporary measure pending a full telemetry solution, so perhaps it's not worth me defending..

notifications_utils/logging.py
```diff
@@ -17,11 +19,63 @@
...
+
+
 def init_app(app, statsd_client=None, extra_filters: Sequence[logging.Filter] = tuple()):
     app.config.setdefault("NOTIFY_LOG_LEVEL", "INFO")
     app.config.setdefault("NOTIFY_APP_NAME", "none")
     app.config.setdefault("NOTIFY_LOG_PATH", "./log/application.log")
     app.config.setdefault("NOTIFY_RUNTIME_PLATFORM", None)
+    app.config.setdefault(
+        "NOTIFY_REQUEST_LOG_LEVEL",
+        "WARNING" if app.config["NOTIFY_RUNTIME_PLATFORM"] == "paas" else "NOTSET",
```
idavidmcdonald, line 43: I just want to check I have understood things correctly here. With the above defaults, when we release this code to our apps we expect to see the following changes

- On the PaaS, if an app serves a 5xx response then it will also issue a warning level log line corresponding
- On the PaaS, if an app services a 2xx response then it will not issue a log line (because log level is info and is less than warning)
- When running locally, if an app serves any response then it will log the the 'before request' log and the 'after request' log
- When runnning in ECS, if an app serves any response then it will log the the 'before request' log and the 'after request' log

If so, it seems reasonable that for ECS by default we will want lots of logging. It also seems reasonable for local development too.

I would be open to not having PaaS giving us extra warning logs for 5xx responses (we already get these from the gorouter) but happy to do so if you think it is useful.

tests/test_logging.py
```diff
@@ -94,3 +97,240 @@ def test_base_json_formatter_contains_service_id(tmpdir):
     service_id_filter = logging.ServiceIdFilter()
     assert json.loads(logging.BaseJSONFormatter().format(record))["message"] == "message to log"
     assert service_id_filter.filter(record).service_id == "no-service-id"
+
+
+@pytest.mark.parametrize(
+    "status_code,expected_after_level",
+    (
+        (200, builtin_logging.INFO),
+        (201, builtin_logging.INFO),
```
idavidmcdonald, line 106: Very optional: Could put in a 3xx if you want a bit more test coverage (but equally fine to not)

idavidmcdonald (commented):
Nice work Rob.

This looks pretty good, in particular on the test coverage which is very thorough.

Do have a think about my comments. I'm happy to approve it if you decide that you don't want to make any changes (although I think I'd like to see `duration_real` become `request_time`).

I haven't tested this out locally but I'm finishing for the day so wanted to leave my comments now so you have something to look at first thing on Monday.

I'll try and run it locally on Monday (but don't let that be a blocker to you).

notifications_utils/logging.py
```diff
@@ -17,11 +19,63 @@
...
+        )
+
+    @app.after_request
+    def after_request(response):
+        context = {
+            "status": response.status_code,
+            "duration_real": (
+                (time.perf_counter() - request.before_request_real_time)
+                if hasattr(request, "before_request_real_time")
+                else None
```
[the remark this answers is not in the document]
risicle: I wasn't able to 100% confidently say it couldn't happen if something got configured or used weirdly so I covered the possibility.

notifications_utils/logging.py
risicle: > When running locally, if an app serves any response then it will log the the 'before request' log and the 'after request' log

> When runnning in ECS, if an app serves any response then it will log the the 'before request' log and the 'after request' log

Both only if someone has set their `NOTIFY_LOG_LEVEL` to `DEBUG`

> I would be open to not having PaaS giving us extra warning logs for 5xx responses (we already get these from the gorouter) but happy to do so if you think it is useful.

Meh I suppose I could default `NOTIFY_REQUEST_LOG_LEVEL` to `CRITICAL` if we're on paas.

idavidmcdonald: Ah yes, I hadn't clocked the fact that this also depends on `NOTIFY_LOG_LEVEL` as the parent logger. Thanks, the above makes sense and helps clarify my understanding.

idavidmcdonald (approved):
Thanks Rob, looks good!

One tiny comment that you can take or leave.

One thing you'll need to do so the other apps can start using this is follow https://github.com/alphagov/notifications-utils/tree/main#publishing-a-new-version and update the version number.

tests/conftest.py
```diff
@@ -35,6 +42,20 @@ def celery_app(mocker):
     ctx.pop()
 
 
+@pytest.fixture
+def app_with_mocked_logger(mocker, tmpdir):
+    """Patch `create_logger` to return a mock logger that is made accessible on `app.logger`"""
+    mocker.patch(
+        "flask.sansio.app.create_logger",
```
leohemsted, line 49: oops flask needs to be above version 3.0 to mock this - i'll add a PR to fix this

risicle: Sorry missed this.. hmm yes there's a broader problem with dependency versions, libraries and tests. But that isn't a problem unique to our project.

PR #1078, opened 2023-12-06
Make package compatible with Python 3.11 and 3.12

quis:
This pull request means the package can be installed in a Python 3.11 or Python 3.12 environment.

See individual commits for detail.

### DIFF
--- a/tests/test_serialised_model.py
+++ b/tests/test_serialised_model.py
@@ -18,38 +18,47 @@ def test_cant_be_instatiated_with_abstract_properties():
     with pytest.raises(TypeError) as e:
         SerialisedModel()
 
-    if sys.version_info < (3, 9):
+    if sys.version_info >= (3, 12):
         assert str(e.value) == (
-            "Can't instantiate abstract class SerialisedModel with abstract methods ALLOWED_PROPERTIES"
+            "Can't instantiate abstract class SerialisedModel without an implementation "
+            "for abstract method 'ALLOWED_PROPERTIES'"
         )
     else:
-        assert "Can't instantiate abstract class SerialisedModel with abstract method ALLOWED_PROPERTIES"
+        assert str(e.value) == (
+            "Can't instantiate abstract class SerialisedModel with abstract method ALLOWED_PROPERTIES"
+        )
 
     with pytest.raises(TypeError) as e:
         Custom()
 
-    if sys.version_info < (3, 9):
-        assert str(e.value) == ("Can't instantiate abstract class Custom with abstract methods ALLOWED_PROPERTIES")
+    if sys.version_info >= (3, 12):
+        assert str(e.value) == (
+            "Can't instantiate abstract class Custom without an implementation "
+            "for abstract method 'ALLOWED_PROPERTIES'"
+        )
     else:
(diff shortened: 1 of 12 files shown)


tests/test_serialised_model.py
```diff
@@ -70,7 +79,12 @@ def foo(self):
     with pytest.raises(AttributeError) as e:
         assert Custom({"foo": "NOPE"}).foo == "bar"
```
leohemsted, line 80: this test confuses me

`Custom({})` fails because `foo` is required to be passed in because it's in `ALLOWED_PROPERTIES`.

`Custom({"foo": "NOPE"})` fails because there's no `foo.setter` attribute

i'm not sure this is testing any behaviour, other than "you can create a SerialisedModel that is fundamentally broken and will never work" 🤔

can we just delete the whole test? or rework it? Is there any unusual behaviour/interactions we might want to check around properties? I know 

(unrelatedly i only looked at it because on 3.10 the error is `can't set attribute "foo"` so doesn't match either case)

tests/test_template_types.py
```diff
@@ -31,54 +30,94 @@
...
             Template,
-            ("Can't instantiate abstract class Template with abstract methods __str__"),
-            marks=pytest.mark.skipif(sys.version_info >= (3, 9), reason="‘methods’ will be singular"),
+            (
+                "Can't instantiate abstract class Template with abstract method __str__",
+                (
+                    "Can't instantiate abstract class Template without an implementation "
+                    "for abstract method '__str__'"
+                ),
+            ),
```
leohemsted, line 43: i dont love this - i think this makes the test a bit messy and makes harder to spot and tidy up when we've completed our migration

we could just have an if statement in the code like we do in the SerialisedModel tests

```
def test_abstract_classes_cant_be_instantiated(template_class, expected_3_10_error, expected_3_9_error):
    with pytest.raises(TypeError) as error:
        template_class({})
    if sys.version_info < (3, 10):
        assert str(error.value) == expected_3_9_error
    else:
        assert str(error.value) == expected_3_10_error

Comment (leohemsted) [reply tests/test_template_types.py:43]: or just super simple:

```
assert f"Can't instantiate abstract class {template_class}" in str(error.value) 
assert f"__str__" in str(error.value) 
```

(this is more in line with our guidance we're publishing to users about error messages anyway)

File: tests/test_serialised_model.py (hunk 1 above)
Comment (quis) [reply tests/test_serialised_model.py]: Tried to explain and clean up in 1c14d165e6b102dc827e08fd4ff18837822510f9

File: tests/test_template_types.py [hunk 3]
```diff
@@ -31,54 +30,23 @@
...
-            marks=pytest.mark.skipif(sys.version_info < (3, 9), reason="‘method’ will be pluralised"),
-        ),
+        Template,
+        Template,
+        BaseEmailTemplate,
+        BaseEmailTemplate,
+        BaseLetterTemplate,
+        BaseLetterTemplate,
+        BaseBroadcastTemplate,
+        BaseBroadcastTemplate,
```
leohemsted, line 42: can dedupe this

```suggestion
        Template,
        BaseEmailTemplate,
        BaseLetterTemplate,
        BaseBroadcastTemplate,
```

leohemsted (approved):
looks good, thanks for tweaking those tests (just a tiny cleanup)

PR #1079, opened 2023-12-06
Remove Emergency Alerts

quis:
We are removing emergency alerts since it’s been [split into its own codebase](https://github.com/alphagov/emergency-alerts-utils).

This pull request removes all of the `polygons` module, `template.BroadcastPreviewTemplate` and `template.BroadcastMessageTemplate`, and their dependencies.

*** 

Should not be merged before:
- [x] https://github.com/alphagov/notifications-admin/pull/4922
- [x] https://github.com/alphagov/notifications-api/pull/3970


leohemsted (commented):
I approve of this enthusiastically, but we should remove all usages of `polygons`, `BroadcastPreviewTemplate` and `BroadcastMessageTemplate` from admin/api before we merge this, or we run the risk of not being able to upgrade to some new version `74.0.1` which we may want to do without having the lengthy unpicking of all the templates and things

leohemsted (approved):

PR #1081, opened 2023-12-07
Allow hiding notify tag on letter print template

samuelhwilliams:
This tag should only be present on the first page of a PDF we generate to send to DVLA.

In supporting bilingual letters, we generate two PDFs: a Welsh variant followed by an English variant.

We need to stop inserting the 'NOTIFY' tag on the English variant, and so need to be able to toggle the tag off, otherwise DVLA's printers get confused.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 73.2.0
+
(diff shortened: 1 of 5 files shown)


CrystalPea (approved):
Looks good.

PR #1082, opened 2023-12-07
Fix utils packaging

samuelhwilliams:
In an attempt to exclude the `tests` package, we excluded a lot of
notifications_utils.

```
In [1]: from setuptools import find_packages

In [2]: find_packages()
Out[2]:
['tests',
 'notifications_utils',
 'notifications_utils.clients',
 'notifications_utils.testing',
 'notifications_utils.countries',
 'notifications_utils.clients.encryption',
 'notifications_utils.clients.redis',
 'notifications_utils.clients.antivirus',
 'notifications_utils.clients.statsd',
 'notifications_utils.clients.zendesk']
```

We should just exclude `tests`, rather than including the (top-level)
`notifications_utils`. The latter excludes all of the subpackages, which
then makes them inaccessible at their dot paths.

---

Was leading to errors like this when bumping utils on template-preview:

```
./scripts/run_with_docker.sh make test
Running custom command
ruff check .
black --check .
All done! ✨ 🍰 ✨
31 files would be left unchanged.
pytest -n auto --maxfail=10
ImportError while loading conftest '/home/vcap/app/tests/conftest.py'.
tests/conftest.py:8: in <module>
    from app import create_app
app/__init__.py:10: in <module>
    from notifications_utils.clients.encryption.encryption_client import Encryption
E   ModuleNotFoundError: No module named 'notifications_utils.clients'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 73.2.1
+
+* Fix utils packaging to allow access to subpackages again.
+
 ## 73.2.0
 
 * Adds a `include_notify_tag` parameter to `LetterPrintTemplate` so that bilingual letters can disable the NOTIFY tag on the English pages of a letter.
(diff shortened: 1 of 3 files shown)


quis (approved):
Oopsie

PR #1084, opened 2023-12-29
logging: add microseconds to time format via mixin class

risicle:
https://trello.com/c/pWDzU11r/567-fractional-timestamps-for-app-logs

Ideally we'd just include this in our `TIME_FORMAT`, but python's `strftime` has no ability to include even milliseconds. The python-default logging `datefmt` is augmented with milliseconds in the stdlib logging module via a hack, but this hack is disabled when a custom `datefmt` is supplied. So we have to add back our own hack to include precise time.

Use microsecond granularity as it should be enough to correctly order even log records that are generated immediately-consecutively in the same thread.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 74.2.0
+
+* Change logging's date formatting to include microseconds
+
 ## 74.1.0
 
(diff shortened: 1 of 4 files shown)


idavidmcdonald (approved):

risicle:
Rebased to adjust version number bump

PR #1085, opened 2023-12-29
A couple of QoL changes

leohemsted:
* reuse requests sessions (for small perf boosts)
* remove unused subdependency geojson
* fix the version bump script in case you ever ran `pip install -e .` or similar in your venv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 74.1.0
+
+* remove the geojson dependency (another emergency alerts module)
+* reuse a single session for all antivirus/zendesk requests to get http keepalive advantages
+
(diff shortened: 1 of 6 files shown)


risicle (approved):
Nice stuff

sakisv:
> - reuse requests sessions (for small perf boosts)

How small? 👀 
(All the repo subscriptions were transferred to my personal email, so I'm just unsubscribing one PR-at-a-time 🙈 )

leohemsted:
@sakisv it reuses the http connection for up to 30 secs, so skips the tls renegotiation. for things like calling in a loop, big wins. for antivirus, where itll be used by different eventlets across different admin workers, probably increasingly negligible, but it's a good habit to get into

PR #1095, opened 2024-02-05
logging: also attach handlers etc. to `celery.worker` and `celery.redirected`

risicle:
https://trello.com/c/q5ehwcty/628-try-and-get-full-celery-logs-for-the-paas-into-logit

We're currently missing useful logs unless we keep all stderr from celery workers, but:

 - we don't on paas
 - it's horrible and unstructured, so it would be better if we didn't on ecs either

Adding `celery.redirected` should include messages that have been crudely structured from raw stdout/stderr messages, which we'll need because some important messages e.g. from kombu are sent through raw `print` calls.

I have tested this (minimally - it's hard to convince it to emit the right log messages from within celery) with the `notify-delivery-worker-sender`.

### DIFF
--- a/notifications_utils/logging.py
+++ b/notifications_utils/logging.py
@@ -89,7 +89,12 @@ def init_app(app, statsd_client=None, extra_filters: Sequence[logging.Filter] = 
 
     handlers = get_handlers(app, extra_filters=extra_filters)
     loglevel = logging.getLevelName(app.config["NOTIFY_LOG_LEVEL"])
-    loggers = [app.logger, logging.getLogger("utils")]
+    loggers = [
+        app.logger,
+        logging.getLogger("utils"),
+        logging.getLogger("celery.worker"),
+        logging.getLogger("celery.redirected"),  # stdout/stderr
+    ]
     for logger_instance, handler in product(loggers, handlers):
         logger_instance.addHandler(handler)
         logger_instance.setLevel(loglevel)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 74.9.0
+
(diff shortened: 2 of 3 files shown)


notifications_utils/logging.py
```diff
@@ -89,7 +89,12 @@ def after_request(response):
 
     handlers = get_handlers(app, extra_filters=extra_filters)
     loglevel = logging.getLevelName(app.config["NOTIFY_LOG_LEVEL"])
-    loggers = [app.logger, logging.getLogger("utils")]
+    loggers = [
+        app.logger,
+        logging.getLogger("utils"),
+        logging.getLogger("celery.worker"),
+        logging.getLogger("celery.redirected"),  # stdout/stderr
```
idavidmcdonald, line 96: How did you find/pick these 2 logger names?

I can see `celery.redirected` being referenced in the celery code (https://github.com/celery/celery/blob/v5.2.7/celery/app/log.py#L80)

I can also see `celery` and `celery.task` being referenced in that same file for logger names but can't see any `celery.worker` anywhere?

risicle: https://github.com/celery/celery/blob/32a285dd956756322e8a9c9310731e8419a3660b/celery/utils/log.py#L106

risicle: FWIW longer-term I think we should consider consuming the `setup_loggers` signal, preventing https://github.com/celery/celery/blob/32a285dd956756322e8a9c9310731e8419a3660b/celery/app/log.py#L102 from running and allowing us to set up the loggers ourselves without interference. But that would need more fiddling & testing.

idavidmcdonald (approved):
Let's give it a go

risicle:
Sorry need re-review now :cry:

risicle:
Oh you beat me thanks.

PR #1096, opened 2024-02-06
Fix barcodes bilingual letters

whpearson:
Always hide barcodes when printing a letter

We used to only hide them on a page with Notify tag, but that
lead to a problem with bilingual letters when on first page of
English version of the letter barcodes were visible and were causing
problems for DVLA.


tombye (approved):
Looks good to me 👍🏻

PR #1098, opened 2024-02-08
Add option to ignore list of web paths from request logging

corlettb:
What
----

- Add option to change set of web paths to debug request logging for successful calls. Defaults to /_status and /metrics.
- Add new fields to log messages:
  - user_agent
  - host
  - path

Why
----

- Our logs get too many successful status and metrics calls. This will remove them from our logs.
- Additional fields will help produce more useful logit dashboards.

### DIFF
--- a/notifications_utils/logging.py
+++ b/notifications_utils/logging.py
@@ -25,6 +25,9 @@ def _common_request_extra_log_context():
         "url": request.url,
         "endpoint": request.endpoint,
         "remote_addr": request.remote_addr,
+        "user_agent": request.user_agent.string,
+        "host": request.host.split(":", 1)[0],
+        "path": request.path,
         "parent_span_id": getattr(request, "parent_span_id", None),
         # pid and is available on LogRecord by default, as `process` but I don't see
         # a straightforward way of selectively including it only in certain log messages -
@@ -40,6 +43,7 @@ def init_app(app, statsd_client=None, extra_filters: Sequence[logging.Filter] = 
     app.config.setdefault("NOTIFY_APP_NAME", "none")
     app.config.setdefault("NOTIFY_LOG_PATH", "./log/application.log")
     app.config.setdefault("NOTIFY_RUNTIME_PLATFORM", None)
+    app.config.setdefault("NOTIFY_LOG_DEBUG_PATH_LIST", {"/_status", "/metrics"})
     app.config.setdefault(
         "NOTIFY_REQUEST_LOG_LEVEL",
         "CRITICAL" if app.config["NOTIFY_RUNTIME_PLATFORM"] == "paas" else "NOTSET",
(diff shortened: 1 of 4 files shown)


notifications_utils/logging.py
```diff
@@ -40,43 +43,49 @@ def init_app(app, statsd_client=None, extra_filters: Sequence[logging.Filter] =
     app.config.setdefault("NOTIFY_APP_NAME", "none")
     app.config.setdefault("NOTIFY_LOG_PATH", "./log/application.log")
     app.config.setdefault("NOTIFY_RUNTIME_PLATFORM", None)
+    app.config.setdefault("NOTIFY_LOG_IGNORE_PATH_LIST", "/_status;/metrics")
```
risicle, line 46: Suggest making this a list - I don't think we have anything that automatically reads flask config vars from env vars and if/where we added that we could do the list conversion.

This is the common pattern there seems to be for complex value parsing https://github.com/alphagov/notifications-api/blob/f5ff1aa6de14876cd8ed34b52caf552c1fa90142/app/config.py#L89

risicle: Actually if you make it a `set` here you can avoid having to put together your dict down below

notifications_utils/logging.py
```diff
@@ -40,43 +43,49 @@ def init_app(app, statsd_client=None, extra_filters: Sequence[logging.Filter] =
...
-        )
+        if request.path not in log_ignore_path_dict:
+            # emit an early log message to record that the request was received by the app
+            context = _common_request_extra_log_context()
+            current_app.logger.getChild("request").log(
+                logging.DEBUG,
+                "Received request %(method)s %(url)s",
+                context,
+                extra=context,
+            )
```
risicle, line 68: I would leave these because they're at `DEBUG` level anyway.

notifications_utils/logging.py
corlettb: Updated to a set.

notifications_utils/logging.py
corlettb: I've left the messages as DEBUG and I've left the post request message as DEBUG.

tests/test_logging.py
```diff
@@ -280,6 +301,9 @@ def some_route():
                 "url": "http://localhost/",
                 "method": "GET",
                 "endpoint": "some_route",
+                "host": "localhost",
+                "path": "/",
+                "user_agent": "Werkzeug/3.0.1",
```
risicle, line 306: This will "break" every time we update bump werkzeug won't it? Suggest either using `mock.ANY` or even e.g. `AnyStringMatching("Werkzeug.*")`

corlettb: Good point. Updated with your suggestion.

risicle (approved):
Not exactly how I would have written the test but ok

PR #1102, opened 2024-02-16
return structured time_taken field in celery log lines

leohemsted:


c4seyla (approved):

notifications_utils/celery.py
```diff
@@ -44,7 +50,15 @@ def on_success(self, retval, task_id, args, kwargs):
         def on_failure(self, exc, task_id, args, kwargs, einfo):
             # enables request id tracing for these logs
             with self.app_context():
-                app.logger.exception("Celery task %s (queue: %s) failed", self.name, self.queue_name)
+                elapsed_time = time.monotonic() - self.start
+
+                app.logger.exception(
+                    "Celery task %s (queue: %s) failed after %s",
```
risicle, line 56: `AssertionError: assert 'Celery task 3dd53545-d348-4328-ad21-15501a5e8063 (queue: test-queue) failed after 5.000' in ['Celery task 3dd53545-d348-4328-ad21-15501a5e8063 (queue: test-queue) failed after 5.0']` probably due to this.

risicle: BTW according to docs, `logging.exception` "should only be called from an exception handler". I assume we know that it is?

leohemsted: i'm gonna change it to `.4f` to be consistent with the success log

leohemsted: done

risicle (approved):

PR #1106, opened 2024-03-08
Rename encryption_client to signing_client

corlettb:
What
----

Rename encryption_client to signing_client.

Why
----

The original python Class was not doing any encryption. The strings passed in could be decoded with base64. They were just signed to prevent tampering. Renaming this class to avoid confusion.

Warning
----

This is a breaking change. Downstream projects will need to update their references to the encryption class.


klssmith:
I think the changes look fine, the only thing I'd suggest is that the version is changed to `75.0.0` due to the breaking change

corlettb:
> I think the changes look fine, the only thing I'd suggest is that the version is changed to `75.0.0` due to the breaking change

Updated to a major version.

klssmith (approved):

PR #1111, opened 2024-04-12
Add `InsensitiveSet`

quis:
This behaves like a regular `set` except it determines uniqueness based on the normalised value of the members.


whpearson (approved):
Looks good

PR #1117, opened 2024-05-08
Enable pyupgrade and other linting rules

quis:
[Pyupgrade](https://pypi.org/project/pyupgrade/) automatically upgrades syntax for newer versions of Python. Newer syntax is often friendlier and less verbose – for example f-string formatting rather than calls to `str.format`.

It has a total of [42 rules](https://docs.astral.sh/ruff/rules/#pyupgrade-up).

This pull request adds it to our linter config, and then fixes all the instances of old syntax it found (the vast majority of these were automatic by running Ruff with the `--fix` flag).

***

This pull request It then does the same for these other optional rules:
- [`flake8-comprehensions`](https://pypi.org/project/flake8-comprehensions) – removes unnecessary calls  to `list`/`set`/`dict` when writing comprehensions 
- [`flake8-implicit-str-concat`](https://pypi.org/project/flake8-implicit-str-concat) – turns `"a" "b"` into `"ab"`
- [`flake8-raise`](https://pypi.org/project/flake8-raise) – automatically adds `from e` when `raise`ing, where appropriate
- [`flake8-pie`](https://pypi.org/project/flake8-pie/) – ‘misc lints’


notifications_utils/markdown.py
```diff
@@ -23,45 +23,37 @@
...
     r"|"
     r"\s*$)"
-    % (
-        mistune._pure_pattern(mistune.BlockGrammar.def_links),
-        mistune._pure_pattern(mistune.BlockGrammar.def_footnotes),
-    )
 )
 mistune.BlockGrammar.list_item = re.compile(
-    r"^(( *)(?:[•*-]|\d+\.)[^\n]*" r"(?:\n(?!\2(?:[•*-]|\d+\.))[^\n]*)*)", flags=re.M
+    r"^(( *)(?:[•*-]|\d+\.)[^\n]*(?:\n(?!\2(?:[•*-]|\d+\.))[^\n]*)*)", flags=re.M
```
leohemsted, line 35: i dont think we need to change anything now but i wonder if this was originally a multiline string that got one-lined, and now smooshed together, and if we think it's more readable over multiple lines?

```py
mistune.BlockGrammar.list_item = re.compile(
  (
    r"^(( *)(?:[•*-]|\d+\.)[^\n]*"
    r"(?:\n(?!\2(?:[•*-]|\d+\.))[^\n]*)*)"
  ),
  flags=re.M
)
```

tests/test_field.py
```diff
@@ -216,7 +216,7 @@ def test_handling_of_missing_values(content, values, expected):
         99.99999,
         "off",
         "exclude",
-        "no" "any random string",
+        "noany random string",
```
leohemsted, line 219:
```suggestion
        "no",
        "any random string",
```

leohemsted (approved):
i might suggest we add the commits hashes to `.git-blame-ignore-revs` to make it easier to find meaningful code changes

notifications_utils/markdown.py
quis: Black re-smooshes it if you try to do this 😞

notifications_utils/clients/redis/redis_client.py
```diff
@@ -2,13 +2,16 @@
...
+# (`Type` is deprecated in favour of `type` but we need to match the
+# signature of the method we are stubbing)
+from typing import Type  # noqa: UP035
 
 from flask import current_app
 from flask_redis import FlaskRedis
 
 # expose redis exceptions so that they can be caught
-from redis.exceptions import RedisError  # noqa
+from redis.exceptions import RedisError  # noqa: F401
```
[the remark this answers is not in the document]
quis: Done in [ed82ae9](https://github.com/alphagov/notifications-utils/pull/1117/commits/ed82ae989439647d06ccf0d099b5597addd9987a)

quis:
Added commits to `.git-blame-ignore-revs` in [27311ac](https://github.com/alphagov/notifications-utils/pull/1117/commits/27311ac7d952fb067dc93a0eb279462ebb4c9ef4)

leohemsted (approved):

PR #1118, opened 2024-05-09
Add a command to copy `pyproject.toml` into apps

quis:
This will let us share linter config between the apps. In the future it might even let us share common dependencies between the apps, since `pyproject.toml` can do dependency management.

This will require adding a `make` command to each app like:
```bash
python -c "from notifications_utils.version_tools import copy_pyproject_toml; copy_pyproject_toml()"
```

It also updates a couple of bits of config in `pyproject.toml` so that we’ll have parity when we drop this into the admin app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 77.1.0
+
+* Add `version_tools.copy_pyproject_toml` to share linter config between apps
+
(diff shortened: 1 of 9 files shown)


whpearson:
Changes look good, I'll approve once the merge conflicts have been fixed.

whpearson (approved):
Shared config would be good. Changes makes sense.

PR #1120, opened 2024-05-22
Ensure no stale data in redis if redis errors when deleting cache key

idavidmcdonald:
For context, when the admin app wants to get data from our API
it will first try and get the data from redis and if the data
doesn't exist in redis, it will get the data from the API and then
finish by setting that data in redis. When the admin app wants to
update data for our API, it currently calls out to the API to
update the data and then deletes any existing/relevant data cached
in redis. The subsequent time it tries to get this data that it just
updated, it will use the usual approach for getting data that will
set the redis key too.

There is a problem with when the admin app wants to update something
in the database. At the moment, it will start by calling the API
to update the database and if that is successful it will then attempt
to delete the relevant cache keys in redis. But redis may not always
be available and in that case, it will
- fail to delete the cache key
- catch and ignore the exception raised when trying to delete the
  redis cache key
This leads to the change the user requested being made in the
database, but the cache still has the old data in it! This is bad
because our apps check the cache first and this can result in us
sending out incorrect emails using old templates, for example. We
have a runbook to manually recover from this position if redis
has downtime and delete queries fail during this time:
https://github.com/alphagov/notifications-manuals/wiki/Support-Runbook#deal-with-redis-outages

Note, there is no issue with `cache.set` calls. If the call to redis
fails as part of this, no stale data is produced. The database hasn't
been changed and either redis hasn't been changed or hasn't had
the current data added to it.

This commit changes our caching logic so that we can't end up with
stale data in redis if a delete to redis fails. If a delete to redis
fails, then we error early and don't attempt to update the database.
The trade off we make here is that now a user will see an error
page if their request failed to delete from redis, whereas before
they would have gotten a 200 (but ended up with stale data). We
think it is worse to have stale data, then it is to fail a
users request.

## Replicating locally
If you want to see the behaviour before and after you can replicate redis having an error deleting a key by inserting `raise Exception` above this line:
https://github.com/alphagov/notifications-utils/blob/main/notifications_utils/clients/redis/redis_client.py#L177. If running notify using notifications-local, you can install your local utils by 

```
docker exec -it notify-admin bash
pip install -e ../utils
```

And then inspect your Redis using:
```
docker exec -it redis bash
redis-cli
```

And inspect your Postgres using
```
psql postgresql://notify:notify@localhost:5433/notification_api
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 77.2.1
+
+* Change redis delete behaviour to error, rather than end up with stale data, if Redis is unavailable.
+
 ## 77.2.0
 
 * `NotifyTask`: include pid and other structured fields in completion log messages
--- a/notifications_utils/clients/redis/request_cache.py
+++ b/notifications_utils/clients/redis/request_cache.py
@@ -68,13 +68,23 @@ class RequestCache:
         def _delete(client_method):
             @wraps(client_method)
             def new_client_method(*args, **kwargs):
+                redis_key = self._make_key(key_format, client_method, args, kwargs)
+
                 # It is important to attempt the redis deletion first and raise an exception
                 # if it is unsuccessful. If we didn't, then we risk having a successful API
                 # call that updates the database, but redis left with stale data. Stale data
                 # is worse then failing the users requests
-                redis_key = self._make_key(key_format, client_method, args, kwargs)
                 self.redis_client.delete(redis_key, raise_exception=True)
-                return client_method(*args, **kwargs)
+
+                api_response = client_method(*args, **kwargs)
+
+                # We also attempt another redis deletion after the API. This is to deal with
+                # the race condition where another request repopulates redis with the old data
+                # before the database has been updated. We want to raise an exception if the call
+                # to redis here fails because that will hopefully prompt the user that something
+                # went wrong and they should retry their action (hopefully resolving the problem).
+                self.redis_client.delete(redis_key, raise_exception=True)
+                return api_response
 
             return new_client_method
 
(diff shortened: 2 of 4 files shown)


risicle:
If apps are deleting redis keys _before_ they update the database, there's a race condition where another request re-populates the redis key with the old value before the database has been updated.

idavidmcdonald:
Oo very good catch Rob! 

So I can think of one option that will help this problem, but can't think of a way to get a perfect solution.

With my current code, you point out the following problem:

1. Request A deletes cache key
2. Request B sets cache key with the original data
3. Request A updates our database with the new data
4. We end with stale data in Redis

Proposed solution:

1. Request A deletes cache key
2. Request B sets cache key with the original data
3. Request A updates our database with the new data
4. Request A deletes the cache key again
5. We end with no stale data in Redis

For this option, we could have the old data in redis up until step 4... but that was what we had with the current version of utils (the cache holds stale data until the API call has completed). 

For this option, if redis went down and then someone tried to update a resource, then we would error immediately and redis would remain unchanged and our database would remain unchanged. No stale data - good.

There is an edge case though. If redis goes down after step 2. Then we will have stale data in the cache after step 2, but step 4 won't be able to clear the cache because Redis is down. This is exactly the issue we have in production at the moment! But at the moment, we will see this problem for any deletes that are attempted whilst Redis is down, whereas with the proposed solution, we will only see stale data in redis if redis goes down between steps 2 and 4. This should much less frequent and so whilst it doesn't completely stop us having stale data in redis, it should massively reduce the amount of stale data in redis if redis had a wobble. 

Thoughts?

If we went for this approach, we would also have to decide if the second delete attempt would raise an uncaught exception or not (I would suggest it does so the user knows something is up and they will likely try their action again rather than assuming it was successful).

The other thing I did some googling of is locks in Redis but didn't find a way to say "don't let anyone update this key until something else has happened" - plus adding locks adds potentially more complexity than it is worth.

risicle:
If we had a simpler architecture I would suggest we take a `FOR UPDATE` lock on the db row before we start trying to delete redis keys, but the db is in a separate app :upside_down_face:

risicle:
But yes, that appears a better solution while not perfect.

Because of this sort of issue, some people suggest pushing cache population/dirtying into the app with the database.

idavidmcdonald:
I've added a new commit on here that deletes the cache a second time (after the API update) just to remove the risk of the race condition. See commit message for more details.

Doing some more thinking on this, maybe the long term direction we could aim for is the API is responsible for caching (or at least deletion of the cache). But the above change should mean we don't end up with stale data in redis unless we have both a race condition and redis going down at a specific time. Given that should be very rare, it probably isn't worth the effort of doing anything drastic at the moment and instead knowing we have a 'probably good enough' solution with this tweak here.

leohemsted (approved):
looks good but you'll need to resolve conflicts and nudge me again

leohemsted (approved):

PR #1121, opened 2024-05-31
Update Zendesk client for new form

klssmith:
We are switching to use a new Zendesk form, so this changes the Zendesk
client. We now need to pass in the ID of the new form, and instead of
populating the "Ticket category" dropdown we want to populate a dropdown
called "Task type". This will only ever have one thing selected, so
doesn't need to be able to take a list of values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 77.0.0
+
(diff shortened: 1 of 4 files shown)


whpearson (approved):
Changes make sense and have verified katie tested the ids are correct.

PR #1123, opened 2024-06-10
Bump subdeps

leohemsted:
Many libraries we're only using much more recent versions than utils has pinned anyway (like flask or pypdf), so it's easy to bring them up to the oldest version in use without any risk.

Others have outstanding CVEs (like jinja) so it's good to make sure we always have secure versions.

Others like boto3 we've already assessed the changelogs elsewhere[^1]

Many major changes were just dropping support for old versions of python

Some like cachetools needed a bit of checking as they changed import structure (but we're not affected in this instance)

[^1]: https://github.com/alphagov/notifications-api/pull/4082


quis (approved):
Good to go once rebased

setup.py
```diff
@@ -23,24 +23,24 @@
...
-        "pytz>=2020.4",
+        "pytz>=2024.1",
         "smartypants>=2.0.1",
-        "pypdf>=3.9.0",
-        "itsdangerous>=1.1.0",
-        "govuk-bank-holidays>=0.10,<1.0",
-        "boto3>=1.19.4",
+        "pypdf>=3.13.0",
+        "itsdangerous>=2.1.2",
+        "govuk-bank-holidays>=0.14,<1.0",
```
quis, line 42: Optional, but I don’t think we should restrict the major version here unless there’s a known problem with `1.x` (in this case `0.14` is the highest version and there is no `1.x branch)
```suggestion
        "govuk-bank-holidays>=0.14",
```

leohemsted: Done

quis (approved):

PR #1124, opened 2024-06-11
Restrict postcodes to valid UK postcode zones

quis:
The postcode zone is the first 1, 2 or 3 letters of the postcode. There are about 120 postcode zones defined in the UK.

At the moment we allow postcodes to start with any 1 or 2 letters. This means we allow plenty of postcode which aren’t real, for example `NF1 1AA`.

By restricting the first n letters of the postcode to only valid postcode zones we can reduce the number of letters which DVLA manually flag to us as having invalid addresses.

***

List taken from: http://www.ons.gov.uk/ons/guide-method/geography/products/postcode-directories/-nspp-/onspd-user-guide-and-version-notes.zip


leohemsted (approved):
Looks good. Any idea how often this data is updated? I suspect pretty infrequently but good to be aware of any changes

PR #1125, opened 2024-06-12
Invalid phone error refactor

leohemsted:
https://trello.com/c/I67noSbf/10-update-phone-number-validation-to-optionally-allow-uk-landlines

More refactoring of `InvalidPhoneError` ahead of adding UK landlines.

This one's quite a big one - I'm taking an opinionated stance that rather than writing code like `raise InvalidPhoneError('some error message')`, it's nicer to write `raise InvalidPhoneError(code=InvalidPhoneError.Codes.SOME_ERROR)`. Then, the exception class itself has a list of all the error codes, and a list of all the human-readable messages for surfacing on the front-end etc.

This means we can define these messages once, and only need to surface them at exit points (API error json serialisation, form rendering) and can just use the codes if we need to branch.

We're not doing work on the API v2 error slugs just yet but it's useful to have a starting point for that work I think.

Best viewed commit by commit.

Open to discussion about class names, file locations, if the error codes should be nested or not, if there's better ways to represent than an enum of codes and a map of said enums to strings, etc.

### DIFF
--- a/notifications_utils/recipient_validation/errors.py
+++ b/notifications_utils/recipient_validation/errors.py
@@ -33,6 +33,14 @@ class InvalidPhoneError(InvalidRecipientError):
         Codes.UNSUPPORTED_COUNTRY_CODE: "Country code not found - double check the mobile number you entered",
     }
 
+    LEGACY_V2_API_ERROR_MESSAGES = ERROR_MESSAGES | {
+        Codes.TOO_LONG: "Too many digits",
+        Codes.TOO_SHORT: "Not enough digits",
+        Codes.NOT_A_UK_MOBILE: "Not a UK mobile number",
+        Codes.UNKNOWN_CHARACTER: "Must not contain letters or symbols",
+        Codes.UNSUPPORTED_COUNTRY_CODE: "Not a valid country prefix",
+    }
+
     def __init__(self, *, code: Codes = Codes.INVALID_NUMBER):
         """
         Create an InvalidPhoneError. The code must be present in InvalidPhoneError.ERROR_MESSAGES or this will raise a
@@ -40,6 +48,9 @@ class InvalidPhoneError(InvalidRecipientError):
         """
         self.code = code
         super().__init__(message=self.ERROR_MESSAGES[code])
+
+    def get_legacy_v2_api_error_message(self):
+        return self.LEGACY_V2_API_ERROR_MESSAGES[self.code]
 
 
 class InvalidAddressError(InvalidRecipientError):
--- a/notifications_utils/recipient_validation/errors.py
+++ b/notifications_utils/recipient_validation/errors.py
@@ -1,7 +1,10 @@
+from enum import StrEnum, auto
+
+
 class InvalidRecipientError(Exception):
     message = "Not a valid recipient address"
 
-    def __init__(self, message=None):
+    def __init__(self, message: str = None):
         super().__init__(message or self.message)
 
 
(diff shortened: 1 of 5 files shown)


notifications_utils/recipient_validation/errors.py
```diff
@@ -10,7 +18,42 @@ class InvalidEmailError(InvalidRecipientError):
 
 
 class InvalidPhoneError(InvalidRecipientError):
-    message = "Not a valid phone number"
+    class Codes(StrEnum):
+        DEFAULT = auto()
```
leohemsted, line 22: `default` as a code doesn't really make sense in isolation, so should change this to something like `invalid_number` and then just use that as the default in the constructor

```suggestion
        INVALID_NUMBER = auto()
```

notifications_utils/recipient_validation/errors.py
```diff
@@ -10,7 +18,42 @@ class InvalidEmailError(InvalidRecipientError):
...
+        Codes.TOO_LONG: "Too many digits",
+        Codes.TOO_SHORT: "Not enough digits",
+        Codes.NOT_A_UK_MOBILE: "Not a UK mobile number",
+        Codes.UNKNOWN_CHARACTER: "Must not contain letters or symbols",
+        Codes.UNSUPPORTED_COUNTRY_CODE: "Not a valid country prefix",
+    }
+
+    def __init__(self, *, code: Codes = Codes.DEFAULT):
+        """
+        Create an InvalidPhoneError. The code must be present in InvalidPhoneError.ERROR_MESSAGES or this will
```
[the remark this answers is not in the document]
leohemsted: Good spot! fixed

tombye:
I don't really use Python enough that I can comment about class names, file locations or the use of enums but this all makes sense to me as an approach. I can also see how I might use it in code I'm writing at the moment that would have to handle these errors. One very small niggle is that the commit message for https://github.com/alphagov/notifications-utils/pull/1125/commits/c2f049a34a9bc63d95957c41b68aee2a536ff3ab has a point that says
> Define all error messages in `InvalidPhoneError.ERROR_MESSAGES`. This
  should be the source of truth. If a specific place needs

Can the end of this be added please? I'm interested to know the idea for how to handle specific needs here.

notifications_utils/recipient_validation/errors.py
```diff
@@ -1,7 +1,15 @@
...
     message = "Not a valid recipient address"
+    code = "UNKNOWN"
 
-    def __init__(self, message=None):
+    def __init__(self, message: str = None, code: str = None):
+        if code:
+            self.code = code
+            if not message:
+                message = code
         super().__init__(message or self.message)
```
quis, line 13: I had a hard time reasoning about this conditional.

I think this is equivalent (and makes the precedence clearer) but I still can’t see where `"UNKNOWN"` or `self.code` is ever used – do we need those lines?
```suggestion
        message = message or code or self.message
        self.code = code or "UNKNOWN"
        super().__init__(message)
```

leohemsted: that's fair about the conditional.

i guess i was wondering about making this the same across all the three email/phone/address validation errors but that didn't happen - i can probs move `self.code` to just be in `InvalidPhoneError` for now. it's hard to imagine us handling `InvalidRecipientError` where we both:

* care about specific error codes (eg too long, too short, invalid country code) more than just "what is the error message to display"
* don't know whether we're validating an sms, email, or letter

i'll update the base class to remove `self.code` completely and just keep the default message there for now

leohemsted: Done (squashed into https://github.com/alphagov/notifications-utils/pull/1125/commits/a0ebe1a9ee08b97b23315acb082f4c8e293bc5f1)

CrystalPea (approved):

quis (approved):

PR #1126, opened 2024-06-19
Phone number validation

leohemsted:
https://trello.com/c/I67noSbf/10-update-phone-number-validation-to-optionally-allow-uk-landlines

🚨 This PR is targeting invalid-phone-error-refactor for a simpler diff 🚨 

Adds a new function for UK landline validation. Instead of hand-writing more code, instead we can lean on google's [`libphonenumber`](https://github.com/google/libphonenumber), which we already have as a dependency anyway.

we already use phonenumbers in `format_phone_number_human_readable` as it nicely formats numbers. It has a whole suite of validation features too.

We often find other teams asking us how we validate phone numbers so they can replicate it on their own frontend. We link to this repo but it would be much nicer if we could just point them to a well supported third party library with a variety of implementations in different languages.

We could potentially run validation in parallel with our regular output to ensure it matches. I'm not too worried if things we currently fail are allowed in by phonenumbers, as the chances are they're more correct than we are - but it's important that any numbers we currently accept and deliver to also work in phonenumbers.

-------

TODO: 
- [ ] Merge #1125 
- [x] Investigate: which of the other functions in the file do we need to replicate?
  - [ ] ~normalise_phone_number~ - used in sms sender normalisation (so often dealing with things that aren't phone numbers, or other valid short numbers eg `119`)
  - [x] is_uk_phone_number
  - [x] get_international_phone_info
  - [x] _is_a_crown_dependency_number
  - [ ] ~get_international_prefix~ - internal utils use only, don't need to replicate
  - [ ] ~get_billable_units_for_prefix~ - internal utils use only, don't need to replicate
  - [x] use_numeric_sender
  - [ ] ~validate_uk_phone_number~ - internal utils use only, don't need to replicate
  - [x] validate_phone_number
  - [ ] ~try_validate_and_format_phone_number~ - not needed as this is used for inbound sms, sms reply to, search params (none of which are uk landline specific)
  - [x] format_phone_number_human_readable

### DIFF
--- a/notifications_utils/recipient_validation/errors.py
+++ b/notifications_utils/recipient_validation/errors.py
@@ -28,7 +28,7 @@ class InvalidPhoneError(InvalidRecipientError):
         # this catches numbers with the right length but wrong digits
         # for example UK numbers cannot start "06" as that hasn't been assigned to a purpose by ofcom,
         # or a 9 digit UK number that does not start "01" or "0800".
-        Codes.INVALID_NUMBER: "TODO: CONTENT! Number is not valid – double check the mobile number you entered",
+        Codes.INVALID_NUMBER: "Number is not valid – double check the phone number you entered",  # TODO: CONTENT!
         Codes.TOO_LONG: "Mobile number is too long",
         Codes.TOO_SHORT: "Mobile number is too short",
         Codes.NOT_A_UK_MOBILE: (
@@ -55,7 +55,7 @@ class InvalidPhoneError(InvalidRecipientError):
         super().__init__(message=self.ERROR_MESSAGES[code])
 
     @classmethod
-    def from_phonenumbers_validationresult(cls, reason: phonenumbers.ValidationResult) -> str:
+    def from_phonenumbers_validation_result(cls, reason: phonenumbers.ValidationResult) -> str:
         match reason:
             case phonenumbers.ValidationResult.TOO_LONG:
                 code = cls.Codes.TOO_LONG
--- a/notifications_utils/recipient_validation/errors.py
+++ b/notifications_utils/recipient_validation/errors.py
@@ -1,5 +1,7 @@
 from enum import StrEnum, auto
 
+import phonenumbers
+
 
 class InvalidRecipientError(Exception):
     message = "Not a valid recipient address"
@@ -23,7 +25,10 @@ class InvalidPhoneError(InvalidRecipientError):
 
     # TODO: Move this somewhere else maybe? Maybe not?
     ERROR_MESSAGES = {
-        Codes.INVALID_NUMBER: "Not a valid phone number",
+        # this catches numbers with the right length but wrong digits
+        # for example UK numbers cannot start "06" as that hasn't been assigned to a purpose by ofcom,
+        # or a 9 digit UK number that does not start "01" or "0800".
+        Codes.INVALID_NUMBER: "TODO: CONTENT! Number is not valid – double check the mobile number you entered",
         Codes.TOO_LONG: "Mobile number is too long",
         Codes.TOO_SHORT: "Mobile number is too short",
         Codes.NOT_A_UK_MOBILE: (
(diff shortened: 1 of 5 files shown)


notifications_utils/recipient_validation/errors.py
```diff
@@ -23,7 +25,10 @@ class Codes(StrEnum):
 
     # TODO: Move this somewhere else maybe? Maybe not?
     ERROR_MESSAGES = {
-        Codes.INVALID_NUMBER: "Not a valid phone number",
+        # this catches numbers with the right length but wrong digits
+        # for example UK numbers cannot start "06" as that hasn't been assigned to a purpose by ofcom,
+        # or a 9 digit UK number that does not start "01" or "0800".
+        Codes.INVALID_NUMBER: "TODO: CONTENT! Number is not valid – double check the mobile number you entered",
```
CrystalPea, line 31: we could give this a content that's not TODO, and move TODO to comment to look at later.

notifications_utils/recipient_validation/errors.py
```diff
@@ -49,6 +54,26 @@ def __init__(self, *, code: Codes = Codes.INVALID_NUMBER):
         self.code = code
         super().__init__(message=self.ERROR_MESSAGES[code])
 
+    @classmethod
+    def from_phonenumbers_validationresult(cls, reason: phonenumbers.ValidationResult) -> str:
```
CrystalPea, line 58: I think if we use snake case for function name it should be `validation_result`

tests/recipient_validation/test_phone_number.py
```diff
@@ -391,3 +391,155 @@ def test_try_format_recipient_doesnt_throw():
 
 def test_format_phone_number_human_readable_doenst_throw():
     assert format_phone_number_human_readable("ALPHANUM3R1C") == "ALPHANUM3R1C"
+
+
+class TestPhoneNumbeClass:
```
rparke, line 396: typo? Shouldn't this be `class TestPhoneNumberClass`?

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -233,16 +232,35 @@ def _validate_forced_international_number(phone_number: str) -> phonenumbers.Pho
...
+        return str(self.number.country_code)
+
     def is_uk_phone_number(self):
         """
-        Returns if the number's country code is +44 - note, this includes any regions that also use the +44 country code
-        such as isle of man/jersey/guernsey. You can distinguish these with `number._is_a_crown_dependency_number`
+        Returns if the number starts with +44. Note, this includes international numbers for crown dependencies such as
+        jersey/guernsey.
+
+        # TODO: check if we still need this - looking at api, this might be able to be removed entirely since it's
```
CrystalPea, line 278: We might want to look into this - also the function name is confusing - it points at boolean return but it doesn't really return a boolean.

tests/recipient_validation/test_phone_number.py
```diff
@@ -473,15 +464,15 @@ def test_get_normalised_format_works_for_international_numbers(self, phone_numbe
     @pytest.mark.parametrize(
         "phone_number, expected_formatted",
         [
-            ("07900900123", "07900 900123"),  # UK
```
CrystalPea: tv numbers not being valid would be a bother - we'd have to have a huddle about this on tech catch up sometime - but if I understand correctly, for now the new validation will only be used for OneLogin, so not urgent.

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -220,7 +222,7 @@ def validate_phone_number(self, phone_number: str) -> phonenumbers.PhoneNumber:
         return number
 
     @staticmethod
-    def _validate_aggressively_normalised_number(phone_number: str) -> phonenumbers.PhoneNumber | None:
+    def _aggressively_normalise_number(phone_number: str) -> str:
```
CrystalPea, line 225: maybe we could find a better name?

CrystalPea (commented):
Let's work on these? ☺️

tests/recipient_validation/test_phone_number.py
CrystalPea: Fixed

notifications_utils/recipient_validation/errors.py
CrystalPea: done

notifications_utils/recipient_validation/errors.py
CrystalPea: Done

notifications_utils/recipient_validation/phone_number.py
CrystalPea: we decided to leave it be for now - tests use it, and it's not a priority to get rid of it.

tests/recipient_validation/test_phone_number.py
CrystalPea: We mentioned this in a Trello card for introducing new validation code more widely: 

https://trello.com/c/nLANAG47/62-replace-old-validation-code-entirely-with-new-code-and-remove-old-validation-code-from-utils

rparke (approved):

PR #1127, opened 2024-06-26
Put more common configuration and dependencies in `notifications-utils`

quis:
This will let us:
- remove more dependencies from each app’s requirements files
- share common config for pre-commit hooks, rather than having to manually maintain and update one in each app

### DIFF
--- a/notifications_utils/version_tools.py
+++ b/notifications_utils/version_tools.py
@@ -8,6 +8,7 @@ repo_name = "alphagov/notifications-utils"
 config_files = {
     "pyproject.toml",
     "requirements_for_test_common.txt",
+    ".pre-commit-config.yaml",
 }
 
 
--- a/notifications_utils/version_tools.py
+++ b/notifications_utils/version_tools.py
@@ -80,7 +80,7 @@ def get_file_contents_from_github(branch_or_tag, path):
     return requests.get(f"https://raw.githubusercontent.com/{repo_name}/{branch_or_tag}/{path}").text
 
 
-def copy_pyproject_toml():
+def copy_config():
     local_utils_version = get_app_version()
     remote_contents = get_file_contents_from_github(local_utils_version, "pyproject.toml")
     pyproject_file.write_text(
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -40,7 +40,7 @@
         "pypdf>=3.13.0",
         "itsdangerous>=2.1.2",
         "govuk-bank-holidays>=0.14",
-        "boto3>=1.34.100",
+        "boto3[crt]>=1.34.100",
```
leohemsted, line 43: i also wonder about whether this `crt` should be here or should just be an override on doc download. unsure

requirements_for_test_common.txt
```diff
@@ -0,0 +1,12 @@
+beautifulsoup4==4.11.1
```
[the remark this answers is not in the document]
quis: It is used, but for some reason the module is called `bs4` not `beautifulsoup` or something more sensible:
https://github.com/alphagov/notifications-admin/blob/28c493959ecc7a44d694c7ce1abcffee2afc99d1/tests/__init__.py#L7

setup.py
quis: According to this there are some performance benefits to using `crt` so maybe we should have it everywhere:
https://aws.amazon.com/blogs/storage/improving-amazon-s3-throughput-for-the-aws-cli-and-boto3-with-the-aws-common-runtime/

requirements_for_test_common.txt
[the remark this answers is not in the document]
leohemsted: sorry, i typo'd and meant to say "it's required by _api_ but shuoldn't be as it's never imported". (turns out we used to use it in api for some cap xml emergency alert stuff)

guess we can just remove it from the api requirements.in though. im not as worried about unused test requirements as i am about unused prod runtime requirements

notifications_utils/version_tools.py
```diff
@@ -4,8 +4,12 @@
 
 requirements_file = pathlib.Path("requirements.in")
 frozen_requirements_file = pathlib.Path("requirements.txt")
-pyproject_file = pathlib.Path("pyproject.toml")
 repo_name = "alphagov/notifications-utils"
+config_files = {
+    "pyproject.toml",
+    "requirements_for_test_common.txt",
+    ".pre-commit-config.yaml",
```
leohemsted, line 11: i think this fine for now looking at our files but this may make it a pain if we want to have specific additional pre commit hooks for specific repos.

eg if we wanted to add some frontend scss linting/formatting stuff to admin

PR #1128, opened 2024-06-28
Return Zendesk ticket ID when creating a ticket

klssmith:
In order to update a ticket, you need to know its ID. This returns the
ticket ID from the `send_ticket_to_zendesk` method so that we can update
a ticket immediately after creating it. We want to start adding an
internal note if a user who is not logged in creates a ticket.


whpearson (approved):
Looks good

PR #1131, opened 2024-07-17
Add validation for fixed abode addresses to `PostalAddress`

klssmith:
The `has_no_fixed_abode_address` method checks that no address lines
simply consist of "NFA" or "NFA," and that "no fixed address" or "no
fixed abode" is not in the address (all case insensitive). Letters sent
to no fixed abode addresses have been causing issues since they often
contain invalid addresses and weren't intended to be sent.

### DIFF
--- a/notifications_utils/recipient_validation/postal_address.py
+++ b/notifications_utils/recipient_validation/postal_address.py
@@ -147,6 +147,19 @@ class PostalAddress:
         )
 
     @property
+    def has_no_fixed_abode_address(self):
+        """
+        We don't want users to sent to no fixed abode addresses, so validate that
+        - no lines just consist of "NFA" (case insensitive)
+        - the address does not contain "no fixed abode" or "no fixed address" (case insensitive)
+        """
+        if any(line.lower() == "nfa" for line in self.normalised_lines):
+            return True
+        if re.search(r"no fixed (abode|address)", self.normalised, re.IGNORECASE):
+            return True
+        return False
+
+    @property
     def has_invalid_country_for_bfpo_address(self):
         """We don't want users to specify the country if they provide a BFPO number. Some BFPO numbers may resolve
         to non-UK addresses, but this will be handled as part of the BFPO delivery."""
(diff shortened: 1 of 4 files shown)


notifications_utils/recipient_validation/postal_address.py
```diff
@@ -146,6 +146,19 @@ def has_invalid_characters(self):
...
         )
 
+    @property
+    def has_no_fixed_abode_address(self):
+        """
+        We don't want users to sent to no fixed abode addresses, so validate that
+        - no lines just consist of "NFA" or "NFA," (case insensitive)
+        - the address does not contain "no fixed abode" or "no fixed address" (case insensitive)
+        """
+        if any(re.search(r"^nfa,?$", line, re.IGNORECASE) for line in self.normalised_lines):
```
quis, line 156: Normalised lines will never have trailing commas:
https://github.com/alphagov/notifications-utils/blob/0a04a4f126bb172245195eee9fe120d6ae87469f/notifications_utils/recipient_validation/postal_address.py#L37

klssmith: Good spot, I've simplified that line now since it's doesn't need to check for a trailing comma

quis (commented):
Should we also add this check to the `valid` property so these addresses are rejected?

https://github.com/alphagov/notifications-utils/blob/113a6da4ed30802ab0052efd7a565b8b1fc8c10d/notifications_utils/recipient_validation/postal_address.py#L228-L236

klssmith:
I've now updated the `valid` property, and changed the version bump to a major bump since this will change what letters are considered valid even if you don't specifically call `has_no_fixed_abode_address`.

quis (approved):

klssmith:
@quis could you re-approve please? I had to fix merge conflicts in the changelog, so your review from yesterday now counts as stale

quis (approved):

leohemsted (approved):

PR #1134, opened 2024-07-25
disallow premium numbers when validating

rparke:
Premium rate numbers are a known vector of attack. This updates the phonenumber validation code to raise an exception and fail if a service tries to create an sms notification destined for a premium rate number.

### DIFF
--- a/notifications_utils/recipient_validation/phone_number.py
+++ b/notifications_utils/recipient_validation/phone_number.py
@@ -16,6 +16,17 @@ from notifications_utils.recipient_validation.errors import InvalidPhoneError
 
 UK_PREFIX = "44"
 
+ALLOW_LIST = {
+    phonenumbers.PhoneNumberType.FIXED_LINE,
+    phonenumbers.PhoneNumberType.MOBILE,
+    phonenumbers.PhoneNumberType.FIXED_LINE_OR_MOBILE,  # ambiguous case where a number could be either landline/mobile
+    phonenumbers.PhoneNumberType.UAN,
+    phonenumbers.PhoneNumberType.PERSONAL_NUMBER,
+}
+
+DENY_LIST = [
+    phonenumbers.PhoneNumberType.PREMIUM_RATE,
+]
 
 international_phone_info = namedtuple(
     "PhoneNumber",
@@ -220,8 +231,7 @@ class PhoneNumber:
                 number = forced_international_number
             else:
                 raise InvalidPhoneError.from_phonenumbers_validation_result(reason)
-
-        if not phonenumbers.is_valid_number(number):
+        if not (phonenumbers.is_valid_number(number) & self._is_allowed_phone_number_type(number)):
             # is_possible just checks the length of a number for that country/region. is_valid checks if it's
             # a valid sequence of numbers. This doesn't cover "is this number registered to an MNO".
             # For example UK numbers cannot start "06" as that hasn't been assigned to a purpose by ofcom
@@ -231,6 +241,14 @@ class PhoneNumber:
                 raise InvalidPhoneError(code=InvalidPhoneError.Codes.INVALID_NUMBER)
 
         return number
+
+    @staticmethod
+    def _is_allowed_phone_number_type(phone_number: phonenumbers.PhoneNumber) -> bool:
+        if phonenumbers.number_type(phone_number) in DENY_LIST:
+            return False
+        if phonenumbers.number_type(phone_number) in ALLOW_LIST:
+            return True
+        return False
 
     @staticmethod
     def _is_tv_number(phone_number) -> bool:
--- a/tests/recipient_validation/test_phone_number.py
+++ b/tests/recipient_validation/test_phone_number.py
@@ -58,11 +58,6 @@ valid_uk_landlines = [
     "020 7946 0991",  # london
     "030 1234 5678",  # non-geographic
     "0550 123 4567",  # corporate numbering and voip services
-    "0800 123 4567",  # freephone
-    "0800 123 456",  # shorter freephone
-    "0800 11 11",  # shortest freephone
-    "0845 46 46",  # short premium
-    "0900 123 4567",  # premium
 ]
 
 invalid_uk_landlines = [
@@ -71,6 +66,8 @@ invalid_uk_landlines = [
     "0300 46 46",  # short but not 01x or 08x
     "0800 11 12",  # short but not 01x or 08x
     "0845 46 31",  # short but not 01x or 08x
+    "0845 46 46",  # short premium
+    "0900 123 4567",  # premium
 ]
 
 invalid_uk_mobile_phone_numbers = sum(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,16 @@
 # CHANGELOG
 
+## 82.3.0
+* Extends the validation logic for the `PhoneNumber` class to disallow premium rate numbers
+
 ## 82.2.1
 
 * Add fix to recipient_validation/phone_number.py to raise correct error if a service tries to send to an international number without that permission
 
 ## 82.2.0
-
 * Add `unsubscribe_link` argument to email templates
 
 ## 82.1.2
-
 * Write updated version number to `requirements.txt` if no `requirements.in` file found
 
 ## 82.1.1
--- a/notifications_utils/version.py
+++ b/notifications_utils/version.py
@@ -5,4 +5,4 @@
 # - `make version-minor` for new features
 # - `make version-patch` for bug fixes
 
-__version__ = "82.2.1"  # dec63cb8da7dab6a8ff9993e711d3c96
+__version__ = "82.3.0"  # c6ba6a75ce00d7c25a2c8fbfe9fb1dcb


tests/recipient_validation/test_phone_number.py
```diff
@@ -61,8 +61,6 @@
     "0800 123 4567",  # freephone
     "0800 123 456",  # shorter freephone
     "0800 11 11",  # shortest freephone
-    "0845 46 46",  # short premium
-    "0900 123 4567",  # premium
```
rparke, line 65: Not sure about the naming convention for these lists.

Premium numbers **are** valid, we just don't want them to pass validation.

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -225,6 +226,13 @@ def validate_phone_number(self, phone_number: str) -> phonenumbers.PhoneNumber:
 
         return number
 
+    @staticmethod
+    def _is_premium_number(phone_number) -> bool:
+        phonenumber_type = phonenumbers.number_type(phone_number)
+        if phonenumber_type == phonenumbers.PhoneNumberType.PREMIUM_RATE:
+            return True
+        return False
```
quis, line 234: Don’t think you need the `if` here:
```suggestion
        phonenumber_type = phonenumbers.number_type(phone_number)
        return phonenumber_type == phonenumbers.PhoneNumberType.PREMIUM_RATE:
```
And maybe not even the assignment:
```suggestion
        return phonenumbers.number_type(phone_number) == phonenumbers.PhoneNumberType.PREMIUM_RATE:
```

rparke: Agreed, just pushed the refactor

tests/recipient_validation/test_phone_number.py
```diff
@@ -73,6 +71,11 @@
     "0845 46 31",  # short but not 01x or 08x
 ]
 
+premium_uk_landlines = [
+    "0845 46 46",  # short premium
+    "0900 123 4567",  # premium
```
kr8n3r, line 76: what about 0870, 0871, 0872 and 0873

rparke: I can add some more test cases, these were the ones we were currently validating against

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -225,6 +226,10 @@ def validate_phone_number(self, phone_number: str) -> phonenumbers.PhoneNumber:
 
         return number
 
+    @staticmethod
+    def _is_premium_number(phone_number) -> bool:
+        return phonenumbers.number_type(phone_number) == phonenumbers.PhoneNumberType.PREMIUM_RATE
```
leohemsted, line 231: as mentioned before, probably good to flip this over to be `def _is_landline_or_mobile(phone_number: PhoneNumber) -> bool:` (i'll leave checking what the PhoneNumberType enum actually contains to you, and to check if 01x, 02x, 03x and 07x numbers have different types in any way)

notifications_utils/recipient_validation/errors.py
```diff
@@ -36,6 +37,7 @@ class Codes(StrEnum):
         ),
         Codes.UNKNOWN_CHARACTER: "Mobile numbers can only include: 0 1 2 3 4 5 6 7 8 9 ( ) + -",
         Codes.UNSUPPORTED_COUNTRY_CODE: "Country code not found - double check the mobile number you entered",
+        Codes.PREMIUM_NUMBER: "This is a premium rate Number - this is not supported",
```
leohemsted, line 40: worth checking with saima about what the error message should be.

esp if we're changing the validation to not just catch premium rate, i'd start with something like "This phone number is not supported" or "Not a valid landline or mobile" as starter for ten

notifications_utils/recipient_validation/phone_number.py
rparke: I have moved this into an allow and a deny list:

```
ALLOW_LIST = [
    phonenumbers.PhoneNumberType.FIXED_LINE,
    phonenumbers.PhoneNumberType.MOBILE,
    phonenumbers.PhoneNumberType.FIXED_LINE_OR_MOBILE,
    phonenumbers.PhoneNumberType.UAN,
    phonenumbers.PhoneNumberType.TOLL_FREE,
    phonenumbers.PhoneNumberType.PERSONAL_NUMBER,
]

DENY_LIST = [
    phonenumbers.PhoneNumberType.PREMIUM_RATE,
]
```

This is currently comprehensive for the different number types we use in our tests (we need all of these in the allow list). It _may_ be that we don't want to be this permissive (benefit that this new validation code gives us better visibility about exactly what types of numbers we're allowing). I don't know if that's a question for this ticket or not.

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -16,6 +16,18 @@
 
 UK_PREFIX = "44"
 
+ALLOW_LIST = [
+    phonenumbers.PhoneNumberType.FIXED_LINE,
+    phonenumbers.PhoneNumberType.MOBILE,
+    phonenumbers.PhoneNumberType.FIXED_LINE_OR_MOBILE,
+    phonenumbers.PhoneNumberType.UAN,
```
CrystalPea, line 23: do we want to allow UANs? Aren't they more like account numbers than landlines? Can SMS be sent to them?

CrystalPea (commented):
Hi, nice one, I left a couple comments - let me know what you think.

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -16,6 +16,18 @@
 
 UK_PREFIX = "44"
 
+ALLOW_LIST = [
+    phonenumbers.PhoneNumberType.FIXED_LINE,
+    phonenumbers.PhoneNumberType.MOBILE,
+    phonenumbers.PhoneNumberType.FIXED_LINE_OR_MOBILE,
```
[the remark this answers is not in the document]
rparke: As far as I can tell, they mean different things.

So each number type in that list basically evaluates to an int, so when you do comparisons like `phonenumbers.number_type(phone_number) == phonenumebrs.PhoneNumberType.MOBILE`.

That expression will evaluate to something like `2 == 2` --> True.

So because each phone number will have a single unique type, we need to check for everything.

notifications_utils/recipient_validation/phone_number.py
rparke: Good question, so this is the list of allowed numbers that make our tests pass. So for a first iteration, our current validation methods expect these numbers to pass, so I'm writing the new validation code to let them pass too.

We probably do want to be more restrictive, but It would be good to have a discussion with yourself and @leohemsted about what number types we actually want to allow.

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -16,6 +16,18 @@
...
+    phonenumbers.PhoneNumberType.UAN,
+    phonenumbers.PhoneNumberType.TOLL_FREE,
+    phonenumbers.PhoneNumberType.PERSONAL_NUMBER,
+]
```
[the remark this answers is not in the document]
rparke: Makes sense, we don't need to do any mutation logic etc on it that requires a list.

tests/recipient_validation/test_phone_number.py
rparke: Put these into the same list and removed the extra test

notifications_utils/recipient_validation/errors.py
```diff
@@ -36,6 +37,7 @@ class Codes(StrEnum):
         ),
         Codes.UNKNOWN_CHARACTER: "Mobile numbers can only include: 0 1 2 3 4 5 6 7 8 9 ( ) + -",
         Codes.UNSUPPORTED_COUNTRY_CODE: "Country code not found - double check the mobile number you entered",
+        Codes.PREMIUM_NUMBER: "Premium rate phone numbers are not supported",
```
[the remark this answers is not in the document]
rparke: Yep, removed.

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -16,6 +16,18 @@
 
 UK_PREFIX = "44"
 
+ALLOW_LIST = [
+    phonenumbers.PhoneNumberType.FIXED_LINE,
+    phonenumbers.PhoneNumberType.MOBILE,
+    phonenumbers.PhoneNumberType.FIXED_LINE_OR_MOBILE,
+    phonenumbers.PhoneNumberType.UAN,
+    phonenumbers.PhoneNumberType.TOLL_FREE,
```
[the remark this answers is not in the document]
rparke: I've removed toll free/freephone numbers.

notifications_utils/recipient_validation/phone_number.py
CrystalPea: We had a huddle and decided that it doesn't hurt to allow these - it is much less likely we will encounter them in the wild than 01 and 02 landlines, but they are all standard rate.

notifications_utils/recipient_validation/phone_number.py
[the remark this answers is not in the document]
CrystalPea: @rparke will add a comment explaining the ambiguous case.

CrystalPea (approved):

CrystalPea (approved):

PR #1138, opened 2024-08-06
Fix validation error for international numbers for services without international permissions

rparke:
Separate out check for if service can't send to international but tries to

When this check was happening with all the rest of the number validation, when it failed we re-run it with a throoghly normalised number, which made phonenumbers libary think it's a UK landline that's too long for US numbers.

When we do it separately, we raise the NOT_A_UK_MOBILE error as we should, and we don't try to overnormalise when we get it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 82.2.1
+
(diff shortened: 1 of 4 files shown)


CrystalPea (approved):
:shipit:

PR #1143, opened 2024-08-22
Add allow landline flag

rparke:
The default behaviour of `notifications_utils.recipient_validation.phone_number.PhoneNumber` validates landline numbers. We want to be able to toggle this on or off _via_ a constructor argument so that this code can be used to validate phone numbers for **all** services, not just those with the `sms_to_uk_landline` permission.

This change adds the constructor argument `allow_landline` and makes changes to the validation code to throw an exception if a valid landline is passed when `allow_landline=False`.

### DIFF
--- a/notifications_utils/recipient_validation/phone_number.py
+++ b/notifications_utils/recipient_validation/phone_number.py
@@ -174,36 +174,43 @@ def format_phone_number_human_readable(phone_number):
 
 class PhoneNumber:
     """
-    A class that contains phone number validation.
+    A class that parses and performs validation checks on phonenumbers against service permissions
 
-    Supports mobile and landline numbers. When creating an object you must specify whether you are expecting
-    international phone numbers to be allowed or not.
+    Can be instantiated for all Phone Numbers other than premium numbers. Instansiation checks that the number
+    you are trying to send to is possible, validate checks the number against a services permissions passed to it
+    to ensure it can send.
+
+    Examples:
+        number = PhoneNumber("07910777555")
+        number.validate(allow_international_number = False, allow_uk_landline = False)
     """
 
-    def __init__(self, phone_number: str, *, allow_international: bool, allow_landline: bool = False) -> None:
-        self.raw_input = phone_number
-        self.allow_international = allow_international
-        self.allow_landline = allow_landline
+    def __init__(self, phone_number: str) -> None:
         try:
-            self.number = self.validate_phone_number(phone_number)
+            self.number = self.parse_phone_number(phone_number)
         except InvalidPhoneError:
             phone_number = self._thoroughly_normalise_number(phone_number)
-            self.number = self.validate_phone_number(phone_number)
-        self._raise_if_service_cannot_send_to_uk_landline_but_tries_to(phone_number)
-        self._raise_if_service_cannot_send_to_international_but_tries_to(phone_number)
+            self.number = self.parse_phone_number(phone_number)
+        self._phone_number = phone_number
 
-    def _raise_if_service_cannot_send_to_international_but_tries_to(self, phone_number):
-        number = self._try_parse_number(phone_number)
-        if not self.allow_international and str(number.country_code) != UK_PREFIX:
+    def _raise_if_service_cannot_send_to_international_but_tries_to(self, allow_international: bool = False):
+        number = self._try_parse_number(self._phone_number)
+        if not allow_international and str(number.country_code) != UK_PREFIX:
             raise InvalidPhoneError(code=InvalidPhoneError.Codes.NOT_A_UK_MOBILE)
 
-    def _raise_if_service_cannot_send_to_uk_landline_but_tries_to(self, phone_number):
-        phone_number = self._try_parse_number(phone_number)
-        if phone_number.country_code != 44:
+    def _raise_if_service_cannot_send_to_uk_landline_but_tries_to(
+        self, allow_uk_landline: bool = False, allow_international_number: bool = False
+    ):
+        phone_number = self._try_parse_number(self._phone_number)
+        if phone_number.country_code != int(UK_PREFIX):
             return
         is_landline = phonenumbers.number_type(phone_number) in LANDLINE_CODES
-        if not self.allow_landline and is_landline:
+        if not allow_uk_landline and is_landline:
             raise InvalidPhoneError(code=InvalidPhoneError.Codes.NOT_A_UK_MOBILE)
+
+    def validate(self, allow_international_number: bool = False, allow_uk_landline: bool = False) -> None:
+        self._raise_if_service_cannot_send_to_international_but_tries_to(allow_international=allow_international_number)
+        self._raise_if_service_cannot_send_to_uk_landline_but_tries_to(allow_uk_landline=allow_uk_landline)
 
     @staticmethod
     def _try_parse_number(phone_number):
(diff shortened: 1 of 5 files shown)


notifications_utils/recipients.py
```diff
@@ -321,9 +321,16 @@ def _get_error_for_field(self, key, value):  # noqa: C901
                     email_address.validate_email_address(value)
                 if self.template_type == "sms":
                     if self.allow_sms_to_uk_landline:
```
CrystalPea, line 323: I think we don't need this if/else statement here now?

We should always do what's in the `if`, and never what's in the `else`, right?

CrystalPea (changes requested):
Hey, nice work, only I think for services that do not have 'allow_sms_to_uk_landline` flag, we still use old validation code, but we should be using the new one?

More details in code comment.

notifications_utils/recipients.py
rparke: Currently we can do both. We will be removing the `phone_number.validate_phone_number` function in the next ticket. So in the long run, yes. But the scope of this PR is just adding the allow_landline flag. (In practice we could probably just remove the branching here and just use the `if` ... probably won't break anything, but I think that's beyond the scope of this PR).

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -219,6 +226,8 @@ def validate_phone_number(self, phone_number: str) -> phonenumbers.PhoneNumber:
         # of those cases separately before we parse with the phonenumbers library
         self._raise_if_phone_number_contains_invalid_characters(phone_number)
 
+        self._raise_if_service_cannot_send_to_landline_but_tries_to(phone_number)
```
CrystalPea, line 229: [Suggestion] I think this check should happen next to where we call `_raise_if_service_cannot_send_to_international_but_tries_to` in init - this is because it has to do with service permissions, not with number being correct or not.

notifications_utils/recipients.py
CrystalPea: Hmmm - I can see how it could be out of scope - but since it's just removing an if statement, and most likely it will not break anything, isn't it more efficient to do it here? If you feel that will muddy the waters for you though - that's fair, it's ok to do it in a separate PR.

tests/recipient_validation/test_phone_number.py
```diff
@@ -561,5 +561,5 @@ def test_tv_number_passes(self, phone_number, expected_valid_number):
     )
     def test_international_does_not_normalise_to_uk_number(self, phone_number, expected_error_code):
         with pytest.raises(InvalidPhoneError) as exc:
-            PhoneNumber(phone_number, allow_international=False)
+            PhoneNumber(phone_number, allow_international=False, allow_landline=True)
```
CrystalPea, line 564: We are missing some unit tests: currently there is no test for when we use `PhoneNumber` class, but `allow_landline=False`. So we are never actually testing the happy path with that flag set to false, and we never test the `_raise_if_service_cannot_send_to_landline_but_tries_to` doing its job and raising an error, either.

Let's add the missing tests?

rparke: Good spot. I've noticed that we also don't have this test for international numbers. Probably something to rectify at a later date.

CrystalPea (approved):

leohemsted:
while we're making a breaking change, is this a suitable time to move towards a less restrictive flow?

```
x = PhoneNumber("+1 555 867 5309")

assert x.is_international() is True
assert x.is_landline() is False

y = PhoneNumber("+441234 567890")

assert y.is_international() is True
assert y.is_landline() is False

# still raises if the phone isn't valid at all
with pytest.raises(InvalidPhoneNumber):
  PhoneNumber("+44 12345678901234567890")

# what do we do with valid phones we dont think we'll ever handle
hmmmm = PhoneNumber("0845 1234 1234")
```

rparke:
> while we're making a breaking change, is this a suitable time to move towards a less restrictive flow?
> 
> ie something like:
> 
> ```
> x = PhoneNumber("+1 555 867 5309")
> 
> assert x.is_international() is True
> assert x.is_landline() is False
> 
> # or maybe we call a validate method that we pass flags to that raises
> with pytest.raises(InvalidPhoneNumber):
>   x.validate(allow_international=False, allow_landline=True)
> 
> # still raises if the phone isn't valid at all
> with pytest.raises(InvalidPhoneNumber):
>   PhoneNumber("+44 12345678901234567890")
> 
> # what do we do with valid phones we dont think we'll ever handle - maybe just raise?
> hmmmm = PhoneNumber("0845 1234 1234")
> ```

I guess the question is: what should an instantiated PhoneNumber object represent? Are we saying that if it's instantiated it's a valid number for somewhere in the world: "it could be parsed as a number, potentially after some massaging from us (stripping zeros etc etc)" It also has some methods that give you information about that number: so `PhoneNumber.validate` runs checks on behalf of api/admin calling it.

notifications_utils/recipients.py
```diff
@@ -321,9 +321,18 @@ def _get_error_for_field(self, key, value):  # noqa: C901
...
-                        PhoneNumber(value, allow_international=self.allow_international_sms)
+                        number = PhoneNumber(
+                            value,
+                        )
```
quis, line 326:
```suggestion
                        number = PhoneNumber(value)
```

notifications_utils/recipients.py
```diff
@@ -321,9 +321,18 @@ def _get_error_for_field(self, key, value):  # noqa: C901
...
+                        number.validate_phone_number(
+                            allow_international_number=self.allow_international_sms,
+                            allow_uk_landline=self.allow_sms_to_uk_landline,
+                        )
                     else:
-                        phone_number.validate_phone_number(value, international=self.allow_international_sms)
+                        phone_number.validate_phone_number(
+                            value,
+                            international=self.allow_international_sms,
+                        )
```
quis, line 335: Why are we still using the old way of doing things here?

quis (changes requested):

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -197,18 +216,20 @@ def _try_parse_number(phone_number):
         except phonenumbers.NumberParseException as e:
             raise InvalidPhoneError(code=InvalidPhoneError.Codes.INVALID_NUMBER) from e
 
+    @staticmethod
+    def _raise_if_phone_number_is_empty(number: str) -> None:
+        if number == "":
```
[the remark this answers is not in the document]
rparke: Can add it, I think it should always be a string, but can't hurt

notifications_utils/recipients.py
rparke: To be removed in the next ticket.

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -197,18 +216,20 @@ def _try_parse_number(phone_number):
...
             raise InvalidPhoneError(code=InvalidPhoneError.Codes.UNKNOWN_CHARACTER)
 
-    @staticmethod
-    def _raise_if_phone_number_is_empty(number: str) -> None:
-        if number == "":
-            raise InvalidPhoneError(code=InvalidPhoneError.Codes.TOO_SHORT)
-
-    def validate_phone_number(self, phone_number: str) -> phonenumbers.PhoneNumber:
+    # split into private method "parse_phone_number" and have a public method
+    # "validate_phone_number_against_service_permissions" that can be called to check against service permissions
```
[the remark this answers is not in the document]
rparke: removed, they aren't needed

quis (approved):

PR #1146, opened 2024-09-01
Empty strings for phone numbers return "TOO_SHORT" error message

rparke:
Adds validation check to PhoneNumber so that the error `TOO_SHORT` is raised if an empty string is passed into the object as a phone number

Users of the v2 api expect, and often require, consistent error messages to be returned from the API as their code often does comparisons against errors as strings to handle notify errors. In the case of empty strings, an error corresponding to InvalidPhoneError.Codes.TOO_SHORT is expected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+# 82.6.1
+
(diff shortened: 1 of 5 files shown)


leohemsted (approved):
imagine not even checking your input at all before passing it to a third party api 😭

PR #1147, opened 2024-09-02
`LazyLocalGetter` improvements

risicle:
On the long road to https://trello.com/c/ORGrd1jn/498-upgrade-docker-images-for-our-ecs-apps-to-debian-bookworm

Because of the odd way the `api` app (at least) uses its api clients it becomes useful to be able to see a `LazyLocalGetter`'s inner type without populating it.

But to ensure this information is never *wrong* expose this as an _enforced_ `expected_type` mechanism.

~Also add `LazyLocalGetterResetter` which is (in this iteration) just a simple wrapper around a list that allows all the registered `LazyLocalGetter`s to be reset in a single call.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,16 @@
 # CHANGELOG
 
+## 82.7.0
+
+* Add `expected_type` mechanism to `LazyLocalGetter`
+
 # 82.6.1
 
 * Adds validation check to `PhoneNumber` so that it returns the expected error message `TOO_SHORT` if an empty string is passed. This has caused issues with users of the v2 API getting inconsistent error messages
 
 ## 82.6.0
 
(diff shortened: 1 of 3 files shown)


robinjam (approved):
Seems fine to me

robinjam:
What are the semantics of `LazyLocalGette.clear` and `LazyLocalGetterResetter.reset`? I'm assuming they only reset the context var for the current thread (as opposed to for all threads)?

Is it worth adding a little comment explaining this?

risicle:
Perhaps - I'm considering getting rid of the resetter entirely actually - maybe I should just maintain a list of argument-less lambdas in each app for resetting things. This way people could easily add arbitrary things to that list like cache decorator resetters (I dream).

risicle:
Gone.

robinjam (approved):

PR #1148, opened 2024-09-09
Migrate `AssetFingerprinter` into utils

quis:
So we can share it between our frontend apps.

This continues its journey, having been:
- introduced here: https://github.com/Crown-Commercial-Service/digitalmarketplace-buyer-frontend/pull/111
- refactored into shared code here: https://github.com/Crown-Commercial-Service/digitalmarketplace-utils/pull/102
- copied into Notify here: https://github.com/alphagov/notifications-admin/pull/171 (“They have it in a shared codebase, we only have one frontend app so don’t need to do that.”)
- copied into Document Download here: https://github.com/alphagov/document-download-frontend/pull/1

### DIFF
--- a/tests/test_asset_fingerprinter.py
+++ b/tests/test_asset_fingerprinter.py
@@ -55,7 +55,7 @@ class TestAssetFingerprint:
         js_hash = asset_fingerprinter.get_asset_fingerprint("application.js")
         assert js_hash != css_hash
 
-    def test_hash_gets_cached(self, mocker):
+    def test_gets_fingerprint_from_cache(self, mocker):
         get_file_content_mock = mocker.patch.object(AssetFingerprinter, "get_asset_file_contents")
         get_file_content_mock.return_value = b"""
             body {
--- a/tests/test_asset_fingerprinter.py
+++ b/tests/test_asset_fingerprinter.py
@@ -1,0 +1,98 @@
+from notifications_utils.asset_fingerprinter import AssetFingerprinter
+
+
+class TestAssetFingerprint:
+    def test_url_format(self, mocker):
+        get_file_content_mock = mocker.patch.object(AssetFingerprinter, "get_asset_file_contents")
+        get_file_content_mock.return_value = """
+            body {
+                font-family: nta;
+            }
+        """.encode(
+            "utf-8"
+        )
+        asset_fingerprinter = AssetFingerprinter(asset_root="/suppliers/static/")
+        assert (
+            asset_fingerprinter.get_url("application.css")
+            == "/suppliers/static/application.css?418e6f4a6cdf1142e45c072ed3e1c90a"  # noqa
+        )
+        assert (
+            asset_fingerprinter.get_url("application-ie6.css")
+            == "/suppliers/static/application-ie6.css?418e6f4a6cdf1142e45c072ed3e1c90a"  # noqa
+        )
+
+    def test_building_file_path(self, mocker):
+        get_file_content_mock = mocker.patch.object(AssetFingerprinter, "get_asset_file_contents")
+        get_file_content_mock.return_value = """
+            document.write('Hello world!');
+        """.encode(
+            "utf-8"
+        )
+        fingerprinter = AssetFingerprinter()
+        fingerprinter.get_url("javascripts/application.js")
+        fingerprinter.get_asset_file_contents.assert_called_with("app/static/javascripts/application.js")
+
+    def test_hashes_are_consistent(self, mocker):
+        get_file_content_mock = mocker.patch.object(AssetFingerprinter, "get_asset_file_contents")
+        get_file_content_mock.return_value = """
+            body {
+                font-family: nta;
+            }
+        """.encode(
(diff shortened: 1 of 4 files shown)


tests/test_asset_fingerprinter.py
```diff
@@ -0,0 +1,86 @@
...
+    def test_hash_gets_cached(self, mocker):
+        get_file_content_mock = mocker.patch.object(AssetFingerprinter, "get_asset_file_contents")
+        get_file_content_mock.return_value = b"""
+            body {
+                font-family: nta;
+            }
+        """
+        fingerprinter = AssetFingerprinter()
+        assert fingerprinter.get_url("application.css") == "/static/application.css?418e6f4a6cdf1142e45c072ed3e1c90a"
+        fingerprinter._cache["application.css"] = "a1a1a1"
```
CrystalPea, line 67: aren't we manually setting that cache here?

If we are, then what are we testing?

tests/test_asset_fingerprinter.py
```diff
@@ -0,0 +1,86 @@
...
+            )
+            == "/static/application.css"
+        )
+        assert fingerprinter._cache == {}
+
+
+class TestAssetFingerprintWithUnicode:
+    def test_can_read_self(self):
+        "Ralph’s apostrophe is a string containing a unicode character"
+        AssetFingerprinter(filesystem_path="tests/").get_url("test_asset_fingerprinter.py")
```
CrystalPea, line 86: I don't think we are asserting anything in this test?

CrystalPea (commented):
Hi, nice work in de-duplicating the AssetFingerprinter class across our apps 🙌🏼 . I left a couple comments regarding the tests - lemme know what you think ☺️

tests/test_asset_fingerprinter.py
quis: I think it’s testing that the value returned is whatever’s in the cache, i.e. it isn’t recomputing it every time

tests/test_asset_fingerprinter.py
quis: This test checks that the fingerprinter can load a file (in this case the test file itself) containing a non-ascii character (Ralph’s apostrophe). This might not be a problem now we don’t have to support Python 2. But it is a cute test

tests/test_asset_fingerprinter.py
CrystalPea: Ahh fair, so maybe the test name needs changing, for example to `test_get_fingerprint_from_cache` - because it's about getting a thing from cache, not it getting cached.

tests/test_asset_fingerprinter.py
CrystalPea: Hah, fair, only the string is not matched with any mock or anything, it's just there on it's own, I don't understand how this works tbh 😳

quis: Correct, it’s not doing any mocking. It opens and reads the file at `tests/test_asset_fingerprinter.py`, which contains the string. Just so happens that that file is the same one that contains the test itself.

tests/test_asset_fingerprinter.py
```diff
@@ -0,0 +1,86 @@
...
+            }
+        """
+        css_hash = asset_fingerprinter.get_asset_fingerprint("application.css")
+        get_file_content_mock.return_value = b"""
+            document.write('Hello world!');
+        """
+        js_hash = asset_fingerprinter.get_asset_fingerprint("application.js")
+        assert js_hash != css_hash
+
+    def test_hash_gets_cached(self, mocker):
```
quis, line 58:
```suggestion
    def test_gets_fingerprint_from_cache(self, mocker):
```

tests/test_asset_fingerprinter.py
quis: Changed in [c205017](https://github.com/alphagov/notifications-utils/pull/1148/commits/c205017152448763612fc90dbf9a523c204279b4)

tests/test_asset_fingerprinter.py
CrystalPea: To me it looks like this test isn't testing anything? There is no assertion, no comparison... How would I make it fail?

quis: If it couldn’t read the file it would raise an exception (causing the test to fail). You could make it fail by modifying `AssetFingerprinter.get_asset_file_contents` to open the file as ASCII instead of bytes

tombye: Could you make that explicit by wrapping it in a try/catch, as described by https://miguendes.me/how-to-check-if-an-exception-is-raised-or-not-with-pytest?

CrystalPea: @quis Ahh I get it now, it scans the whole test file - that's why it contains that string, too.

It took me so long to get what's happening there, because the test structure is very unusual.

CrystalPea (approved):

leohemsted (approved):

PR #1149, opened 2024-09-12
Updates phone_numbers library to 8.13.45 to fix a validation bug with certain Jersey phone numbers

rparke:
It was discovered that a subset of valid Jersey phone numbers were being incorrectly invalidated by the phone_numbers library code due to a bug in the regex phone numbers were being validated against for the Jersey region. This new code provides a fix

### DIFF
--- a/tests/recipient_validation/test_phone_number.py
+++ b/tests/recipient_validation/test_phone_number.py
@@ -46,6 +46,14 @@ valid_international_phone_numbers = [
     "+43 676 111 222 333 4",  # Austrian 13 digit phone numbers
 ]
 
+# placeholder to be removed when the old validation code is removed
+# categorising these numbers as international numbers will cause old checks
+# for uk numbers to fail
+valid_channel_island_numbers = [
+    "+447797292290",  # Jersey
+    "+447797333214",  # Jersey
+]
+
 
 valid_mobile_phone_numbers = valid_uk_mobile_phone_numbers + valid_international_phone_numbers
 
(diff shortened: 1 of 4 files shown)


tests/recipient_validation/test_phone_number.py
```diff
@@ -564,6 +572,13 @@ def test_international_does_not_normalise_to_uk_number(self, phone_number, expec
             PhoneNumber(phone_number, allow_international=False)
         assert exc.value.code == expected_error_code
 
+    @pytest.mark.parametrize("phone_number", valid_channel_island_numbers)
```
rparke, line 575: Not sure if we need an explicit test here. To my mind we shouldn't be indirectly testing library code. Obviously this gives us peace of mind right now that the bug is fixed for some of the patterns listed as an issue, but we aren't actually testing our code here, rather a dependency.

CrystalPea: That is true - I'm happy to not have this unit test 🙌🏼

quis: If the library broke this again, would we want to know about it before we upgraded the dependency to the newest version? I would argue yes.

rparke: I guess the problem is ... there are lots of ways this can break. It didn't break for all Jersey telephone numbers, and it's not really possible to predict what _types_ of numbers it will be broken for in the future.

quis (approved):

risicle:
You're likely going to have problems propagating this to your app because of the 83.0.0 change.

rparke:
> You're likely going to have problems propagating this to your app because of the 83.0.0 change.

Yeh, I've just realised that. Not wanting to touch the idea of resets on main with a ten foot pole, is a revert a good idea or is our changelog and commit history just going to look a bit too messy?

I'm just conscious that until this change gets into the API, certain valid numbers can't be reached by it.

risicle:
:shrug: microservices.

The real way to solve this would be a stable-branch fix, but I doubt anyone's set up the tooling to be able to handle anything other than a single fast-forward release stream.

risicle:
I guess there's nothing for it but to do a partial-revert.

PR #1153, opened 2024-09-17
Update Zendesk client to add a new optional argument for the date a user was created

klssmith:
The Zendesk client now takes an optional argument called `user_created_date`. If provided, this value is used to populate a new field we have on the form which is a calendar date picker. This will let us analyse how long the people opening tickets have been using Notify for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 84.2.0
+
+* The Zendesk client takes a new optional argument, `user_created_at` which populates a new field on the Notify Zendesk form if provided.
+
 ## 84.1.1
 
 * Remove GIR 0AA from valid postcodes
--- a/notifications_utils/clients/zendesk/zendesk_client.py
+++ b/notifications_utils/clients/zendesk/zendesk_client.py
@@ -4,6 +4,7 @@ import enum
 import typing
 from urllib.parse import urlencode
 
+import pytz
 import requests
 from flask import current_app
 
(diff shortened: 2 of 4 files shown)


quis (commented):
Would it be better if the client accepted a `datetime` argument and handled string formatting? This way you’d guarantee consistent date formats no matter what the app code calls it with

klssmith:
This new argument will only be added from one place - that's in admin when you use the feedback form. That's the only place we create a ticket where we know who the authenticated user is and have this information available.

In admin, we have the user created_at in a string format of `"2023-11-07T08:34:54.857402Z"`, which needs to change into the format of `"YYYY-MM-DD"` in local time for use here. [This is what I currently have for the admin part](https://github.com/alphagov/notifications-admin/compare/send-user-creation-date-to-zendesk?expand=1). Do you think it would be better if admin changes that string to  a UTC datetime, and the conversion to BST and back to a string is done here? I'm slightly undecided since the date only gets sent from one place

quis:
Maybe it’s fine if you know it’s only ever going to be called from one place, but it feels like a bit of a smell to me.

You have a function that accepts strings in any format but:
- expects them to be in the YYYY/MM/DD
- expects them to have gone through a conversion to local time first

…and:
- doesn’t document this
- doesn’t/can’t check this programmatically

If anyone else was going to reuse this function that’s a lot of hidden context. Whereas if it has to be called it with a `datetime` it’s much less ambiguous

quis (approved):

PR #1155, opened 2024-09-19
TMP version_tools: use importlib to make some mechanisms more reliable

risicle:
The mechanism currently used by `version_tools.copy_config()` has a few problems - it relies on:

 - always having network access to github
 - the specific layout of github's url structure
 - the layout of files in the new target version of utils being the same as that expected by the version of `copy_config()` that's running.
 - the new target version of utils being a released, tagged version (which isn't always the case during development)

This changes `copy_config()` to use the "correct" python mechanism for accessing auxiliary files from packages.

In doing this, `requirements_for_test.txt` and `requirements_for_test_common.txt` here become `requirements_for_test.in` and `requirements_for_test_common.in` and it is suggested that apps using utils' requirements_for_test should be using it to freeze their test dependencies, allowing `pip-sync` to be used in their freezing workflow. This leads to a workflow where the freeze-command is able to _enter_ the just-frozen-requirements venv, from where it will be possible to fetch the updated files via `copy_config()`.


leohemsted:
superceded by https://github.com/alphagov/notifications-utils/pull/1166

PR #1156, opened 2024-09-20
Rewrite `SerialisedModel` to use annotations

quis:
Upstreams the work from https://github.com/alphagov/notifications-admin/pull/5210

# Before 

```python
class User:
    ALLOWED_PROPERTIES = {"name"}
```

# After

```python
class User:
    name: str
```

### DIFF
--- a/notifications_utils/serialised_model.py
+++ b/notifications_utils/serialised_model.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 
 
-class SerialisedModel(ABC):
+class SerialisedModel:
     """
     A SerialisedModel takes a dictionary, typically created by
     serialising a database object. It then takes the value of specified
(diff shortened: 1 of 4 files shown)


notifications_utils/serialised_model.py
```diff
@@ -1,33 +1,48 @@
...
-        pass
+    def __new__(cls, *args, **kwargs):
+        for parent in cls.__mro__:
+            cls.__annotations__ = getattr(parent, "__annotations__", {}) | cls.__annotations__
+        return super().__new__(cls)
 
     def __init__(self, _dict):
-        for property in self.ALLOWED_PROPERTIES:
-            setattr(self, property, _dict[property])
+        for property, type_ in self.__annotations__.items():
```
klssmith, line 33: If the plan is to use this `__init__` method in `SerialisedModel` in admin without changing it, then I think you will need to add the `self._dict = _dict or {}` line that admin has. But if you're going to override it then this is fine as it is.

quis: Keeping it strict here. This is how the API uses it at the moment (all fields are required in the `dict`).

klssmith (approved):

PR #1158, opened 2024-10-02
Remove old phonenumber validation

rparke:


tests/recipient_validation/test_phone_number.py
```diff
@@ -93,7 +83,7 @@
                 ),
             ),
             (
-                InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.TOO_SHORT],
+                InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.INVALID_NUMBER],
```
rparke, line 86: The phonenumbers library throws different errors when trying to run `phonenumbers.is_valid_number(number)` against a number that has been successfully parsed as "possible", but is shown to be invalid when the library tries to validate the number against it's metadata catalogue (it contains a list of regular expressions for each country and matches against them).

This just returns a boolean (it's either valid or it isn't) but doesn't attempt to figure out how or why it didn't match any known regular expression. Without superimposing the old validation code onto this that **does** do a check against the strings for length _etc._ I'm not sure if we can get exactly the same error messages to return.

tests/recipient_validation/test_phone_number.py
```diff
@@ -119,11 +109,11 @@
...
 
 invalid_international_numbers = [
-    ("80100000000", InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.UNSUPPORTED_COUNTRY_CODE]),
-    ("1234567", InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.TOO_SHORT]),
+    ("80100000000", InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.TOO_LONG]),
+    ("1234567", InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.INVALID_NUMBER]),
     (
         "+682 1234",
-        InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.TOO_SHORT],
+        InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.INVALID_NUMBER],
```
rparke, line 116: as on line 86

tests/recipient_validation/test_phone_number.py
```diff
@@ -141,10 +131,8 @@
         )
     )
     + invalid_international_numbers
-    + [
-        (num, InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.NOT_A_UK_MOBILE])
-        for num in valid_uk_landlines + invalid_uk_landlines
-    ]
+    + [(num, InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.INVALID_NUMBER]) for num in invalid_uk_landlines]
+    + [(num, InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.NOT_A_UK_MOBILE]) for num in valid_uk_landlines]
```
rparke, line 135: as on line 86

tests/recipient_validation/test_phone_number.py
```diff
@@ -361,15 +362,16 @@ def test_validates_against_guestlist_of_international_phone_numbers(recipient_nu
         ("447900900123", "07900 900123"),  # UK
         ("20-12-1234-1234", "+20 12 12341234"),  # Egypt
         ("00201212341234", "+20 12 12341234"),  # Egypt
-        ("1664 0000000", "+1 664-000-0000"),  # Montserrat
+        ("16644918766", "+1 664-491-8766"),  # Montserrat
         ("7 499 1231212", "+7 499 123-12-12"),  # Moscow (Russia)
         ("1-202-555-0104", "+1 202-555-0104"),  # Washington DC (USA)
-        ("+23051234567", "+230 5123 4567"),  # Mauritius
+        ("+2304031001", "+230 403 1001"),  # Mauritius
```
rparke, line 368: Fixes numbers that were in fact not valid for the country

tests/recipient_validation/test_phone_number.py
```diff
@@ -383,19 +385,19 @@ def test_format_uk_and_international_phone_numbers(phone_number, expected_format
         ("foo", "foo"),
         ("TeSt@ExAmPl3.com", "test@exampl3.com"),
         ("+4407900 900 123", "447900900123"),
-        ("+1 800 555 5555", "18005555555"),
+        ("+1 928-282-4541", "19282824541"),
```
rparke, line 388: Fixes number that was not in fact valid for the country

tests/recipient_validation/test_phone_number.py
```diff
@@ -383,19 +385,19 @@ def test_format_uk_and_international_phone_numbers(phone_number, expected_format
...
     ],
 )
 def test_format_recipient(recipient, expected_formatted):
     assert format_recipient(recipient) == expected_formatted
 
 
-def test_try_format_recipient_doesnt_throw():
-    assert try_validate_and_format_phone_number("ALPHANUM3R1C") == "ALPHANUM3R1C"
+# def test_try_format_recipient_doesnt_throw():
+#     assert try_validate_and_format_phone_number("ALPHANUM3R1C") == "ALPHANUM3R1C"
```
rparke, line 396: currently commented out, confirming we can delete as we don't have an equivalent method. Invalid characters would be caught at the instantiation of the object, not during a try_and_validate_and_format_method

tests/recipient_validation/test_phone_number.py
```diff
@@ -383,19 +385,19 @@ def test_format_uk_and_international_phone_numbers(phone_number, expected_format
...
 
 
-def test_format_phone_number_human_readable_doenst_throw():
-    assert format_phone_number_human_readable("ALPHANUM3R1C") == "ALPHANUM3R1C"
+# def test_format_phone_number_human_readable_doenst_throw():
+#     assert format_phone_number_human_readable("ALPHANUM3R1C") == "ALPHANUM3R1C"
```
rparke, line 400: currently commented out, confirming we can delete as we don't have an equivalent method. Invalid characters would be caught at the instantiation of the object, not during a try_and_validate_and_format_method

tests/recipient_validation/test_phone_number.py
CrystalPea: That explanation would do well in a commit message for this change 🙌🏼

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -210,7 +84,17 @@ def _raise_if_service_cannot_send_to_uk_landline_but_tries_to(
 
     def validate(self, allow_international_number: bool = False, allow_uk_landline: bool = False) -> None:
         self._raise_if_service_cannot_send_to_international_but_tries_to(allow_international=allow_international_number)
-        self._raise_if_service_cannot_send_to_uk_landline_but_tries_to(allow_uk_landline=allow_uk_landline)
+        try:
+            self._raise_if_service_cannot_send_to_uk_landline_but_tries_to(allow_uk_landline=allow_uk_landline)
+        except InvalidPhoneError:
```
CrystalPea, line 89: what's the use case when we need to catch and do this extra normalisation? Shouldn't the phone number be already normalised from when we initialise the class?

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -251,22 +135,28 @@ def parse_phone_number(self, phone_number: str) -> phonenumbers.PhoneNumber:
...
+
         if not (phonenumbers.is_valid_number(number) & self._is_allowed_phone_number_type(number)):
             # is_possible just checks the length of a number for that country/region. is_valid checks if it's
             # a valid sequence of numbers. This doesn't cover "is this number registered to an MNO".
             # For example UK numbers cannot start "06" as that hasn't been assigned to a purpose by ofcom
             if self._is_tv_number(number):
                 return number
             else:
-                raise InvalidPhoneError(code=InvalidPhoneError.Codes.INVALID_NUMBER)
+                if phonenumbers.number_type(number) == phonenumbers.PhoneNumberType.UNKNOWN:
```
CrystalPea, line 154: All those three cases result in the same error raised - do they need to be three separate cases?

tests/recipient_validation/test_phone_number.py
```diff
@@ -252,52 +243,59 @@ def test_get_international_info(phone_number, expected_info):
...
         "801234-7890",
         "(8-0)-1234-7890",
     ],
 )
 def test_get_international_info_raises(phone_number):
     with pytest.raises(InvalidPhoneError) as error:
-        get_international_phone_info(phone_number)
-    assert str(error.value) == InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.UNSUPPORTED_COUNTRY_CODE]
+        number = PhoneNumber(phone_number)
+        number.get_international_phone_info()
```
CrystalPea, line 264: I wonder, does this test raise on calling this line, or on instantiation? My hypothesis is the test might still work without line 264? But then of course we'd need to rename it - maybe to something like `test_PhoneNumber_raises_for_invalid_international_numbers()`?

tests/recipient_validation/test_phone_number.py
```diff
@@ -252,52 +243,59 @@ def test_get_international_info(phone_number, expected_info):
...
+    invalid_number = str(error.value) == InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.INVALID_NUMBER]
+    too_long = str(error.value) == InvalidPhoneError.ERROR_MESSAGES[InvalidPhoneError.Codes.TOO_LONG]
```
CrystalPea, line 266: Do we care what exact error message we get here? If not, maybe we could just check that it raises `InvalidPhoneError`?

tests/recipient_validation/test_phone_number.py
CrystalPea: Would be good to gather all changes where you change a number to be actually a valid number into one commit with commit message like your comment here.

tests/recipient_validation/test_phone_number.py
CrystalPea: I think the point of the method is to not catch invalid characters.

Are we testing `format_recipient()` elsewhere? If we do, and we are sure nowhere else uses `try_validate_and_format_phone_number()`, then we can remove this test.

tests/recipient_validation/test_phone_number.py
CrystalPea: It looks like we are using this a lot in admin repo: https://github.com/search?q=repo%3Aalphagov%2Fnotifications-admin%20format_phone_number_human_readable&type=code

So maybe we actually we do need it? Or an equivalent?

tests/recipient_validation/test_phone_number.py
CrystalPea: It looks like we do use it in API repo: https://github.com/search?q=repo%3Aalphagov%2Fnotifications-api%20try_validate_and_format_phone_number&type=code

So looks like we need that method, or an equivalent.

tests/test_recipient_csv.py
```diff
@@ -708,21 +708,10 @@ def test_bad_or_missing_data(
...
-            1234
-            +447900123
-        """,
-            {0, 1, 2},
-        ),
         (
             """
             phone number, country
-            1-202-555-0104, USA
-            +12025550104, USA
```
CrystalPea, line 724: why reduction in lines and test cases?

Looks like we are only testing happy path now?

tests/test_recipient_validation.py
```diff
@@ -14,7 +14,7 @@
         ("foo", "foo"),
         ("TeSt@ExAmPl3.com", "test@exampl3.com"),
         ("+4407900 900 123", "447900900123"),
-        ("+1 800 555 5555", "18005555555"),
+        ("+1 202-555-0104", "12025550104"),  # invalid number, swap test case for working number
```
CrystalPea, line 17: This shouldn't be a code comment, but a commit message instead.

CrystalPea (commented):
Hi, this is quite a beefy story, I appreciate you working on this.

I left some comments, let me know what you think.

More generally, I'd recommend:

- organising the PR into commit messages that tell a story - so it's easier to review now, and easier to understand in the future why decisions have been taken
- double checking that old functions we are getting rid of are not used anywhere. It looks like at least two are still used in api/admin?
- also the build is failing, so maybe some more tests to fix?

Also the PR is still in draft - is it mbecause it's not ready for review? If so, let me know what is left to be done. Otherwise, move to "Ready to review?"

tests/test_recipient_csv.py
rparke: These pass, it was just for ease of testing one edge case. They'll be added back in

tests/recipient_validation/test_phone_number.py
rparke: I'm not sure we actually need this test now. We already test that invalid international numbers raise in `test_phone_number_rejects_invalid_international_values`. Now that `get_international_info` will never raise an exception (only instantiation and validation raise exceptions now), this test is completely redundant.

rparke: I've removed the test, because I consider it redundant. (It doesn't test anything apart from what errors are thrown when you try and parse an invalid number). Can discuss when we re-review if that's needed.

notifications_utils/recipient_validation/phone_number.py
rparke: I'll add the following in a commit message:

"Adds code to handle cases where an international number is incorrectly parsed as a UK number the service doesn't have permission to send to.

There are certain cases where a valid international number is accidentally parsed as a UK landline. In these cases we want to try validating the number, but explicitly as an international number. If this fails, we want to turn the number back into a UK number and raise the correct validation error."

Without this code `tests/test_recipient_csv.py::test_international_recipients` because it thinks that `2304031000` (a valid Mauritius number) is actually meant to be a UK landline (this parses correctly) and then validation fails because the service cannot send to a UK landline. If this gets parsed as `+2304031000` it is correctly parsed as a Mauritus number and the test passes. As far as I'm aware this is an edge case that we want to support, and we can support it without causing other tests to fail.

So the logic goes:

- throw an exception if this is a landline and the service can't send to landlines
- catch the exception and make sure this isn't actually an international number we've mistaken for a landline
- if the error still throws, assume it was meant to be a landline and throw the original exception

notifications_utils/recipient_validation/phone_number.py
rparke: This was just some code I needed to clean up when I was trying to debug an issue. It should all go on one line.

tests/recipient_validation/test_phone_number.py
rparke: We'll do validating and formatting separately in API and Admin as it follows our new pattern. I don't think we should have a function that is equivalent to `try_validate_and_format_phone_number` our new validation flow should be:

```
number = PhoneNumber("ALPHANUM3R1C")
number.validate(<pass international and landline arguments as needed>) # optional, we may or may not want to check service permissions against the number
formatted_number = number.get_normalised_format()
```

Our new code forbids alphanumeric numbers:

```
    @staticmethod
    def _raise_if_phone_number_contains_invalid_characters(number: str) -> None:
        chars = set(number)
        if chars - {*ALL_WHITESPACE + "()-+" + "0123456789"}:
            raise InvalidPhoneError(code=InvalidPhoneError.Codes.UNKNOWN_CHARACTER)
```

We do however check that the human readable and normalised formats works correctly for valid numbers. Which I believe is the only thing we want to make sure works properly for the cases where human readable and normalised formates are needed in API and Admin

tests/recipient_validation/test_phone_number.py
rparke: As here https://github.com/alphagov/notifications-utils/pull/1158#discussion_r1819575413

notifications_utils/recipient_validation/phone_number.py
CrystalPea: Ahh, that makes sense, thank you!

tests/recipient_validation/test_phone_number.py
CrystalPea: > Our new code forbids alphanumeric numbers (and the above code snippet would fail on the first line with "ALPHANUM3R1C" as an argument

But we might be getting an alphanumeric sender for our inbound sms, right? According to [this commit](https://github.com/alphagov/notifications-api/commit/3daf039fbe58e8f517537ce49534e23cc10e6bc5#diff-846b9511257439f27f518ef400cd0c75e74f737087ad044b02b656e6798375d2R33) at least (it is a very old commit, so maybe not anymore? Not sure)

So does it mean API will have its own code that allows alphanumerics for that particular case?

Or is it a case where we thought we might need alphanumerics support back when implementing inbound sms in 2017, but it turned out we don't need them?

rparke: I don't know. I know Leo has is explicitly raise an exception if we try now to it's not like it's an edge case where we didn't realise ... easy enough to add back in if needed though

CrystalPea: Decision record:

After a Slack discussion between Leo, Richard and me, we decided the following:

- we do need to support alphanumeric senders for inbound SMS
- we will move the validate and format functions from utils to admin and API, respectively (since one of the function has been used only by admin, and another was used just by API)
- these functions will parse the numbers by initialising them with PhoneNumber class, but if there is InvalidPhoneError, they will simply return the number, like the functions we are removing form utils do.

risicle (approved):

PR #1160, opened 2024-10-28
logging: don't `calculate_content_length()` for `after_request` logs of streaming responses

risicle:
Adding this (useful) field to the logs causes the response to be eagerly consumed in an attempt to calculate the final length, which negates the point of streaming responses. This may be why we observed our streaming responses apparently "not working" recently.

However simply making them skip the calculation for streaming responses does change behaviour - it means `after_request` logs are now emitted when we return the response status and headers. Which would leave us with no visibility of when the streamed request "ended". So this also adds a new log message for streamed responses being "closed".

Unfortunately from what I can see, flask/werkzeug doesn't really keep a tally of bytes returned in streamed responses, so it's not trivially possible to include a "final" response size in this latter log message.

### DIFF
--- a/tests/test_logging.py
+++ b/tests/test_logging.py
@@ -92,11 +92,13 @@ def test_base_json_formatter_contains_service_id(tmpdir):
         (503, builtin_logging.WARNING, True),
     ),
 )
+@pytest.mark.parametrize("stream_response", (False, True))
 def test_app_request_logs_level_by_status_code(
     app_with_mocked_logger,
     status_code,
     expected_after_level,
     with_request_helper,
+    stream_response,
 ):
     app = app_with_mocked_logger
     app.config["NOTIFY_ENVIRONMENT"] = "foo"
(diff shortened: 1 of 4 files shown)


tests/test_logging.py
```diff
@@ -195,6 +196,62 @@ def some_route():
...
+        mock.call(
+            mock.ANY,
+            AnyStringMatching(r".*streaming response.*", flags=re.I),
+            mock.ANY,
+            extra=mock.ANY,
+        )
+        not in mock_req_logger.log.call_args_list
+    )
+
+    time.sleep(0.05)
```
robinjam, line 209: I'm not sure I understand what this `assert` and `sleep` are for

risicle: The "streaming response closed" log message should only be emitted after the response is consumed/closed, so the assert first checks that it hasn't already been emitted before we've done that.

The sleep is because this test also covers the behaviour of the `request_time` field: the `after_request` log message should have a value of at least 0.05s and then the `call_on_close` log message should be at least 0.05s after that, i.e. at least 0.1s.

robinjam: Cool, makes sense, thanks!

robinjam (approved):
LGTM 👍

PR #1164, opened 2024-11-07
Freeze all test requirements

quis:
Pulling this out of https://github.com/alphagov/notifications-utils/pull/1155 so it’s easier to review and roll out separately.

***

We’re going to roll this out across all the apps, so anyone running the tests in an app will have all the same dependencies as someone else, not just the same app dependencies.

We have to remove `-e .` from `requirements.in` and instead explicitly list `setup.py` in the pip-compile command. Otherwise this line will cause notifications_utils itself to be included in the compiled requirements, which is not something we want.


quis:
@risicle yes, good spot, have fixed, squashed and pushed now

PR #1165, opened 2024-11-11
raise an exception if we cant fetch from github

leohemsted:
currently, it just returns whatever is at the URL (for example an error page). We saw problems on concourse where it just wrote an error page to a requirements_common file and then just tried to carry on happily

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+
+## 89.0.1
+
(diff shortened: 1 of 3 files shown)


robinjam (approved):

quis (approved):
Think I also saw Github rate limiting our IPs earlier today.

Longer term this will be fixed by https://github.com/alphagov/notifications-utils/pull/1155

leohemsted:
ahhh that checks out. probs my fault for triggering so many test runs

at least this'll make knowing what happened easier

PR #1166, opened 2024-11-12
use importlib to improve version_tools scripts

leohemsted:
The mechanism currently used by `version_tools.copy_config()` has a few problems - it relies on:

 - always having network access to github
 - the specific layout of github's url structure
 - the layout of files in the new target version of utils being the same as that expected by the version of `copy_config()` that's running.
 - the new target version of utils being a released, tagged version (which isn't always the case during development)

This changes `copy_config()` to use the "correct" python mechanism for accessing auxiliary files from packages.

it is suggested that apps using utils' requirements_for_test should be using it to freeze their test dependencies, allowing `pip-sync` to be used in their freezing workflow.

---

Note: We removed a call to `pip-sync` from the original version of this PR, as that would _uninstall_ the local development copy of notifications-utils, meaning you can't run version-bump etc etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## 89.1.0
+
+* Change version_tools to a package that uses importlib to grab common reqs/config, rather than fetching common files from github. This repo's usage of those common files are now symlinks to the sources of truth found within notification_utils/version_tools/
 
 ## 89.0.1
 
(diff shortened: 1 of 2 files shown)


quis (commented):
Think it’s `requirements_for_test_common.in` that needs to move, not `requirements_for_test.in`

quis:
This will also need a major version bump

robinjam (approved):

PR #1174, opened 2024-12-03
Install packages using `uv`

quis:
uv is:
- between 10× and 100× faster than pip
- a drop-in replacement for `pip install`, `pip compile`, etc

Because we don’t build a Docker image from this repo we need to put `pip install uv` in the `bootstrap` step instead.


quis:
Closing this because it’s more complicated than other repos (we install packages at the system level on CI which UV doesn’t like to do by default)

PR #1179, opened 2024-12-18
`RequestCache`: add `CacheResultWrapper` to allow dynamic cache decisions

risicle:
Returning a `CacheResultWrapper` from a `RequestCache.set`-wrapped function allows the function to contextually decide whether the result should be cached.

Internally this uses `functools.singledispatch` to detect `CacheResultWrapper` instances, so should allow further customization if the design of `CacheResultWrapper` isn't convenient for some situation.

### DIFF
--- a/notifications_utils/clients/redis/request_cache.py
+++ b/notifications_utils/clients/redis/request_cache.py
@@ -1,16 +1,72 @@
 import json
 from contextlib import suppress
+from dataclasses import dataclass
 from datetime import timedelta
-from functools import wraps
+from functools import singledispatch, wraps
 from inspect import signature
+from typing import TypeAlias
 from uuid import UUID
+
+_JSON: TypeAlias = dict[str, "_JSON"] | list["_JSON"] | str | int | float | bool | None
 
 
 class RequestCache:
     DEFAULT_TTL = int(timedelta(days=28).total_seconds())
 
+    @dataclass
+    class CacheResultWrapper:
+        """
+        Allows the result returned from a `RequestCache.set`-wrapped function to
+        be annotated with a dynamically-determined "decision" on whether this result
+        should be cached or not (and for how long).
+
+        The truthiness of `cache_decision` controls whether a result will be cached.
+
+        Setting `ttl_in_seconds_override` to its default `None` will result in the
+        `ttl_in_seconds` value specified at decoration-time being used.
+
+        Either way, the `value` is extracted by the `RequestCache.set` decorator and
+        the `CacheResultWrapper` is discarded before returning the value on its own
+        (and possibly caching the value).
+        """
+
+        value: _JSON
+        cache_decision: bool
+        ttl_in_seconds_override: int | None = None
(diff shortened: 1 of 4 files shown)


notifications_utils/clients/redis/request_cache.py
```diff
@@ -1,17 +1,56 @@
...
 from inspect import signature
 from uuid import UUID
 
 
 class RequestCache:
     DEFAULT_TTL = int(timedelta(days=28).total_seconds())
 
+    CacheResultWrapper = namedtuple("CacheResultWrapper", ("value", "cache_decision"))
+    CacheResultWrapper.__doc__ = """
+        Allows the result returned from a `RequestCache.set`-wrapped function to
```
leohemsted, line 15: this docstring feels like implementation notes, rather than an actual docstring that describes what the thing is for, and how to use it. in particular, i think it's worth being explicit in the docstring that `value` is the value that will be cached and must be jsondumpable, and cache_decision is a bool

notifications_utils/clients/redis/request_cache.py
```diff
@@ -1,17 +1,56 @@
...
+
+        self.get_cache_value = singledispatch(get_cache_value)
+
+        @self.get_cache_decision.register
+        def _(result: RequestCache.CacheResultWrapper):
+            return result.cache_decision
+
+        @self.get_cache_value.register
+        def _(result: RequestCache.CacheResultWrapper):
+            return result.value
```
leohemsted, line 52: it would be cool to chat at a tech catchup some time about singledispatch - i wasn't familiar with it before so might be fun to let people know how it works

notifications_utils/clients/redis/request_cache.py
```diff
@@ -1,17 +1,56 @@
...
+    CacheResultWrapper.__doc__ = """
+        Allows the result returned from a `RequestCache.set`-wrapped function to
+        be annotated with a dynamically-determined "decision" on whether this result
+        should be cached or not (and for how long).
+
+        Possible values for cache_decision:
+         - True: cache normally with the `ttl_in_seconds` supplied at decoration-time
+         - False: do not cache result
+         - Any other value is used in place of the decoration-time `ttl_in_seconds`
+           and caches the result
```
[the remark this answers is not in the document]
risicle: Yup agreed will do this when I get to it.

notifications_utils/clients/redis/request_cache.py
risicle: I was mainly trying to ensure the behaviour of `cache_decision` was communicated to the potential user of the class.

notifications_utils/clients/redis/request_cache.py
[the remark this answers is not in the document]
risicle: I'm sure we've already got that json type snippet somewhere...

leohemsted (approved):

quis (approved):

PR #1184, opened 2025-01-10
Revert changes in 94.0 to remove old phone number validation code with fix to emergency contact issue

rparke:
We have a fix pending for admin for the emergency contact list issue. This removes the old validation code again and adds a flag `should_validate_phone_number `which lets us toggle validation on/off for CSVs so that we can validate new uploads but be more relaxed when trying to send from old emergency contact lists in S3 that may contain invalid numbers.


quis (approved):
Assuming this has mostly been reviewed already. Addition of new flag looks sensible and is tested 👍🏻

PR #1193, opened 2025-02-06
NotifyTask: emit log message and tally on statsd when task retried

risicle:
Why we were not doing this before I have no idea.


rparke (approved):
If you want to push this out to api/admin will need to revert 95.0.0 because split app fun

risicle:
I don't think there's an immediate urgency for this, but :weary:

PR #1194, opened 2025-02-12
Add lint rule (RUF100) for inapplicable `# noqa`

quis:
Noticed this during a PR review and think it’s better to use a rule (where one exists) rather than a discussion back and forth.

> A noqa directive that no longer matches any diagnostic violations is likely included by mistake, and should be removed to avoid confusion.

– https://docs.astral.sh/ruff/rules/unused-noqa/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 95.1.1
+
+* Add `RUF100` rule to linter config (checks for inapplicable uses of `# noqa`)
+
 # 95.1.0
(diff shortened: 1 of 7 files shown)


klssmith (approved):

risicle:
This seems a bit short-sighted to me, because ruff/black/whatever is far from the only tool that uses `# noqa` for skipping analysis, and this effectively makes any uses of it illegal for anything that ruff doesn't understand. I also don't really understand the value of it (but that goes for any of the formatting stuff)

PR #1196, opened 2025-02-15
eventlet: add `ContextRecyclingEventletWorker`

risicle:
Because eventlet's `GreenPool` discards `GreenThreads` after they have performed a task, to reduce wasteful continual re-creation of thread-local resources this class will maintain a pool of thread contexts suitable for reuse with new `GreenThread`s. In theory at least this means we will never have more thread contexts than the maximum number of concurrent `GreenThread`s handling connections we've ever had.

See https://greenlet.readthedocs.io/en/latest/api.html#greenlet.greenlet.gr_context for documentation on `gr_context`.


CrystalPea (approved):

risicle:
(of course, this doesn't actually set anything to _use_ it yet, that will require further discussion..)

PR #1197, opened 2025-02-19
Implement `InsensitiveSet.__contains__`

quis:
I was surprised that `"first name" in InsensitiveSet(['firstname'])` evaluated to `False`.

This commit makes it work the way you’d expect.

We don’t use this in production code yet, but don’t think it can hurt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 95.2.0
+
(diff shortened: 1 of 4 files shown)


rparke (approved):

risicle:
This is the first time I've noticed `InsensitiveSet` - this certainly doesn't look like the only case where it won't behave correctly - e.g. no mutations after initial construction make any effort to normalize new values.

PR #1199, opened 2025-02-26
Ensure `NotifyTask` is a polymorphic subclass of `Celery.Task`

quis:
If we override methods in our subclass we should ensure they should have the same signature. This commit adds a test which checks for that.

`run` is an exception because it is dynamically created based on the function which the task will call.


klssmith (approved):
Nice idea to add a test for this!

PR #1201, opened 2025-03-04
Replace `setup.py` with `pyproject.toml`

quis:
As of https://peps.python.org/pep-0621/ the Python community has standardised on using `pyproject.toml` for defining a project’s metadata.

In Python 3.12 setuptools (which we use in `setup.py`) won’t be installed by default, so we may want to move towards not using it.

For both reasons this change will future-proof our code.

This commit follows the guidance in https://packaging.python.org/en/latest/guides/modernize-setup-py-project/ (including the part about moving all static metadata to `pyproject.toml`).

It still uses setuptools for reading the version. This can be changed later if we want to drop setuptools entirely.


Makefile
```diff
@@ -19,7 +19,7 @@ test: ## Run tests
 	ruff check .
 	ruff format --check .
 	pytest -n auto
-	python setup.py sdist
+	python -m build
```
leohemsted, line 22: got me thinking - what does this do? (i mean, it packages up the library in to a tar.gz in the dist folder

but do we need this? we don't have any ambitions to distribute this more widely where we'd need to build the package to push to pypi or something, and when we install `pip install notifications-utils @ git+https://<url>.git@1.0.0` that just creates a checkout

can we get rid of this completely?

quis: It helps detect mistakes which would make the package uninstallable.

For example if you remove the `name` line from `pyproject.toml`:
```diff
- name = "notifications-utils"
```

The tests and linters will still pass. But trying to run `pip install -e ../notifications-utils` from one of the apps will result in an error.

Running `python -m build` as part of the tests catches this error before it gets merged – you would see this error:
```
configuration error: `project` must contain ['name'] properties
```

I’m sure there’s another example which is a more likely mistake someone could make, but I can’t think of it right now.

PR #1202, opened 2025-03-04
Fix qrcode with amp

spatel033:
QR codes in letters whose URL includes an ‘&' sign will have it encoded as the '&’ HTML entity. More details on [this card](https://trello.com/c/YESkaUO9/94-qr-codes-with-logos-in-have-encoded-as-amp). There is generic html escape which does not handle the URL use case. So when creating the QR code unescape html to get the original URL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 97.0.1
+
+* Fix QR code URL with '&' encoded as '&amp;'
+
 ## 97.0.0
 
 * for bilingual letters preview, only show barcodes on the first page (patch).
--- a/notifications_utils/markdown.py
+++ b/notifications_utils/markdown.py
@@ -6,7 +6,7 @@ import mistune
 from ordered_set import OrderedSet
 
 from notifications_utils import MAGIC_SEQUENCE, magic_sequence_regex
-from notifications_utils.formatters import create_sanitised_html_for_url, replace_svg_dashes
+from notifications_utils.formatters import create_sanitised_html_for_url, replace_svg_dashes, unescape_strict
 from notifications_utils.qr_code import (
     QR_CODE_MAX_BYTES,
     QrCodeTooLong,
@@ -67,7 +67,7 @@ class NotifyLetterMarkdownPreviewRenderer(mistune.Renderer):
         if "<span class='placeholder" in data or '<span class="placeholder' in data:
             placeholder = qr_code_placeholder(data)
             return replace_svg_dashes(placeholder)
-
+        data = unescape_strict(data)
         qr_data = qr_code_as_svg(data)
         return f"<div class='qrcode'>{replace_svg_dashes(qr_data)}</div>"
 
--- a/notifications_utils/version.py
+++ b/notifications_utils/version.py
@@ -5,4 +5,4 @@
 # - `make version-minor` for new features
 # - `make version-patch` for bug fixes
 
-__version__ = "97.0.0"  # 30cda215ff802d90fad96e1efd98a03b
+__version__ = "97.0.1"  # 00ec821d5d989563632f396ed0dd1b3e
--- a/tests/test_markdown.py
+++ b/tests/test_markdown.py
@@ -656,6 +656,31 @@ def test_letter_qr_code_works_with_extra_whitespace():
             "notifications_utils.markdown.qr_code_as_svg",
             "arbitrary data not a URL",
         ),
+        (
+            "qr: www.gov.uk/search?q=foo&r=bar",
+            "notifications_utils.markdown.qr_code_as_svg",
+            "www.gov.uk/search?q=foo&r=bar",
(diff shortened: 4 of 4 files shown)


notifications_utils/formatters.py
```diff
@@ -178,6 +186,14 @@ def escape_html(value):
     return value
 
 
+def unescape_url(value):
```
quis, line 189: I’d strongly consider using `unescape_strict` from this module, rather than re-implementing a variation of the same thing.

tests/test_template_types.py
```diff
@@ -829,6 +829,26 @@ def test_letter_preview_renders_QR_code_correctly(jinja_template):
     assert jinja_template_locals["message"] == expected_qr_code_svg
 
 
+@mock.patch("notifications_utils.template.LetterPreviewTemplate.jinja_template.render")
+def test_letter_preview_url_with_query_params_renders_QR_code_correctly(jinja_template):
+    str(
+        LetterPreviewTemplate(
+            {
+                "content": "This is your link:\n\nqr: www.gov.uk/search?q=foo&r=bar",
```
quis, line 837: Think this needs parametrizing with a bunch of different content to make sure various characters are all being escaped/unescaped properly

tests/test_template_types.py
```diff
@@ -829,6 +829,26 @@ def test_letter_preview_renders_QR_code_correctly(jinja_template):
...
...m1 0h2m2 0h3m4 0h1m1 0h3m1 0h1m-29 1h1m1 0h3m1 0h1m1 0h3m2 0h2m4 0h1m2 0h1m1 0h3m1 0h1m-29 1h1m5 0h1m1 0h1m2 0h3m4 0h1m1 0h1m1 0h1m5 0h1m-29 1h7m1 0h1m1 0h1m1 0h1m1 0h1m1 0h1m1 0h1m1 0h1m1 0h7m-20 1h1m1 0h1m2 0h2m3 0h1m-19 1h7m1 0h1m2 0h1m1 0h1m2 0h1m1 0h1m3 0h2m3 0h1m-27 1h1m6 0h3m1 0h1m1 0h1m3 0h8m1 0h1m-24 1h2m2 0h2m2 0h3m1 0h2m1 0h1m1 0h2m1 0h2m-26 1h1m1 0h2m2 0h4m1 0h1m1 0h3m3 0h1m1 0h2m1 0h1m2 0h1m-26 1h1m1 0h5m2 0h2m1 0h3m1 0h1m5 0h2m-26 1h2m1 0h1m3 0h1m1 0h2m6 0h3m1 0h4m2 0h1m-29 1h1m1 0h1m2 0h2m2 0h3m4 0h3m1 0h1m1 0h2m1 0h3m-21 1h2m2 0h1m6 0h1m1 0h1m1 0h4m1 0h1m-28 1h2m3 0h4m1 0h1m4 0h2m1 0h1m1 0h2m3 0h4m-29 1h1m6 0h1m3 0h5m2 0h2m1 0h4m3 0h1m-29 1h1m2 0h2m1 0h1m1 0h3m2 0h1m1 0h2m8 0h1m-26 1h1m1 0h1m1 0h1m2 0h1m1 0h1m3 0h3m3 0h1m1 0h4m-25 1h1m4 0h2m1 0h1m1 0h1m1 0h1m1 0h2m1 0h1m2 0h7m1 0h1m-21 1h1m1 0h1m2 0h1m2 0h1m1 0h3m3 0h2m1 0h2m-29 1h7m1 0h1m3 0h1m1 0h2m2 0h3m1 0h1m1 0h3m-27 1h1m5 0h1m1 0h3m7 0h3m3 0h1m3 0h1m-29 1h1m1 0h3m1 0h1m1 0h2m1 0h3m1 0h1m2 0h1m1 0h5m1 0h2m-28 1h1m1 0h3m1 0h1m1 0h2m3 0h4m1 0h1m1 0h2m3 0h2m-27 1h1m1 0h3m1 0h1m1 0h1m3 0h2m4 0h8m1 0h1m-28 1h1m5 0h1m1 0h2m1 0h2m3 0h1m2 0h2m4 0h1m1 0h1m-28 1h7m2 0h6m1 0h2m1 0h1m1 0h3m2 0h1"/></svg></div></p>'  # noqa
```
quis, line 848: It’s really hard to tell if this test is doing the right thing. If you mocked `notifications_utils.qr_code.qr_code_as_svg` then you’d be able to assert that it’s getting called with the correct URL data

quis (changes requested):

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 97.0.1
+
+* Implement `formatters.unescape_url`
```
quis, line 5: This is outdated (sorry for not catching this earlier)
```suggestion
```

quis (changes requested):

quis (approved):

PR #1203, opened 2025-03-04
[bug fix] Only show barcode on first page of letter preview

CrystalPea:
For bilingual letters, only show the barcodes on the Welsh first page, and not on the English first page - as this is how it will look when printed.

Review commit by commit.

Trello ticket: https://trello.com/c/Hr4MDKlM/152-barcodes-show-on-pages-after-the-1st-on-bilingual-letters

Related template-preview PR: https://github.com/alphagov/notifications-template-preview/pull/872

|BEFORE|AFTER|
|:---|:---|
|Screenshot 2025-03-04 at 14 42 34|Screenshot 2025-03-04 at 14 39 39|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 97.0.0
+
+* for bilingual letters preview, only show barcodes on the first page (patch).
+* rename `include_notify_tag` argument to `includes_first_page` (major).
+
 ## 96.0.0
 
 * BREAKING CHANGE: the `gunicorn_defaults` module has been moved to `gunicorn.defaults` to make space for gunicorn-related utils that don't have the restricted-import constraints of the gunicorn defaults. Imports of `notifications_utils.gunicorn_defaults` should be changed to `notifications_utils.gunicorn.defaults`.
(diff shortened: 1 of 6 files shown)


notifications_utils/version.py
```diff
@@ -5,4 +5,4 @@
 # - `make version-minor` for new features
 # - `make version-patch` for bug fixes
 
-__version__ = "95.2.0"  # e423953f9dec74010897c62aa4e035de
+__version__ = "95.2.1"  # cca28798f43ed1eba5cb4b4a4b68797f
```
quis, line 8: If you’re changing the name of an argument this is a breaking change

CHANGELOG.md
```diff
@@ -423,7 +423,7 @@ Removes Emergency-Alerts-related code
 
 ## 73.2.0
 
-* Adds a `include_notify_tag` parameter to `LetterPrintTemplate` so that bilingual letters can disable the NOTIFY tag on the English pages of a letter.
+* Adds a `includes_first_page` parameter to `LetterPrintTemplate` so that bilingual letters can disable the NOTIFY tag on the English pages of a letter.
```
leohemsted, line 431: dont think we should be changing old changelogs? 🤔

CrystalPea: oops, good spot!

notifications_utils/version.py
CrystalPea: Done ☺️

quis (approved):

PR #1204, opened 2025-03-05
Fix external link without http

spatel033:


spatel033:
https://github.com/alphagov/notifications-utils/pull/1202 changes will be deployed first

notifications_utils/markdown.py
```diff
@@ -229,6 +229,10 @@ def block_quote(self, text):
     def link(self, link, title, content):
         if link.startswith("span class='placeholder") and link.endswith("</span"):
             link = f"<{link}>"
+
+        if not link.lower().startswith("http"):
+            link = f"https://{link}"
```
quis, line 234: Looks like we already do this for text messages so don’t need to worry about that: https://github.com/alphagov/notifications-utils/blob/a82c8541b709a735ed11b49fea92d99216d98c1d/notifications_utils/formatters.py#L131-L132

quis (approved):

notifications_utils/markdown.py
spatel033: I was not sure to update the logic for `create_sanitised_html_for_url`. I have refactored `create_sanitised_html_for_url` to handle the external link with title.

PR #1206, opened 2025-03-14
Fix inset text bottom margin in emails

kr8n3r:
Addresses https://trello.com/c/7vrT4qfU/104-inset-text-sometimes-has-no-margin-bottom

To reliably have bottom margin applied across clients, bottom margin had to be moved to a wrapping div element.

**Before**
https://app.emailonacid.com/shared-preview/YNbYn8N9oO

**After**
https://app.emailonacid.com/shared-preview/IzCFf5SNC8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 97.0.4
+
+* Fix inset text block styling
+
 ## 97.0.3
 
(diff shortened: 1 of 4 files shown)


tombye (approved):
Nice work, looks good to me 👍🏻

tombye (approved):
Tested with API sending an email and works as expected. A long time coming so top marks for fixing this! 🚀 👍🏻

PR #1207, opened 2025-03-26
Add s3 multipart upload

spatel033:
Add multipart upload for large file handling. Implement full multipart upload lifecycle (create, upload, complete, abort).
Documentation here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-using-multipart-upload.html

This PR is part of the notifications request report data processing. Here is the [card](https://trello.com/c/sRiFZcNB/1197-function-report-data-processing) for more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 99.2.0
+
+* Add s3 multipart upload lifecycle (create, upload, complete, abort)
+
 ## 99.1.2
 
 * Normalise unusual dashes and spacing characters in phone numbers
--- a/notifications_utils/s3.py
+++ b/notifications_utils/s3.py
@@ -1,7 +1,7 @@
 import urllib
 
 import botocore
-from boto3 import resource
+from boto3 import client, resource
 from flask import current_app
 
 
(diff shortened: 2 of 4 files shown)


klssmith:
I wonder if this would be better in notifications-api, since I'm not sure we have any plans to use this functionality in other repos. We keep shared code here, but have also have app specific S3 code in other repos (e.g. [here in api](https://github.com/alphagov/notifications-api/blob/main/app/aws/s3.py) and [here in admin](https://github.com/alphagov/notifications-admin/tree/main/app/s3_client)). Any thoughts on having it in api vs here?

spatel033:
> I wonder if this would be better in notifications-api, since I'm not sure we have any plans to use this functionality in other repos. We keep shared code here, but have also have app specific S3 code in other repos (e.g. [here in api](https://github.com/alphagov/notifications-api/blob/main/app/aws/s3.py) and [here in admin](https://github.com/alphagov/notifications-admin/tree/main/app/s3_client)). Any thoughts on having it in api vs here?

I noticed that other repositories only handle getting objects from S3, but uploading to S3 is done using the `s3upload` utils. So, I thought to add the multipart upload logic to the utils.

joybytes:
> > I wonder if this would be better in notifications-api, since I'm not sure we have any plans to use this functionality in other repos. We keep shared code here, but have also have app specific S3 code in other repos (e.g. [here in api](https://github.com/alphagov/notifications-api/blob/main/app/aws/s3.py) and [here in admin](https://github.com/alphagov/notifications-admin/tree/main/app/s3_client)). Any thoughts on having it in api vs here?
> 
> I noticed that other repositories only handle getting objects from S3, but uploading to S3 is done using the `s3upload` utils. So, I thought to add the multipart upload logic to the utils.

I dislike us having code spread through different repos, but that's a personal opinion.

If we already have a pattern in the utils repo for this, it makes sense to keep it.

notifications_utils/s3.py
```diff
@@ -45,3 +45,66 @@ def s3download(bucket_name, filename):
...
+        response = s3.upload_part(
+            Bucket=bucket_name,
+            Key=filename,
+            PartNumber=part_number,
+            UploadId=upload_id,
+            Body=data_bytes,
+        )
+        return response
+    except botocore.exceptions.ClientError as e:
+        current_app.logger.error("Unable to upload part to S3 bucket %s", bucket_name)
```
joybytes, line 85: is it useful to include part number on the logs too?

joybytes: I think so, not that I expect us to run into this issue but I think the cap for parts is 10,000 so it would be nice to know if we ever exceed in errors

notifications_utils/s3.py
```diff
@@ -45,3 +45,66 @@ def s3download(bucket_name, filename):
         return key.get()["Body"]
     except botocore.exceptions.ClientError as error:
         raise S3ObjectNotFound(error.response, error.operation_name) from error
+
```
joybytes, line 48: could we also include the upload_id in the logs in addition to bucket_name please

notifications_utils/s3.py
```diff
@@ -45,3 +45,66 @@ def s3download(bucket_name, filename):
...
+
+def s3_multipart_upload_create(
+    bucket_name, file_location, tags=None, metadata=None, content_type="binary/octet-stream"
+):
+    s3 = client("s3")
+
+    args = {"Bucket": bucket_name, "Key": file_location, "ServerSideEncryption": "AES256", "ContentType": content_type}
+
+    if tags:
```
klssmith, line 57: If we have no plans to use tags or metadata, should we remove the code relating to them? It makes things more complex, and can always be added later if necessary, but there are no mentions of using these on the RFC

notifications_utils/s3.py
spatel033: Updated [here](https://github.com/alphagov/notifications-utils/pull/1207/files#diff-454970afc2b70a52d6854993f5478578713f2623a11f8c7785b7d2296949c8ddR92)

joybytes (approved):
looks good to me

notifications_utils/s3.py
```diff
@@ -45,3 +45,60 @@ def s3download(bucket_name, filename):
...
+        response = s3.upload_part(
+            Bucket=bucket_name,
+            Key=filename,
+            PartNumber=part_number,
+            UploadId=upload_id,
+            Body=data_bytes,
+        )
+        return response
+    except botocore.exceptions.ClientError as e:
+        current_app.logger.error("Unable to upload part %s to S3 bucket %s", part_number, bucket_name)
```
leohemsted, line 79: i think this log should include the filename too. maybe the upload ID as well, i'm not sure.

notifications_utils/s3.py
```diff
@@ -45,3 +45,60 @@ def s3download(bucket_name, filename):
...
+def s3_multipart_upload_create(bucket_name, file_location, content_type="binary/octet-stream"):
+    s3 = client("s3")
+
+    args = {"Bucket": bucket_name, "Key": file_location, "ServerSideEncryption": "AES256", "ContentType": content_type}
+
+    try:
+        response = s3.create_multipart_upload(**args)
+        return response
+    except botocore.exceptions.ClientError as e:
+        current_app.logger.error("Unable to create multipart upload to S3 bucket %s", bucket_name)
```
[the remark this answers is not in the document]
leohemsted: also if we make all three of these logs `logger.exception` they'll include stack traces

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 98.0.2
+
+* Add s3 multipart upload lifecycle (create, upload, complete, abort)
```
leohemsted, line 5: this should be a minor version, not a patch version. it's got a new feature

leohemsted (approved):

klssmith (approved):

PR #1209, opened 2025-03-31
Update international SMS rate multipliers

klssmith:
The rate multipliers for international SMS are changing from 1 April
2025, so this change updates the yml file. The `+44` prefix has a rate
multiplier of 2 for Guersey, Jersey and the Isle of Man and 1 for
mainland UK, which requires a change in logic.

### DIFF
--- a/notifications_utils/international_billing_rates.yml
+++ b/notifications_utils/international_billing_rates.yml
@@ -55,7 +55,6 @@
   names:
   - Canada
   - United States
-  - Dominican Republic
 '7':
   attributes:
     alpha: REG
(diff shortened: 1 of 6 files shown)


notifications_utils/international_billing_rates.yml
```diff
@@ -1670,7 +1670,7 @@
     sc: null
     sender_and_registration_info: null
     text_restrictions: null
-  rate_multiplier: 3
+  rate_multiplier: 8
   names:
   - San Marino, Republic of
```
spatel033, line 1675: can't find in the list

klssmith: Yes good spot, I gave this 8 because the list has a row for "All others" which is 8

notifications_utils/international_billing_rates.yml
```diff
@@ -2253,7 +2252,7 @@
     sc: null
     sender_and_registration_info: null
     text_restrictions: null
-  rate_multiplier: 4
+  rate_multiplier: 8
   names:
   - Niue
```
spatel033, line 2257: `Niue` is missing in list

notifications_utils/international_billing_rates.yml
```diff
@@ -2084,7 +2083,7 @@
     sc: 'YES'
     sender_and_registration_info: null
     text_restrictions: null
-  rate_multiplier: 3
+  rate_multiplier: 7
   names:
   - Curacao (former Netherlands Antilles)
```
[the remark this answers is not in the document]
klssmith: This is one that we're waiting to hear back about. It shares the `599` prefix with the Caribbean Netherlands which has a rate multiplier of 7 in the spreadsheet. Since they share the same dialling code and we don't have a way of differentiating the mobile numbers, we've decided to go for the highest rate multiplier for now

notifications_utils/international_billing_rates.yml
klssmith: This value comes from the default of 8 for "All others" (same as Tokelau below)

tests/test_international_billing_rates.py
```diff
@@ -30,7 +30,7 @@ def test_international_billing_rates_are_in_correct_format(country_prefix, value
 
 
 def test_country_codes():
-    assert len(COUNTRY_PREFIXES) == 220
+    assert len(COUNTRY_PREFIXES) == 223
```
spatel033, line 33: We have 231 countries in the list. Does this change cover all of them (except same country codes)?

klssmith: There are one or two tiny countries / areas on the new list that we don't have in our file, such as Saint Helena. We can decide whether to add them later, but we'd want a bit more information before doing that because we also would also want to add details about sending restrictions. Not adding them now keeps the functionality the same.

There are countries that are shown split up on the new pricing list, but that share a dialling code so are shown as one item in our file. And the Dominican Republic is now split up in our file, but is one line in the new pricing list.

Basically, it's complicated and the two data sources don't match exactly, but everything in our yml file should either be in the new list of prices or set to `8`, the default.

spatel033 (approved):

joybytes (approved):

PR #1214, opened 2025-04-15
Extend Placeholder Syntax to include unsafe values

rparke:
Best reviewed commit by commit.

Adds a placeholder of with the form `<my value>::unsafe>` that can be used to sanitise values that may contain vulnerable markdown syntax.

This will have the placeholder return a stub until later PRs can add the sanitisation logic. So, `str(Placeholder("((my name::unsafe)), {"my_name":"Richard"}))` is evaluated as `SANITISED` .

### DIFF
--- a/notifications_utils/field.py
+++ b/notifications_utils/field.py
@@ -1,6 +1,5 @@
 import re
 from enum import StrEnum, auto
-from collections import OrderedDict
 
 from markupsafe import Markup
 from ordered_set import OrderedSet
@@ -20,18 +19,21 @@ class Placeholder:
 
     class Types(StrEnum):
         BASE = auto()
+        UNSAFE = auto()
         CONDITIONAL = auto()
 
-    extended_type_pattern = OrderedDict(
-        {
-            Types.CONDITIONAL: "??",
-        }
-    )
+    # order matters as the placeholder type will be the first type that returns a match
+    # conditional shoud come first because no escaping should happen on conditional
+    # placeholders as they don't contain vulnerable input
+    extended_type_pattern = {
+        Types.CONDITIONAL: re.compile(r".*\?\?.*"),
+        Types.UNSAFE: re.compile(r".*::unsafe$"),
+    }
 
     @property
     def type(self):
         for type, pattern in self.extended_type_pattern.items():
-            if pattern in self.body:
+            if re.match(pattern, self.body):
                 return type
         return self.Types.BASE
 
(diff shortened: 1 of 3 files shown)


tests/test_field.py
```diff
@@ -177,6 +202,10 @@ def test_optional_redacting_of_missing_values(template_content, data, expected):
             "((warning?? This is a warning\n text after linebreak))",
             "<span class='placeholder-conditional'>&#40;&#40;warning??</span> This is a warning\n text after linebreak&#41;&#41;",  # noqa
         ),
+        (
+            "Unsafe placeholder: ((name::unsafe))",
+            "Unsafe placeholder: <span class='placeholder'>&#40;&#40;name::unsafe&#41;&#41;</span>",
```
rparke, line 207: not a frontend dev, I don't know if we want to extend the css here like we do with conditional placeholders to just highlight the `Placeholder.name` part.

What we currently have for conditional placeholders:

Screenshot 2025-04-16 at 15 24 24
Screenshot 2025-04-16 at 15 24 37

notifications_utils/field.py
```diff
@@ -16,17 +18,45 @@ def __init__(self, body):
         # body shouldn’t include leading/trailing brackets, like (( and ))
         self.body = body.lstrip("(").rstrip(")")
 
+    class Types(StrEnum):
+        BASE = auto()
+        UNSAFE = auto()
+        CONDITIONAL = auto()
+
+    extended_type_pattern = OrderedDict(
```
quis, line 26: `dict`s are always ordered [as of Python 3.7](https://mail.python.org/pipermail/python-dev/2017-December/151283.html) so `OrderedDict` [is not needed](https://github.com/alphagov/notifications-utils/pull/958)

notifications_utils/field.py
```diff
@@ -16,17 +18,45 @@ def __init__(self, body):
...
+        {
+            Types.UNSAFE: "::unsafe",
```
quis, line 28: This will match:
- `((example::unsafefoobar))` (should probably be treated as a normal placeholder)
- `((example??foo::unsafe))` (should probably be treated as a conditional placeholder)

notifications_utils/field.py
rparke: removed - just a standard dict now

notifications_utils/field.py
rparke: type attribute now matches on a regex, added additional tests to check against this.

notifications_utils/field.py
```diff
@@ -16,17 +18,38 @@ def __init__(self, body):
         # body shouldn’t include leading/trailing brackets, like (( and ))
         self.body = body.lstrip("(").rstrip(")")
 
+    class Types(StrEnum):
+        BASE = auto()
+        CONDITIONAL = auto()
+
+    extended_type_pattern = OrderedDict(
```
CrystalPea: I'd add a comment here about why the order of the items matters - that is it will determine the order we check for the types, and so the conditional should come first, from what I gather?

CrystalPea (approved):
Nice work!

One comment to add for the benefit of future people working with this code, but other than that, this is ready to go in 😁

PR #1216, opened 2025-04-17
Use non-deprecated format for specifying project license

quis:
`project.license` as a TOML table is deprecated

Seeing this warning locally:

> Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).
>
> By 2026-Feb-18, you need to update your project and remove deprecated calls or your builds will no longer be supported.

https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

New format taken from https://spdx.org/licenses/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ name = "notifications-utils"
 dynamic = ["version"]
 description = "Shared python code for GOV.UK Notify."
 readme = "README.md"
-license = {text = "MIT License"}
+license = "MIT"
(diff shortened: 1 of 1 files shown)


leohemsted (approved):

leohemsted:
does this need a version bump? or just gonna roll this in to the next release?

quis:
Just let it roll – all it does is silences some warnings when running `python -m build`

PR #1218, opened 2025-04-23
`celery.NotifyTask`: don't emit early log if called synchronously

risicle:
Fixup for #1213. This is a misleading log message to emit as we're not an invocation of the stated celery task in this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 99.3.4
+
+* `celery.NotifyTask`: don't emit early log if called synchronously
+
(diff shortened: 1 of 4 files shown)


leohemsted (commented):
needs a version bump?

quis:
https://github.com/alphagov/notifications-utils/blob/84b675433fdd7c17c1eee6e546120aa3a4e5887b/README.md#L43

risicle:
This policy means it's effectively impossible to make any change that isn't going to create a merge conflict once you merge it.

quis (approved):

PR #1219, opened 2025-04-24
Add celery json logging

corlettb:
What
----

Add celery json logging for critical and warning errors.

Why
----

Currently in our celery workers critical celery errors are not getting logged.


notifications_utils/logging/celery.py
```diff
@@ -0,0 +1,52 @@
...
+            "category": category.__name__,
+            "filename": filename,
+            "lineno": lineno,
+        }
+        logger.warning(str(message), log_entry)
+
+    warnings.showwarning = custom_showwarning
+
+
+@setup_logging.connect
```
risicle, line 25: Even though it's the documented way of doing this, I don't like things implicitly doing things on import - i'd much rather instead we did `setup_logging.connect(setup_logging_connect)` from inside `set_up_logging`

tests/logging/test_celery.py
```diff
@@ -0,0 +1,106 @@
...
+    # Set the environment variable
+    env = os.environ.copy()
+    env["CELERY_LOG_LEVEL"] = "INFO"
+
+    # Change the working directory to the mock folder
+    mock_dir = os.path.join(os.path.dirname(__file__), "mock")
+
+    try:
+        # Start the Celery worker process
+        process = subprocess.Popen(
```
risicle, line 69: :blink:

notifications_utils/logging/celery.py
risicle: e.g. I can imagine someone getting confused, "but I commented out the set_up_logging call!"

corlettb: I agree. Unfortunately if I put the the:

```
@setup_logging.connect
def setup_logging_connect(*args, **kwargs):
```

inside the set_up_logging function it no longer works. I can only assume that celery is doing stuff on import.

You got any other ideas on how to do this that I might not have tried?

tests/logging/test_celery.py
corlettb: I don't like it either. I could not think of any other way to test it. I figured this is better than nothing.

notifications_utils/logging/celery.py
robinjam: I also agree, but I think it's just the way Celery works - if you don't register a listener on the setup_logging signal then Celery sets up its default logging automatically (presumably on import).

https://docs.celeryq.dev/en/stable/userguide/signals.html#setup-logging

tests/logging/test_celery.py
```diff
@@ -0,0 +1,106 @@
...
+            command,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+            env=env,
+            cwd=mock_dir,
+        )
+
+        process.wait(timeout=10)
```
robinjam, line 78: Why do we call `wait` before `communicate` here?

Just wondering what happens if the buffer for stdout or stderr fills up here, presumably we get a timeout error?

tests/logging/test_celery.py
robinjam: I can't think of a better way to do it either...

tests/logging/test_celery.py
corlettb: You are right. Not ideal. Updated.

notifications_utils/logging/celery.py
robinjam: I don't feel that strongly about it, I think it's fine the way you've done it

(I suspect `after_setup_logger` won't work for the same reason: because you're registering the listener after the signal has already fired)

risicle: If you leave `def setup_logging_connect(...` in the same place and instead of adding the decorator, add a line to `set_up_logging` that does `setup_logging.connect(setup_logging_connect)`, it should be near-functionally-identical to using it as a decorator (apart from it not having import-time side effects) given the way `celery_dummy.py` calls `set_up_logging` only a few lines after importing this module.

risicle:
Have pushed a commit that assembles a temp dir of symlinks as the dummy celery app dir.

notifications_utils/logging/celery.py
corlettb: The test is passing this way. Nice.

corlettb:
> Have pushed a commit that assembles a temp dir of symlinks as the dummy celery app dir.

Rebased

risicle (approved):
Let's give it a go

PR #1221, opened 2025-04-30
For `RecipientCSV` class, count international text messages

CrystalPea:
This is so that we can effectively check international sms message limit for jobs.

### DIFF
--- a/notifications_utils/recipients.py
+++ b/notifications_utils/recipients.py
@@ -122,6 +122,7 @@ class RecipientCSV:
             or self.too_many_rows
             or (not self.allowed_to_send_to)
             or any(self.rows_with_errors)
+            or self.more_international_sms_than_can_send
         )  # `or` is 3x faster than using `any()` here
 
     @property
--- a/notifications_utils/recipients.py
+++ b/notifications_utils/recipients.py
@@ -63,6 +63,8 @@ class RecipientCSV:
         self.rows_as_list = None
         self.should_validate = should_validate
         self.should_validate_phone_number = should_validate_phone_number
+        # use the international_sms_count property to get this.
+        self._international_sms_count = 0
 
     def __len__(self):
         if not hasattr(self, "_len"):
@@ -129,6 +131,12 @@ class RecipientCSV:
         if not self.guestlist:
             return True
         return all(allowed_to_send_to(row.recipient, self.guestlist) for row in self.rows)
+
+    @property
+    def international_sms_count(self):
+        if self._international_sms_count == 0:
+            self.rows
+        return self._international_sms_count
 
     @property
     def rows(self):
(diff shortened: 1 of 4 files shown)


tests/test_recipient_csv.py
```diff
@@ -738,16 +739,18 @@ def test_bad_or_missing_data(
...
 )
-def test_international_recipients(file_contents, rows_with_bad_recipients):
+def test_international_recipients(file_contents, rows_with_bad_recipients, expectected_count):
     recipients = RecipientCSV(
         file_contents,
         template=_sample_template("sms"),
         allow_international_sms=True,
     )
     assert _index_rows(recipients.rows_with_bad_recipients) == rows_with_bad_recipients
+    assert recipients.international_sms_count == expectected_count
```
quis, line 753: This assertion only passes if the line above is run. Otherwise `international_sms_count` is always `0`. This feels fragile.

If you look at how this works for the existing limit then the class has a property which returns `True`/`False` and does the counting only when it’s needed:
https://github.com/alphagov/notifications-utils/blob/bf94e62d250b12fa0e5dfe50e50ff06ae5f831eb/notifications_utils/recipients.py#L189-L191

CrystalPea: Great catch, thank you! I made an update that fixes this issue, and moved the assertion up in the test.

quis (commented):
It would also be good if `has_errors` returned `True` when a file contains too many international phone numbers. It works this way for the regular limits (via `more_rows_than_can_send`):
https://github.com/alphagov/notifications-utils/blob/bf94e62d250b12fa0e5dfe50e50ff06ae5f831eb/notifications_utils/recipients.py#L114-L123

notifications_utils/recipients.py
```diff
@@ -130,6 +134,21 @@ def allowed_to_send_to(self):
             return True
         return all(allowed_to_send_to(row.recipient, self.guestlist) for row in self.rows)
 
+    @cached_property
+    def international_sms_count(self):
+        if self.template_type != "sms":
+            return 0
+        return sum(self._international_sms_count_generator())
+
+    def _international_sms_count_generator(self):
```
CrystalPea, line 143: isn't having a separate generator much less efficient than getting the count from errors function? That is - we now have to check each phone number twice - once for the other errors, and once to check the international limit, when it could all be done in one go?

CrystalPea (commented):
Hey,

Nice work, one comment around performance.

Someone else will have to approve for us either way, since Github won't let me.

notifications_utils/recipients.py
BlessedDev: > isn't having a separate generator much less efficient than getting the count from errors function? That is - we now have to check each phone number twice - once for the other errors, and once to check the international limit, when it could all be done in one go?

I think the generator could work better because this evaluation for international sms would be independent of the other error evaluations. 
Also it has the benefit of lazy evaluation which could also be beneficial because we are potentially dealing with a large data set and it would be more memory efficient.

BlessedDev (approved):

notifications_utils/recipients.py
CrystalPea: Wow, amazing, well done Chris! And also the magic of teamwork 😁

BlessedDev (approved):

PR #1223, opened 2025-04-30
Remove urls from unsafe placeholders

rparke:
For the first iteration of sanitising placeholders we decide that if a replacement value contains a url, the entire value should be considered compromised and an empty string returned:

### DIFF
--- a/notifications_utils/field.py
+++ b/notifications_utils/field.py
@@ -8,6 +8,7 @@ from notifications_utils.formatters import (
     escape_html,
     strip_and_remove_obscure_whitespace,
     unescaped_formatted_list,
+    url,
 )
 from notifications_utils.insensitive_dict import InsensitiveDict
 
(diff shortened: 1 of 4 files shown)


notifications_utils/field.py
```diff
@@ -166,10 +166,23 @@ def replace_match(self, match):
...
-            return "SANITISED"
+            return self.sanitise_replacement_unsafe(replacement)
 
         return replacement
 
+    def sanitise_replacement_unsafe(self, replacement: str):
+        # if the replacement contains a link consider it all compromised
+        html_pattern = re.compile(
+            r'https?://\S+|www\.\S+'
+        )
```
quis, line 177: Consider reusing `formatters.url`:

https://github.com/alphagov/notifications-utils/blob/bf94e62d250b12fa0e5dfe50e50ff06ae5f831eb/notifications_utils/formatters.py#L47-L55

rparke: ah cool I didn't realise we have that!

PR #1237, opened 2025-05-20
Celery Logging Update

corlettb:
https://trello.com/c/btX99C0D/1085-celery-workers-not-logging-critical-errors

What
----

- Remove the warnings override. Feels a bit dodgy. It only means we miss a few items on startup
- Allow celery beat and celery worker log levels to be specified independently and from our normal config setup. 
- Update tests to include celery beat log tests
- Add test to ensure we do not print celery worker log messages if log level is CRITICAL.

Why
----

Previous implementation lacked beat testing and was using a non-standard configuration

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+.venv
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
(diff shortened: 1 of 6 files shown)


risicle:
> I thought `CRITICAL` were the ones we _did_ want 😕

Or perhaps by "worker logs" you meant the boring, lower-level worker logs, which indeed we don't want.

risicle (approved):

PR #1244, opened 2025-06-09
Specify version of dateutil

quis:
Because this is one of our direct dependencies, we should specify it, rather than rely on being pressent because it’s a subdependency.

This also means we can specify the version, and we want the latest version for Python 3.12 compatibility – see https://github.com/dateutil/dateutil/issues/1284


BlessedDev (approved):
This is timely and support is also being worked on for `python 3.13`. 
https://github.com/dateutil/dateutil/blob/35ed87a02ad777428a9da84ead4d8425b533d4a9/.github/workflows/validate.yml#L24

PR #1248, opened 2025-07-04
Otel Common Code

corlettb:
https://trello.com/c/mmrH8QbC/1154-turn-second-otel-spike-into-a-pr-we-can-merge

What
----

Add otel common code.

The different otel instrumentation are implemented as optional dependencies. 

Add a common config class that can be shared by applications to pick up common configuration.

We are not doing anything on the otel logging front yet.

Why
----

This is part of our push to get into more standard observability.


notifications_utils/clients/otel/otel_client.py
```diff
@@ -0,0 +1,179 @@
...
+        {"service.name": os.getenv("NOTIFY_APP_NAME") or app.config.get("NOTIFY_APP_NAME") or "notifications"}
+    )
+
+    provider = MeterProvider(metric_readers=metric_readers, resource=resource)
+    metrics.set_meter_provider(provider)
+
+    set_tracer_provider(TracerProvider(resource=resource))
+
+    def regex_predicate(baggage_key: str) -> bool:
+        return baggage_key.startswith("^public-.+")
```
risicle, line 66: This could do with some explanation. Why is it checking whether a string of some sort begins with a literal regex pattern? If it was cribbed from https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/processor/opentelemetry-processor-baggage#usage then I think that might be wrong - the code shows no special handling around regexes, it's just a pure string https://github.com/open-telemetry/opentelemetry-python-contrib/blob/80c357bb16b42c76763d85441e53dd4f6087d6c5/processor/opentelemetry-processor-baggage/src/opentelemetry/processor/baggage/processor.py#L61

corlettb: Previously I add all baggage getting added as an attribute. I thought that might be bad if we started passing PII in baggage so I added a regex. I didn't check to see if it works. I probably should have.

I could just remove the whole BaggageSpanProcessor if you want.

It's a case of me try to guess what might be useful to someone in the future rather than it being useful now.

notifications_utils/clients/otel/utils.py
```diff
@@ -0,0 +1,175 @@
...
+# Example 3: Dynamic attributes using both args and kwargs
+# @otel_span(attributes=lambda args, kwargs: {
+#     "first_arg": args[0] if args else None,
+#     "keyword": kwargs.get("keyword")
+# })
+# def do_something(a, keyword=None):
+#     ...
+
+
+def otel_span(
```
risicle, line 49: FWIW Instead of implementing decorators directly, nowadays I tend to use https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager which automagically generates both a decorator and context manager so you can use it as either.

risicle: Though come to think I'm not sure it would allow you to inspect the args & kwargs of the inner function.

risicle: But If we don't have any cases where we're actually _using_ the "callable attrs" yet (don't think I've spotted one in the apps), could we, for now, do away with this and instead just use `otel_span_with_status` seeing as it should work as a decorator too. The only remaining difference I can see is the latter expects a `tracer` to be injected and this conjures up its own. Is that really unreconcilable?

corlettb: I'll go with what you want on this front.

I created these functions to try and minimise the repetitive code in api.

The code actually using these functions changed quite a bit over time. I left them with their full functionality, even though not everything currently uses that to try and future proof them a bit.

So up to you really.

notifications_utils/clients/otel/otel_client.py
```diff
@@ -0,0 +1,179 @@
...
+
+    CeleryInstrumentor().instrument()
+
+
+def _instrument_flask(app: Flask) -> None:
+    from opentelemetry.instrumentation.flask import FlaskInstrumentor
+
+    FlaskInstrumentor().instrument_app(
+        app,
+        excluded_urls=app.config.get("OTEL_PYTHON_FLASK_EXCLUDED_URLS", "_status(\\?.*)?$,/metrics"),
```
risicle, line 123: Is this env var only heeded by us? Might it be clearest if we prefix all the ones that are just consumed by our own code with `NOTIFY_`, even if it means we have a bunch of `NOTIFY_OTEL_..` env vars?

notifications_utils/clients/otel/otel_client.py
corlettb: I've fixed the startswith. Let me know if you rather I removed entirely.

notifications_utils/clients/otel/utils.py
corlettb: I've removed otel_span

notifications_utils/clients/otel/utils.py
```diff
@@ -0,0 +1,175 @@
...
+# Example 3: Dynamic attributes using both args and kwargs
+# @otel_histogram("do_something_duration", attributes=lambda args, kwargs: {
+#     "first_arg": args[0] if args else None,
+#     "keyword": kwargs.get("keyword")
+# })
+# def do_something(a, keyword=None):
+#     ...
+
+
+def otel_histogram(
```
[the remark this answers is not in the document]
corlettb: I've have renamed

notifications_utils/clients/otel/otel_client.py
risicle: No I think you've taken this too far. I just asked if we wanted _environment variable_ names that _we've invented_ to be prefixed with `NOTIFY_`. Certainly don't want e.g. things named `NOTIFY_OTEL_METRIC_EXPORT_INTERVAL` when otel natively consumes `OTEL_METRIC_EXPORT_INTERVAL`

risicle: It doesn't make any sense. Maybe pretend I didn't say anything.

corlettb: Okay I've removed the 'NOTIFY_'.

pyproject.toml
```diff
@@ -31,8 +31,27 @@ dependencies = [
...
+    "opentelemetry-instrumentation-redis==0.54b1",
+    "opentelemetry-instrumentation-sqlalchemy==0.54b1",
+    "opentelemetry-instrumentation-botocore==0.54b1"
+]
+otel-flask = [ "opentelemetry-instrumentation-flask==0.54b1" ]
+otel-celery = [ "opentelemetry-instrumentation-celery==0.54b1" ]
+otel-requests = [ "opentelemetry-instrumentation-requests==0.54b1" ]
+otel-redis = [ "opentelemetry-instrumentation-redis==0.54b1" ]
+otel-sqlalchemy = [ "opentelemetry-instrumentation-sqlalchemy==0.54b1" ]
+otel-botocore = [ "opentelemetry-instrumentation-botocore==0.54b1" ]
```
risicle, line 53: Pinned versions? What's the plan for keeping these updated?

notifications_utils/clients/otel/otel_client.py
robinjam: I think I'd prefer `ALLOW_ALL_BAGGAGE_KEYS` - if we ever put PII in baggage then we've already screwed up and I would prefer to catch that during code review rather than relying on this regex to catch it

PR #1249, opened 2025-07-07
major version bump to 100.0.0

rparke:
We've made breaking changes since 99.8.0 that won't be correctly tagged (and so importable) until version.py is bumped as well

### DIFF
--- a/notifications_utils/version.py
+++ b/notifications_utils/version.py
@@ -5,4 +5,4 @@
 # - `make version-minor` for new features
 # - `make version-patch` for bug fixes
 
-__version__ = "99.8.0"  # 235b3e3580b24d464e4924064ac808f2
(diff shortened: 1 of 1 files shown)


quis (approved):

quis (changes requested):
Actually should be 100.1.0

Not sure how this has disappeared from the changelog:
https://github.com/alphagov/notifications-utils/commit/5686a252ba78bdabad22858173f48e5d8509d145

quis (approved):
Ok, that commit never made it into `main`. Proceed as before

rparke:
> Ok, that commit never made it into `main`. Proceed as before

https://github.com/alphagov/notifications-utils/commit/5686a252ba78bdabad22858173f48e5d8509d145 it's here on main though ... but now it's gone ?

PR #1250, opened 2025-07-07
Test version number matches latest changelog entry

quis:
We ask that all new versions have a corresponding changelog entry.

It’s also possible to add a changelog entry without updating the version which confuses things.

This commit adds a test to make sure they match.

Helpful error message in the case of a mismatch:

image


tests/test_version.py
```diff
@@ -0,0 +1,16 @@
+import re
+from os.path import relpath
+from pathlib import Path
+
+from notifications_utils import version
+
+
+def test_changelog_matches_latest_version():
+    changelog = Path("CHANGELOG.md")
+    versions = [line.strip(" #") for line in changelog.read_text().splitlines() if re.match(r"## \d+\.\d+\.\d+$", line)]
```
rparke, line 10: looks good, I'm assuming the regex is right to match <int>.<int>.<int>

rparke (approved):

PR #1251, opened 2025-07-10
Enforce strict undefined behaviour in Jinja templates

quis:
We use Jinja templates for things like rendering emails and letters to HTML.

These templates don’t use any third-party code like `govuk-frontend-jinja`.

By using [`StrictUndefined`](https://jinja.palletsprojects.com/en/stable/api/#jinja2.StrictUndefined) we have a better chance of catching typos in variable names etc

### DIFF
--- a/notifications_utils/template.py
+++ b/notifications_utils/template.py
@@ -6,7 +6,7 @@ from html import unescape
 from os import path
 from typing import Literal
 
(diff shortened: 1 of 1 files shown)


kr8n3r (approved):
see comment in admin pr https://github.com/alphagov/notifications-admin/pull/5539#pullrequestreview-3074684678

PR #1252, opened 2025-07-10
Add x-forwarded-for information to log files

corlettb:
https://trello.com/c/CjPJrFAk/1160-notifications-api-recording-all-remoteaddr-as-internal-ips

What
----

Add x-forwarded-for info to log files.

Why
----

We want to log the ips we see at each stage of the request.


risicle:
This could probably do with some tests before we merge it.

PR #1253, opened 2025-07-15
Update phonenumbers 9.0.9

rparke:
Major version bump from 8.13.52 to 9.0.9. Looking at the release notes, I'm confident that this is not technically a breaking change for python. The phonenumbers python port of libphonenumber repsects the semantic versioning of the original java codebase.

The [release notes for v9.0.0](https://github.com/google/libphonenumber/releases/tag/v9.0.0) tell us that the major version bump is due to an update to java 8 from 1.7. This is not a breaking change for the python port.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 100.1.0
+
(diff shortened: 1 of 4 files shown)


CrystalPea (approved):
LGTM

PR #1256, opened 2025-08-18
subclass _get_backend to return DisabledBackend object

BlessedDev:
🚨 This is a test branch to investigate disabling the celery result backend and will not be merged into main.
It overrides `celery._get_backend` so that it simply returns the `DisabledBackend` object once `result_backend` is set to None. 🚨

### DIFF
--- a/tests/test_celery.py
+++ b/tests/test_celery.py
@@ -4,6 +4,7 @@ import uuid
 
 import pytest
 from celery import Task
+from celery.backends.base import DisabledBackend
 from flask import g
 from freezegun import freeze_time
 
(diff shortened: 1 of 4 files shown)


tests/test_celery.py
```diff
@@ -256,3 +257,13 @@ def test_method_signatures(celery_app, async_task, method, _value):
     if method == "run":
         return
     assert inspect.signature(getattr(async_task.__class__, method)) == inspect.signature(getattr(Task, method))
+
+
+def test__get_backend_returns_DisabledBackground_object_when_result_backend_is_set_to_None(notify_celery):
+    notify_celery.conf.update({"result_backend": None})
```
quis, line 263: Should test that this works with both the value set to `None` and the config not set at all

tests/test_celery.py
```diff
@@ -256,3 +257,13 @@ def test_method_signatures(celery_app, async_task, method, _value):
...
+    assert isinstance(notify_celery._get_backend(), DisabledBackend)
```
quis, line 264: This would give you some confidence that your code is doing what you want (not trying to discover any backends):
```suggestion
    mock_by_url = mocker.patch("celery.app.backends.by_url")
    mock_by_name = mocker.patch("celery.app.backends.by_name")
    assert isinstance(notify_celery._get_backend(), DisabledBackend)
    assert mock_by_name.called is mock_by_url.called is False
```

Otherwise this test will pass with or without your changes.

PR #1262, opened 2025-10-03
Upgrade to python313

BlessedDev:
This PR upgrades notifications-utils to Python 3.13.

Python 3.13 brings changes to the behaviour of locals() which necessitated a compensatory change to get_remote_version()

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 101.3.0
+
+* Upgrade Python version to 3.13
+* Adjust get_remote_version() to compensate for changes in locals() in Python 3.13
+
 ## 101.2.1
 
 - Remove `clients.redis.rate_limit_cache_key` (no longer used)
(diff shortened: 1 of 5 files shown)


tests/version_tools/test_version_tools.py
```diff
@@ -0,0 +1,7 @@
+from notifications_utils.version_tools import get_file_contents_from_github, get_remote_version
+
+
+def test_get_remote_version():
```
quis, line 4: Not 100% sure we need a test for this but if we’re going to have one it shouldn’t go out to the internet.

Otherwise we:
- cannot run the tests offline 
- will get rate limited by Github’s API when running the tests multiple times

Either:
```suggestion
def test_get_remote_version():
    mock_requests_get = mocker.patch("notifications_utils.version_tools.requests.get", return_value=…)
```

Or:
```suggestion
def test_get_remote_version():
    mock_get_file_content_from_github = mocker.patch("get_file_contents_from_github", return_value='__version__ = "1.2.3"')
```

PR #1264, opened 2025-10-07
Don’t encode parentheses as HTML entities in emails

quis:
Looks like GMail and Outlook re-renders them literally as `&lpar;` rather than `(`.

This only happens when the user puts the entity in their content.

When we changed to support artibrary HTML entities in https://github.com/alphagov/notifications-utils/commit/e7886aa4e1daa8744262c8ee1469410543bff5c5 we kept parentheses as a special-case. This wasn’t necessary.

Probably still a good idea to special-case:
- ampersands (because URL query strings)
- non-breaking spaces (because maybe some clients might try to normalise unicode whitespace characters)

Before | After
---|---
image | image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 101.3.1
+
+* Fix rendering of parentheses in some HTML email clients
+
 ## 101.3.0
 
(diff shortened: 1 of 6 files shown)


tombye (approved):
This makes sense to me 👍🏻

PR #1336, opened 2026-03-30
Encrypting tokens with Fernet

whpearson:
We had decided to encrypt the tokens so that emails in them could not
be read by logit.

The rollout plan is: pass the encryption key in to everywhere to try to decrypt
things in check_token first, then the encryption key to generate_token.

This code uses Fernet to encrypt things. Although it is going to
be more verbose than we want I think.

Fernet does date verification and but we need to try to decrypt twice as the errors are not
as specific as we might want.


notifications_utils/url_safe_token.py
```diff
@@ -1,13 +1,32 @@
...
+from itsdangerous import SignatureExpired, URLSafeTimedSerializer
 
 from notifications_utils.formatters import url_encode_full_stops
 
 
-def generate_token(payload, secret, salt):
+def generate_token(payload, secret, salt, encrypt_secret=None):
+    if encrypt_secret:
+        f = Fernet(encrypt_secret)
+        return f.encrypt(payload.encode(encoding="utf-8"))
```
quis:
```suggestion
        return f.encrypt(payload.encode()
```

> encoding defaults to `'utf-8'`

– https://docs.python.org/3/library/stdtypes.html#str.encode

notifications_utils/url_safe_token.py
```diff
@@ -1,13 +1,32 @@
-from itsdangerous import URLSafeTimedSerializer
+from cryptography.fernet import Fernet, InvalidToken
+from itsdangerous import SignatureExpired, URLSafeTimedSerializer
 
 from notifications_utils.formatters import url_encode_full_stops
 
 
-def generate_token(payload, secret, salt):
+def generate_token(payload, secret, salt, encrypt_secret=None):
```
quis: `encrypt_secret` sounds like a boolean flag, `encryption_secret` would be a better variable name

notifications_utils/url_safe_token.py
```diff
@@ -1,13 +1,32 @@
...
+
     return url_encode_full_stops(URLSafeTimedSerializer(secret).dumps(payload, salt))
 
 
-def check_token(token, secret, salt, max_age_seconds):
+def check_token(token, secret, salt, max_age_seconds, encrypt_secret=None):
+    if encrypt_secret:
+        f = Fernet(encrypt_secret)
+        try:
+            payload = f.decrypt(token).decode(encoding="utf-8")
```
quis:
```suggestion
            payload = f.decrypt(token).decode()
```

> encoding defaults to `'utf-8'`

– https://docs.python.org/3/library/stdtypes.html#bytes.decode

notifications_utils/url_safe_token.py
```diff
@@ -1,13 +1,32 @@
...
+def generate_token(payload, secret, salt, encrypt_secret=None):
+    if encrypt_secret:
+        f = Fernet(encrypt_secret.encode(encoding="utf-8"))
+        return f.encrypt(payload.encode(encoding="utf-8"))
+
     return url_encode_full_stops(URLSafeTimedSerializer(secret).dumps(payload, salt))
 
 
-def check_token(token, secret, salt, max_age_seconds):
+def check_token(token, secret, salt, max_age_seconds, encrypt_secret=None):
```
quis: I’d consider using multiple `except`s to put the happy path first. Pseudo code:

```python
def check_token(…):
    try:    
        return decode(…, max_age_seconds)
    except TypeError:
        pass
    except InvalidToken:
        try:
            decode(…)
            raise SignatureExpired
        except InvalidToken:
            pass
```

quis (commented):
Looks good now. Think it still needs tests which call the function with an `encryption_key` and:
- an expired token
- a token of the wrong type (for example, `1`, `None`, etc)

quis (approved):
Code and tests look good now.

You’ll need to run `make freeze-requirements` so that `cryptography` gets put into `requirements_for_test.txt` and therefore installed on CI.

quis (approved):

requirements_for_test.txt
```diff
@@ -1,5 +1,5 @@
 # This file was autogenerated by uv via the following command:
-#    uv pip compile requirements_for_test.in pyproject.toml --output-file requirements_for_test.txt --exclude-newer 14d
+#    uv pip compile requirements_for_test.in pyproject.toml --output-file requirements_for_test.txt
```
spatel033, line 2: Has this (`--exclude-newer 14d`) been intentionally removed?

quis: This just records how the file was generated. `--exclude-newer` is only applied when running `make refreeze-requirements`. In this PR Will has run `make freeze-requirements` which doesn’t use that flag. TL:DR this is fine.

PR #1345, opened 2026-04-28
 `SerialisedModelCollection`: implement `__iter__` using `get_model_instance_from_item`

risicle:
If we want to be able to use e.g. admin's [`InterruptibleIterableMixin`](https://github.com/alphagov/notifications-admin/blob/d565035bc5a908efc5b06eeb2d3af5b3718ad4bd/app/utils/interruptible_io.py#L161) with a `SerialisedModelCollection`, it needs to be an actual `Iterable` (i.e. implement `__iter__`). Up till now iteration through a `SerialisedModelCollection` has only been working through python's sequence-based iteration which is less flexible.

This is a potentially breaking change, details of which are included in the changelog entry here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 114.0.0
+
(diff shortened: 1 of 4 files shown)


quis (commented):
Worth a quick `assert next(item).x == "foo"` under these lines to prove it’s iterable?

https://github.com/alphagov/notifications-utils/blob/3079e840de0f3d1e9c10801804cdd46e86d5bcaa/tests/test_serialised_model.py#L217-L221

quis (approved):

PR #1353, opened 2026-05-14
`RedisClient`: use `HardEventletTimeout` as default `always_raise` exception, allow configuration of `socket_timeout` and `socket_connect_timeout`

risicle:
https://trello.com/c/PEpdGAPE/1518-implement-solution-for-failed-writes-deletes-to-redis

See individual commits for remarks, these are vaguely-related `RedisClient` changes that should both make our use of redis a bit more reliable during e.g. a failover.

Tested on multiple dev envs, this does indeed have some effect on our robustness during a failover.

### DIFF
--- a/notifications_utils/clients/redis/redis_client.py
+++ b/notifications_utils/clients/redis/redis_client.py
@@ -56,8 +56,10 @@ class RedisClient:
 
     def init_app(self, app):
         self.active = app.config.get("REDIS_ENABLED")
+        socket_timeout = app.config.get("REDIS_SOCKET_TIMEOUT")
+        socket_connect_timeout = app.config.get("REDIS_SOCKET_CONNECT_TIMEOUT")
         if self.active:
-            self.redis_store.init_app(app)
+            self.redis_store.init_app(app, socket_timeout=socket_timeout, socket_connect_timeout=socket_connect_timeout)
 
             self.register_scripts()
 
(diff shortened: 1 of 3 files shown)


notifications_utils/clients/redis/redis_client.py
```diff
@@ -52,12 +52,14 @@ class RedisClient:
     redis_store = FlaskRedis()
     active = False
     scripts = {}
-    always_raise: tuple[type[BaseException], ...] = (EventletTimeout,)
+    always_raise: tuple[type[BaseException], ...] = (HardEventletTimeout,)
 
     def init_app(self, app):
         self.active = app.config.get("REDIS_ENABLED")
+        socket_timeout = app.config.get("REDIS_SOCKET_TIMEOUT")
+        socket_connect_timeout = app.config.get("REDIS_SOCKET_CONNECT_TIMEOUT")
```
quis, line 60: I’ve checked these match https://redis.io/docs/latest/develop/clients/redis-py/produsage/#timeouts

notifications_utils/clients/redis/redis_client.py
```diff
@@ -52,12 +52,14 @@ class RedisClient:
...
         if self.active:
-            self.redis_store.init_app(app)
+            self.redis_store.init_app(app, socket_timeout=socket_timeout, socket_connect_timeout=socket_connect_timeout)
```
quis, line 62: I’ve checked these get passed to py-redis when used as keyword arguments to `init_app`:
https://github.com/underyx/flask-redis/blob/3ad7661ad3f59f7580d1fe02de64cd5e876f6854/flask_redis/client.py#L37-L39

quis (approved):

PR #1357, opened 2026-05-20
Add `file_types.mime_type_from_extension`

quis:
Python’s built in `mimetypes` module doesn’t have mappings for `docx`<sup>*</sup> by default. It can enable some extra extensions by calling `mimetypes.init()` but this is platform dependent.

There’s only a limited and known set of mime types we need to support. So that we consistently serve the right mime type for `.docx` files – both in the admin app and in document download API – let’s make our own mapping.

***

<sup>*</sup>It will from 3.14 onwards: https://github.com/python/cpython/commit/1cf9b6d9b8fbb5ebc3e9b42a3682684a983c78bc

### DIFF
--- a/tests/test_file_types.py
+++ b/tests/test_file_types.py
@@ -5,6 +5,7 @@ from notifications_utils.file_types import (
     format_file_type,
     is_allowed_file_extension,
     is_allowed_mime_type,
+    mime_type_from_extension,
 )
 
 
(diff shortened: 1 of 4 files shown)


tests/test_file_types.py
```diff
@@ -78,6 +79,29 @@ def test_extension_from_mime_type(mime_type, extension):
     assert extension_from_mime_type(mime_type) == extension
 
 
+@pytest.mark.parametrize(
+    "mime_type, extension",
+    (
+        ("pdf", "application/pdf"),
```
BlessedDev, line 85: Should this not be the other way round ie mimetype : "application/pdf" and extension : "pdf" ?

BlessedDev: My apologies, my mistake. It is doing what it is intended to do.

quis: Nah you’re right, it only worked because the argument names were the wrong way round everywhere. I’ve fixed it now

BlessedDev (approved):

PR #1359, opened 2026-05-21
Remove StatsD client

robinjam:
https://trello.com/c/25A3gpXh/1358-replace-statsd-instrumentation-across-notify-with-otel

Removes the StatsD client.

- All of the metrics that _utils_ currently collects via StatsD are also collected via OpenTelemetry.
- The only _app_ that currently collects metrics via StatsD is api, and all of the StatsD-based metrics it collects now have OpenTelemetry equivalents.
- The Grafana dashboards have already been updated to use the new OpenTelemetry-based metrics (with the exception of "Notify summary", which I'm replacing with a set of new dashboards).

I have already prepared PRs bumping utils to this version for anivirus, api and template-preview. These will obviously need rebasing after this goes in.


notifications_utils/clients/statsd/__init__.py
quis, line 1: Can we remove this file and its parent directory entirely?

robinjam: I did remove `__init__.py`, I don't know why the diff isn't showing that on GitHub. If you check out this branch locally you'll see the dir is empty (git doesn't have any concept of directories so it won't remove the dir itself).

robinjam:
Rebased on main

quis (approved):

PR #1361, opened 2026-05-26
`RedisClient`: add `redis_flakey` mechanism to avoid repeated timeouts

risicle:
https://trello.com/c/PEpdGAPE/1518-implement-solution-for-failed-writes-deletes-to-redis

When connections to redis are timing out, though we can reduce the timeout wait, many requests will attempt to access redis on multiple occasions. in these cases it will end up waiting the timeout period each time. These can add up to make a request fail (e.g. run into an `EventletTimeout`) where it could otherwise have passed if it had got the hint that redis was in bad shape the first time it got a timeout.

This adds an app-context-local flag `redis_flakey` that will prevent repeated attempts at accessing redis for calls that
are deemed `skippable`. `skippable` means that the consequences of omitting the redis request are not catastrophic and won't cause a consistency problem (just slightly poorer performance).

Call sites can themselves set the `skippable` argument, but by default `get`s are considered skippable and `set`s aren't.

`RequestCache.set` overrides this to set `skippable=True` for its `RedisClient.set` calls because the inner `client_method` call *should* be side-effect-free - i.e. nothing has just happened that requires us to invalidate an existing cache entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 120.1.0
+
+* `RedisClient`: add `redis_flakey` mechanism to avoid repeated timeouts
+
 ## 120.0.0
 
 * Removes `BaseLetterTemplate.too_many_pages` (subclasses which define `page_count` should implement this themselves)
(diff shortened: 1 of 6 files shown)


robinjam (approved):
I think I'm happy with this approach. Really sorry it took so long to get reviewed...

Happy to re-approve once you've fixed the merge conflicts

risicle:
It's no problem - it had to have proper consideration.

quis (approved):

PR #1362, opened 2026-05-27
Improve counting of fragments and characters in text message templates 

quis:
# Add some useful extra properties to `BaseSMSTemplate`

## `count_of_characters_above_limit`

This is a useful property which can replace the mid-string-interpolation maths done here in the admin app:
https://github.com/alphagov/notifications-admin/blob/0b04123230f9586dc9e2ceecfac6b2df7812d39a/app/main/views/templates.py#L855

## `non_gsm_characters`

This is a useful property which returns the exact characters causing the message to not be sent as GSM-7, and therefore costing more.

## `count_of_characters_above_previous_fragment_boundary`
    
This is a useful property which we can use to tell users how many characters they need to remove to guarantee their message will be cheaper to send.

# Refactor helper functions into properties of `BaseSMSTemplate`

This makes it harder to introduce bugs where the wrong content or flags are given to these functions, because they are always looking at properties of the current instance, rather than arguments which could be any arbitrary value.

This means removing the following functions from `notifications_utils.template`:
- `get_sms_fragment_count`
- `non_gsm_characters`
- `count_extended_gsm_chars`

None of these functions were not imported by any of our apps. The corresponding properties on `BaseSMSTemplate` existed already.

# Ignore names of placeholders when determining non-GSM characters

Names of placeholders won’t end up in the final message – they will be replaced.

We shouldn’t include them when checking for:
- non-GSM characters which cause the message to change encoding
- extended GSM characters

Otherwise we could wrongly tell the user that their message will be charged as more fragments that is true.

## Before 

image

## After 

image

### DIFF
--- a/tests/test_template_types.py
+++ b/tests/test_template_types.py
@@ -1048,7 +1048,7 @@ def test_character_count_for_sms_templates(
 
 
 @pytest.mark.parametrize(
-    "msg, expected_sms_fragment_count, expected_count_of_characters_above_previous_fragment_boundary",
+    "template_content, expected_sms_fragment_count, expected_count_of_characters_above_previous_fragment_boundary",
     [
         ("à" * 71, 1, 71),  # welsh character in GSM
         ("à" * 160, 1, 160),
@@ -1083,11 +1083,11 @@ def test_character_count_for_sms_templates(
 )
 def test_sms_fragment_count_accounts_for_unicode_and_welsh_characters(
     template_class,
-    msg,
+    template_content,
     expected_sms_fragment_count,
     expected_count_of_characters_above_previous_fragment_boundary,
 ):
-    template = template_class({"content": msg, "template_type": "sms"})
+    template = template_class({"content": template_content, "template_type": "sms"})
     assert template.fragment_count == expected_sms_fragment_count
     assert template.count_of_characters_above_previous_fragment_boundary == (
         expected_count_of_characters_above_previous_fragment_boundary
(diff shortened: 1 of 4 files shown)


tests/test_template_types.py
```diff
@@ -1048,60 +1048,133 @@ def test_character_count_for_sms_templates(
...
+    template = template_class({"content": msg, "template_type": "sms"})
     assert template.fragment_count == expected_sms_fragment_count
+    assert template.count_of_characters_above_previous_fragment_boundary == (
+        expected_count_of_characters_above_previous_fragment_boundary
+    )
 
 
 @pytest.mark.parametrize(
-    "msg, expected_sms_fragment_count",
+    "msg, expected_sms_fragment_count, expected_count_of_characters_above_previous_fragment_boundary",
```
CrystalPea: [suggestion] we could rename msg to test_characters - that would provide more info and be better for potential dyslexic colleagues (gsm and msg are anagrams)

tests/test_template_types.py
```diff
@@ -1732,13 +1805,21 @@ def test_lists_in_combination_with_other_elements_in_letters(markdown, expected)
...
+        (1_000, 84),
+    ),
+)
+def test_message_too_long_ignoring_prefix(template_class, repeat_character_count, expected_count_above_limit):
+    body = ("b" * repeat_character_count) + "((foo))"
     template = template_class(
         {"content": body, "template_type": template_class.template_type}, prefix="a" * 100, values={"foo": "cc"}
     )
-    # content length is prefix + 919 characters (more than limit of 918)
+    # content length is repeated characters plus personalisation (more than limit of 918)
```
CrystalPea: what do repeated characters here mean? 

Ahh it means in the test we are repeating the character ok. Maybe we could say it's template content + personalisation? As those repeated characters stand in for template content, right?

CrystalPea (commented):
Nice prep work! Just a couple small naming comments 🙌🏼 . Other than that it looks great!

quis:
@CrystalPea thanks for the comments, I have renamed the variables in 7dd849d29228f1940ae96756c578b6c589d7afff

CrystalPea (approved):
Looks good :shipit:

PR #1365, opened 2026-06-02
Add `max_items_shown` argument to `formatted_list`

quis:
This is useful when you want to display a list of things in prose, but don’t want that list getting too long.

It provides a way of truncating the output by specifying the maximum number of items to be shown and a word to use when there are more than this number of items.

***

Before:

> Avocado, banana, dragon fruit, mango, papaya and pineapple

After:

> Avocado, banana and more

***

Can replace what we are doing in Jinja here:
https://github.com/alphagov/notifications-admin/blob/1791ba2eb8f0dd45352ce0ded91d6e27d2fa1bfe/app/templates/partials/templates/content-count-message.html#L10-L14

### DIFF
--- a/notifications_utils/formatters.py
+++ b/notifications_utils/formatters.py
@@ -187,7 +187,14 @@ def url_encode_full_stops(value):
 
 
 def unescaped_formatted_list(
-    items, conjunction="and", before_each="‘", after_each="’", separator=", ", prefix="", prefix_plural=""
+    items,
+    *,
+    conjunction="and",
+    before_each="‘",
+    after_each="’",
+    separator=", ",
+    prefix="",
+    prefix_plural="",
 ):
     if prefix:
         prefix += " "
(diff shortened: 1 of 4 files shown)


notifications_utils/formatters.py
```diff
@@ -187,31 +188,44 @@ def url_encode_full_stops(value):
...
+    if len(items) > max_items_shown:
+        cutoff = max(max_items_shown - 1, 1)
+        formatted_items = [f"{before_each}{item}{after_each}" for item in items[:cutoff]] + [word_for_items_not_shown]
+    else:
         formatted_items = [f"{before_each}{item}{after_each}" for item in items]
 
-        first_items = separator.join(formatted_items[:-1])
-        last_item = formatted_items[-1]
-        return f"{prefix_plural}{first_items} {conjunction} {last_item}"
+    if len(items) == 1:
```
CrystalPea, line 219: is there a reason we moved this one lower down? It's nice to have the simplest case on top, no?

CrystalPea (commented):
Nice one! One small comment 🙌🏼

notifications_utils/formatters.py
quis: Because now we add the `before_each` and `after_each` to all the items first, before deciding what to return. In a way this case is still on top, in that it’s the first `return` in the function.

BlessedDev (approved):

PR #1369, opened 2026-06-10
`InsensitiveSet`: reimplement without `OrderedSet`

risicle:
This should allow a significantly faster implementation, using modern python's order-preserving dict, storing normalized items as keys and the original items as their corresponding values.

Behaviour is a compromise between what I deem Sensible, yet following the original `InsensitiveSet`'s behaviour closely enough to pass the unmodified test suite. For example, this implements equality comparison against plain `Iterable`s that requires ordering to match, even though I think it's a bit silly.

Operations and reverse-operations should work, even against a plain `Iterable`.

The existing implementation of `InsensitiveSet` is a little bit odd/incomplete and has quite an inefficient implementation of `__contains__` (which may be the most heavily used of its set operations). A more robust implementation of `InsensitiveSet` will allow us to confidently make more use of it within `Template` & `RecipientCSV` for performance improvements there.

~Note the expected results in the expanded tests are flexible in the accepted intersecting un-normalised values for `__or__`, `__ror__` and `__ior__` (i.e. do they come from the left side or right side?). This is because I'd really prefer to be free to choose whatever algorithm is fastest for these operations and not have users depend heavily on the exact behaviour.~ Instead of this, I'm strictly declaring that un-normalised values will come from the **right-hand-side** of an operation and the **last** occurrence in a value stream containing duplicates. This means I had to make an alteration to `Field.placeholder` to make it prioritise the _first_ occurrence of duplicated placeholders.

A very rudimentary test based on an expanded version of `test_markdown_in_templates` called in a loop showed a modest ~9% speedup in template processing. Given `InsensitiveSet` isn't currently _very_ heavily used in template processing, I think this is a decent result.

The new `InsensitiveSet` is also heavily type-hinted.

### DIFF
--- a/notifications_utils/insensitive_dict.py
+++ b/notifications_utils/insensitive_dict.py
@@ -1,4 +1,9 @@
+from abc import ABCMeta, abstractmethod
+from collections.abc import Iterable, Iterator, MutableSet, Sequence, Set
 from functools import lru_cache
+from itertools import chain, islice
+from types import NotImplementedType
+from typing import Self, TypeVar, overload
 
 from ordered_set import OrderedSet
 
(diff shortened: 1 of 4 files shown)


tests/test_insensitive_dict.py
```diff
@@ -209,3 +210,247 @@ def test_insensitive_set_symetric_difference():
...
+        " 2",
+        " 3",
+        " 4",
+    )
+
+
+def test_insensitive_set_eq_insensitive_set_not_order_sensitive():
+    assert InsensitiveSet(f" {i}" for i in range(8)) == InsensitiveSet(f"{i} " for i in range(7, -1, -1))
+    assert InsensitiveSet(f" {i}" for i in range(8)) == InsensitiveSet(f"{i} " for i in range(8))
+    assert InsensitiveSet(f" {i}" for i in range(8)) != InsensitiveSet(f"{i} " for i in range(7))
```
quis, line 453: Should probably have another test here for comparison against regular set, eg 
```
set(f" {i}" for i in range(8)) == InsensitiveSet(f"{i} " for i in range(7, -1, -1))
```

notifications_utils/insensitive_dict.py
```diff
@@ -54,55 +56,333 @@ def as_dict_with_keys(self, keys):
...
+
+    # Sequence[T]
+
+    @overload
+    def __getitem__(self, index: int) -> T: ...
+
+    @overload
+    def __getitem__(self, index: slice) -> Self: ...
+
+    def __getitem__(self, index):  # noqa: C901 is bunk
```
quis, line 127: What the advantage of this over `return type(self)._from_inner_pairs(list(self._inner.items())[index])`? Not building the whole list in memory?

notifications_utils/insensitive_dict.py
```diff
@@ -54,55 +56,333 @@ def as_dict_with_keys(self, keys):
...
+            if candidate == key:
+                return i
+
+        raise KeyError
+
+    # Accelerate Set[T]
+
+    def __le__(self, other: Set) -> bool:
+        if not isinstance(other, Set):
+            return NotImplemented
```
quis, line 201: Could probably do with a test for each of these lines, something like 
```python
with pytest.raises(…):
    InsensitiveSet() <= 1
```

Also TIL that `NotImplemented` and `NotImplementedError` are different things for different purposes.

notifications_utils/insensitive_dict.py
```diff
@@ -54,55 +56,333 @@ def as_dict_with_keys(self, keys):
...
+    # Accelerate MutableSet[T]
+
+    def remove(self, item: T):
+        del self._inner[self.make_key(item)]
+
+    def pop(self):
+        try:
+            del self._inner[next(reversed(self._inner))]
+        except StopIteration as e:
+            raise KeyError from e
```
quis, line 321: `pop` should:
- accept an optional `position`, the index of the item to pop
- return the value which has been removed

risicle:
🤔 I'm perhaps going to reconsider the "not wanting to guarantee the un-normalized results of a union" thing because the admin tests won't pass without me ensuring the left-hand-side wins.

notifications_utils/insensitive_dict.py
risicle: Agree with the latter, where are you getting the former from?

The mixin implementation for `MutableSet` doesn't know about `position`: https://github.com/python/cpython/blob/a1873300eebe9c634f59592c3333035768f09de9/Lib/_collections_abc.py#L720

quis: `list` (which is ordered and mutable) lets you pop by `pos`. `OrderedSet` implements this too:
https://github.com/rspeer/ordered-set/blob/9dfc91c27c0c56df2536036bec4a5b3f717ef2ba/ordered_set/__init__.py#L251

I’m not precious about including it though, given:
- this is a method we’re unlikely to use directly 
- it’s not something covered by the existing test suite

risicle: Ah yes, `MutableSequence` has `.pop(index)` ... but we're not implementing `MutableSequence` :D

risicle:
Thinking further, it feels like the "right" thing to do is **actually** to always let the **later** value "win", because this is how `.update()` always works for duplicates _within_ an `.update()` item-pair stream. The only way to make it consistently let the **first** value win would be to stick a generator into each `.update()` item-pair stream to do a weird reach-around check..

quis:
The reason we care (and the reason why that admin test fails) is because we want the form label to match the first instance of the placeholder in the template exactly:

image

risicle:
We could always feed the placeholders into the `InsensitiveSet` in reverse.

quis:
I looked into it further and the problem isn’t just with unions:
```python
InsensitiveSet(('NAME', 'name'))
>>> InsensitiveSet(['name'])
```

I guess `Field.placeholders` could do a double-reverse but I don’t love it…

notifications_utils/insensitive_dict.py
```diff
@@ -54,55 +56,333 @@ def as_dict_with_keys(self, keys):
...
-        InsensitiveSet(['FIRST-name'])[0]
-        >>> 'FIRST-name'
-    """
+class AbstractInsensitiveSet(MutableSet[T], Sequence[T], metaclass=ABCMeta):
+    __slots__ = ("_inner",)
+    _inner: dict[T, T]
 
-    def __init__(self, iterable=None, /):
-        return super().__init__(InsensitiveDict.from_keys(iterable or ()).values())
+    @abstractstaticmethod
```
quis: [Deprecated since Python 3.3](https://docs.python.org/3/library/abc.html#abc.abstractstaticmethod), prefer:
```suggestion
    @staticmethod
    @abstractmethod
```

risicle:
Yep, the workaround I've put in is a double-reverse, which I don't love either, but remember previously we were constructing a new `InsensitiveDict` for each `__contains__` call. This is only _one_ additional dict construction.

notifications_utils/insensitive_dict.py
risicle: > Not building the whole list in memory?

This.

risicle:
We need to agree on exact desired behaviour.

risicle:
Chris sez he would ideally like it to work with first and left-most un-normalised values take precedence over later and right-most ones. This makes sense in some ways but it's awkward to make a fast implementation.

risicle:
Have pushed a version that is left-prioritising.

quis (approved):

PR #1381, opened 2026-06-18
Don’t declare `start` on NotifyTask

quis:
It’s better if code that tries to access `start` before it’s been set to an actual time gets an `AttributeError`, rather than unexpectedly handling a `None` in a way which may or may not raise


quis:
Closing in favour of adding a type hint when we come to tackling this module

PR #1382, opened 2026-06-18
Add type checking

quis:
We can’t run this as part of CI yet because it flags many errors. But choosing a tool and command to run gives us a consistent set of errors between environments.

Going to try out [ty](https://github.com/astral-sh/ty/) because it’s way, way faster than mypy.


quis:
Closing in favour of https://github.com/alphagov/notifications-utils/pull/1384

PR #1383, opened 2026-06-18
Fix submodule imports

quis:
Fixes these errors/warnings:

```
error[possibly-missing-submodule]: Submodule `parse` might not have been imported
  --> notifications_utils/s3.py:28:16
   |
28 |         tags = urllib.parse.urlencode(tags)
   |                ^^^^^^^^^^^^
   |
help: Consider explicitly importing `urllib.parse`
```

```
warning[possibly-missing-submodule]: Submodule `exceptions` might not have been imported
  --> notifications_utils/s3.py:94:12
   |
94 |     except botocore.exceptions.ClientError as e:
   |            ^^^^^^^^^^^^^^^^^^^
   |
help: Consider explicitly importing `botocore.exceptions`
```

***

None of the apps import `urllib` or `botocore` from these modules (for example `from notifications_utils.s3 import botocore`) so this isn’t a breaking change

### DIFF
--- a/notifications_utils/formatters.py
+++ b/notifications_utils/formatters.py
@@ -1,8 +1,8 @@
 import math
 import re
 import string
-import urllib
 from html import _replace_charref, escape
+from urllib.parse import quote
 
 import smartypants
 from markupsafe import Markup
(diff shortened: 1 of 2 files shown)


joybytes (approved):
⭐

PR #1384, opened 2026-06-23
Install `mypy`

quis:
Installs mypy for type checking, but does not make it part of CI yet (there are many errors to fix first).

Choosing mypy for consistency with the API codebase.

Also installs the recommended stubs for things we use that aren’t part of the standard library.


risicle (approved):
WFM

PR #1389, opened 2026-06-23
Add type hint for `SanitiseText.ALLOWED_CHARACTERS`

quis:
Fixes this error:
> Need type annotation for "ALLOWED_CHARACTERS" (hint: "ALLOWED_CHARACTERS: set[<type>] = ...")

### DIFF
--- a/notifications_utils/sanitise_text.py
+++ b/notifications_utils/sanitise_text.py
@@ -1,8 +1,9 @@
 import unicodedata
+from collections.abc import Set
 
 
 class SanitiseText:
-    ALLOWED_CHARACTERS: set[str] = set()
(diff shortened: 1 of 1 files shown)


risicle (approved):

risicle:
Though I wonder if we should be preferring the generic `Set` from `collections.abc` instead of the specific implementation `set`. Would give us more flexibility and make the user assume less.

quis:
@risicle good point, changed in [178c185](https://github.com/alphagov/notifications-utils/pull/1389/commits/178c185a72ca62429e9488aaedb93cba83ac0905)

risicle (approved):

PR #1396, opened 2026-06-24
Ignore error about import from `html` library

quis:
mypy says:
> error: Module `html` has no attribute `_replace_charref`  [attr-defined]

Typeshed contains external type annotations for the Python standard library, and is used by mypy.

I’ve fixed this problem upstream in https://github.com/python/typeshed/pull/15925

This fix won’t be in mypy until the next feature release, so we need to ignore the error until then.

### DIFF
--- a/notifications_utils/formatters.py
+++ b/notifications_utils/formatters.py
@@ -1,7 +1,10 @@
 import math
 import re
 import string
-from html import _replace_charref, escape
+from html import (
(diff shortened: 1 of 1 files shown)


risicle (approved):

risicle:
Eek. My `mypy` doesn't like this:

> notifications_utils/formatters.py:5: error: Invalid "type: ignore" comment  [syntax]

PR #1397, opened 2026-06-24
Remove references to gunicorn from HTTP headers in a type-checker-compliant way

quis:
Gunicorn uses the `SERVER` constant here:
https://github.com/benoitc/gunicorn/blob/a8283bbf5e1416d0dd13f994f71ec2761988aeab/gunicorn/http/wsgi.py#L293

To populate the `Server:` HTTP header here:
https://github.com/benoitc/gunicorn/blob/a8283bbf5e1416d0dd13f994f71ec2761988aeab/gunicorn/http/wsgi.py#L426

If we don’t want to reveal to any automated scanners what software our servers are running, we should change this.

image

***

We can keep setting `SERVER_SOFTWARE` to `"None"` as well, since Gunicorn still defines it and might start using it again some day.

***

Both should be ignored by the type checker since the type hints for Gunicorn mark them as final:
https://github.com/python/typeshed/blob/f9e0f8ce9a933648022d618961693ee74cb8d4f4/stubs/gunicorn/gunicorn/__init__.pyi#L6

But we have a good reason for overriding them.

### DIFF
--- a/notifications_utils/gunicorn/defaults.py
+++ b/notifications_utils/gunicorn/defaults.py
@@ -73,4 +73,4 @@ def set_gunicorn_defaults(globals_dict: dict):
         worker_abort=worker_abort,
         worker_int=worker_int,
     )
-    gunicorn.SERVER_SOFTWARE = "None"
+    gunicorn.SERVER_SOFTWARE = "None"  # type: ignore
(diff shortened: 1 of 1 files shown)


quis:
Closing in favour of: https://github.com/alphagov/notifications-utils/pull/1398

PR #1399, opened 2026-06-24
add type hints to `Template`, `Field` and some related classes

risicle:
This doesn't produce any new `mypy` errors for me (and fixes one).

### DIFF
--- a/notifications_utils/insensitive_dict.py
+++ b/notifications_utils/insensitive_dict.py
@@ -3,10 +3,10 @@ from collections.abc import Iterable, Iterator, MutableSet, Sequence, Set
 from functools import lru_cache
 from itertools import chain, islice
 from types import NotImplementedType
-from typing import Self, TypeVar, overload
+from typing import Self, overload
 
 
-class InsensitiveDict(dict):
+class InsensitiveDict[K, V](dict[K, V]):
     """
     `InsensitiveDict` behaves like an ordered dictionary, except it normalises
     case, whitespace, hypens and underscores in keys.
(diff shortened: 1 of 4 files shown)


notifications_utils/formatters.py
```diff
@@ -1,7 +1,7 @@
 import math
 import re
 import string
-from html import _replace_charref, escape
+from html import _replace_charref, escape  # type: ignore[attr-defined]
```
quis: Prefer https://github.com/alphagov/notifications-utils/pull/1396

risicle: Agreed, approved.

notifications_utils/template.py
```diff
@@ -70,11 +71,13 @@ class Template(ABC):
...
         pass
 
+    _values: Mapping[str, Any]
+
     def __init__(
         self,
-        template,
-        values=None,
-        redact_missing_personalisation=False,
+        template: dict[str, str],
```
quis:
```suggestion
        template: dict[str, Any],
```

The fields in `template` ultimately come from what is serialised by this schema:
https://github.com/alphagov/notifications-api/blob/6ab134a7689da6d2e3e52cba78802eeaab571e69/app/schemas.py#L385-L437

This can include `None` and `bool`s, and could include any JSON-serialisable type in the future

risicle: Hmmmm.....

notifications_utils/insensitive_dict.py
```diff
@@ -5,8 +5,11 @@
 from types import NotImplementedType
 from typing import Self, TypeVar, overload
 
+K = TypeVar("K")
+V = TypeVar("V")
 
-class InsensitiveDict(dict):
+
+class InsensitiveDict(dict[K, V]):
```
quis: Python 3.12 and above have [syntactic support for generics](https://docs.python.org/3/reference/compound_stmts.html#type-params) so you can do:
```suggestion
class InsensitiveDict[K, V](dict[K, V]):
```

quis (approved):

PR #1402, opened 2026-06-24
Use `InsensitiveSet` to merge placeholders with personalisation keys

quis:
We want a case-and-white-space-unique sequence of field names, giving preference to the format of keys as they appear in the template’s placeholders.

`InsensitiveSet` now gives us an elegant way of doing this.

This behaviour is already tested here:
https://github.com/alphagov/notifications-utils/blob/500a39050f718007f8888c6d8e95216dc67bf4aa/tests/test_base_template.py#L71-L82

### DIFF
--- a/notifications_utils/template.py
+++ b/notifications_utils/template.py
@@ -122,15 +122,14 @@ class Template(ABC):
 
     @property
     def values(self) -> Mapping[str, Any]:
-        if hasattr(self, "_values"):
-            return self._values
-        return {}
+        return getattr(self, "_values", {})
 
     @values.setter
     def values(self, new_values: Mapping[str, Any] | None):
-        if not new_values:
-            self._values = {}
-        else:
+        with suppress(AttributeError):
+            del self._values
+
+        if new_values:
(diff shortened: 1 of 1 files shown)


notifications_utils/template.py
```diff
@@ -121,20 +121,15 @@ def content_with_placeholders_filled_in(self) -> str:
...
-            placeholders = InsensitiveDict.from_keys(self.placeholders)
-            self._values = InsensitiveDict(new_values).as_dict_with_keys(
-                self.placeholders
-                | {key for key in new_values.keys() if InsensitiveDict.make_key(key) not in placeholders.keys()}
-            )
+        with suppress(AttributeError):
+            del self._values
+
+        if new_values:
+            self._values = InsensitiveDict(new_values).as_dict_with_keys(self.placeholders | set(new_values.keys()))
```
risicle: You shouldn't have to wrap `new_values.keys()` in `set` - `InsensitiveSet` implements `__or__` against plain iterables.

quis: Fixed and rebased

risicle (approved):

PR #1405, opened 2026-06-24
Add type hints for postal addresses

quis:
To test this locally:
```shell
mypy notifications_utils/recipient_validation/postal_address.py
```


notifications_utils/countries/__init__.py
```diff
@@ -18,7 +18,7 @@
 class CountryMapping(InsensitiveDict):
     @staticmethod
     @lru_cache(maxsize=2048, typed=False)
-    def make_key(original_key):
+    def make_key(original_key: str) -> str:
```
risicle: In my new `InsensitiveDict`, I've made it so `make_key` also accepts `Any` and raises a `TypeError` if the value can't be handled, which the `__contains__` implementation catches and turns into `False`

notifications_utils/countries/__init__.py
```diff
@@ -29,16 +29,16 @@ def make_key(original_key):
 
         return SanitiseASCII.encode(normalised)
 
-    def __contains__(self, key):
-        if any(c.isdigit() for c in key):
+    def __contains__(self, key: object) -> bool:
+        if isinstance(key, str) and any(c.isdigit() for c in key):
```
risicle, line 51: Haven't seen this code before, but I'd be more happy with this optimisation if `make_key` didn't happily accept a digit.

notifications_utils/countries/__init__.py
quis: I’ve made this work in the same way now

notifications_utils/countries/__init__.py
```diff
@@ -15,10 +17,12 @@
 )
 
 
-class CountryMapping(InsensitiveDict):
+class CountryMapping[K: str, V: str](InsensitiveDict):
```
risicle:
```suggestion
class CountryMapping[K: str, V: str](AbstractInsensitiveDict[K, V]):
```

But one thing that remains incorrect about this is not having a `.keys()` implementation, which requires an `AbstractInsensitiveSet` as a return type, which suggests splitting the `make_key` function out of `CountryMapping` so it can be shared between the two.

The current behaviour would return an `InsensitiveSet` from `.keys()`, which would follow only its normalisation rules.

notifications_utils/countries/__init__.py
quis: Changed in 16b09e6332d9bbf7e57138ae2014d3ebe38fa270

notifications_utils/countries/__init__.py
quis: Changed in 0359a59

risicle (approved):

PR #1407, opened 2026-06-25
add more type hints to `Template` and related classes

risicle:
This includes a change to `BaseLetterTemplate._address_block` to remove a weird side effect (noticed because we don't declare the `values` `Mapping` to be mutable).

### DIFF
--- a/notifications_utils/template.py
+++ b/notifications_utils/template.py
@@ -1,14 +1,15 @@
+import datetime
 import math
 from abc import ABC, abstractmethod
-from collections.abc import Mapping, Set
+from collections.abc import Mapping, Sequence, Set
 from contextlib import suppress
-from datetime import UTC, datetime
 from functools import lru_cache
 from html import unescape
 from os import path
 from typing import Any, Literal
 
 from jinja2 import Environment, FileSystemLoader, StrictUndefined
+from jinja2 import Template as jinja2_Template
 from markupsafe import Markup
 from ordered_set import OrderedSet
 from werkzeug.utils import cached_property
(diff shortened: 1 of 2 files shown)


notifications_utils/template.py
```diff
@@ -205,29 +209,29 @@ def values(self, value):
...
+    def content_with_placeholders_filled_in(self) -> str:
         # We always call SMSMessageTemplate.__str__ regardless of
         # subclass, to avoid any HTML formatting. SMS templates differ
         # in that the content can include the service name as a prefix.
         # So historically we’ve returned the fully-formatted message,
         # rather than some plain-text represenation of the content. To
         # preserve compatibility for consumers of the API we maintain
         # that behaviour by overriding this method here.
-        return SMSMessageTemplate.__str__(self)
+        return SMSMessageTemplate.__str__(self)  # type: ignore[arg-type]
```
quis: Shouldn’t be needed after https://github.com/alphagov/notifications-utils/pull/1406

notifications_utils/template.py
```diff
@@ -520,22 +538,31 @@ def subject(self) -> str:
...
-    jinja_template = template_env.get_template("email_template.jinja2")
+    jinja_template: jinja2_Template = template_env.get_template("email_template.jinja2")
 
     PREHEADER_LENGTH_IN_CHARACTERS: int = 256
 
+    govuk_banner: bool
+    complete_html: bool
+    brand_logo: Any
+    brand_text: Any
+    brand_colour: Any
```
quis: These will always be strings, if defined
```suggestion
    brand_logo: str | None
    brand_text: str | None
    brand_colour: str | None
```

quis (approved):

PR #1410, opened 2026-06-25
Remove `InsensitiveDict.from_keys()`

quis:
This was mostly used by the internals of the old version of `InsensitiveSet`, or as a way of getting the behavior of `InsensitiveSet` before it existed.

Using `InsensitiveSet` directly is the better approach in all cases now.


risicle (approved):
WFM.

This is something I was actually planning on doing next week.

PR #1413, opened 2026-06-26
Fix type errors in `celery.py`

quis:
See individual commits for details.

***

To test there are no errors locally:
```shell
mypy notifications_utils/celery.py
```

I’ve also installed this locally in the API repo and it’s not caused any test failures.


risicle (approved):
The flask `init_app` pattern is so toxic to type hinting.

PR #1419, opened 2026-06-30
Rp tombstone tests

rparke:


notifications_utils/clients/redis/request_cache.py
```diff
@@ -103,42 +118,75 @@ def _make_key(key_format, client_method, args, kwargs):
...
                 cached = self.redis_client.get(redis_key, skippable=True)
                 if cached:
-                    outer = msgpack.loadb(cached)
-                    if not outer.get("is_tombstone"):
-                        return msgpack.loadb(outer["value"])
+                    try:
+                        outer = msgpack.loads(cached)
+                        cached_is_tombstone = outer.get("is_tombstone")
+                        cached_sv = outer.get("schema_version")
+                        cached_value = msgpack.loads(outer["value"]) if outer.get("value") else None
```
rparke, line 137: I'm wondering whether we should have a utility class that is aware of the correct format for these new wrapped values and can just do all of the msgpack encoding for us.

If nothing else it should help us document in code what this is meant to look like and (hopefully) will reduce errors.

risicle: I think it might just be better to leave that concern "sealed" inside `RequestCache`. I wouldn't expect/want anything outside `RequestCache` to access its values.

risicle:
I note that I (we?) wrote the changelog entry discussing upgrading an existing application, *then* I added a bunch of schema-version-aware features. I don't know if the changelog entry needs updating.

rparke:
> I note that I (we?) wrote the changelog entry discussing upgrading an existing application, _then_ I added a bunch of schema-version-aware features. I don't know if the changelog entry needs updating.

yeh it does and I need to unpick those they got smushed together with a bunch of other fixes.

risicle:
Yep, the rollout procedure is a lot simpler now.

tests/clients/redis/test_redis_client.py
```diff
@@ -74,6 +57,54 @@ def test_delete_by_key_script(app, redis_client_with_live_instance, pattern, key
...
+    ],
+)
+def test_overwrite_by_pattern(
+    app, redis_client_with_live_instance, pattern, key_value_pairs, overwritten_key_value_pairs, number_of_matches
+):
+    for key, value in key_value_pairs:
+        redis_client_with_live_instance.redis_store.set(key, value)
+    assert redis_client_with_live_instance.overwrite_by_pattern(pattern, "overwritten") == number_of_matches
+    for key, value in overwritten_key_value_pairs:
+        assert redis_client_with_live_instance.get(key) == value
```
risicle, line 107: Worth adding a nasty little test that overwrites > 2500 keys?

tests/clients/redis/test_redis_client.py
```diff
@@ -426,3 +457,38 @@ def test_redis_flakey(caplog, app, mocked_redis_client):
...
+                    "timestamp": time_first_set,
+                    "is_tombstone": False,
+                    "schema_version": 1,
+                    "value": msgpack.dumps("return this"),
+                }
+            ),
+        )
+
+    with freeze_time("2018-7-7 15:59:59"):
+        redis_client_with_live_instance.set_if_timestamp_newer(
```
risicle: Worth asserting the return value here, which should indicate the success of the setting.

tests/clients/redis/test_request_cache.py
```diff
@@ -49,36 +61,72 @@ def test_set(
...
+            "value": msgpack.dumps("bar"),
+            "schema_version": 1,
+        }
+
+
+def test_set_if_timestamp_newer(redis_client_with_live_instance, live_cache):
+
+    @live_cache.set("my-key", schema_version=1)
+    def interrupting_function():
+        return "this was set later than the pesimistic timestamp"
```
risicle:
```suggestion
        return "this was set later than the pessimistic timestamp"
```

tests/clients/redis/test_redis_client.py
rparke: are there any nice tricks for generating a set of unique strings from a regex repeatedly?

tests/clients/redis/test_request_cache.py
```diff
@@ -281,31 +377,115 @@ def foo():
...
+                    {
+                        "timestamp": str(datetime.utcnow()),  # we cannot mock out times that redis lua scripts use
+                        "is_tombstone": False,
+                        "value": msgpack.dumps("foo"),
+                        "schema_version": 1,
+                    }
+                ),
+                "1-2-3-bar": msgpack.dumps(
+                    {
+                        "timestamp": str(datetime.utcnow()),  # we cannot mock out times that redis lua scripts use
```
risicle, line 417: Squeezing this into a test parameter will result in a really weird, ever-changing test name.

Given the number of places in the test body you have to do an `if force_delete` branch, I'd consider just making this two separate tests.

tests/clients/redis/test_redis_client.py
risicle: I think there are but we don't need them. We're not really testing the pattern matching itself here, we can just use something fairly trivial

rparke: done

tests/clients/redis/test_redis_client.py
```diff
@@ -426,3 +466,45 @@ def test_redis_flakey(caplog, app, mocked_redis_client):
...
+                    "is_tombstone": False,
+                    "schema_version": 1,
+                    "value": msgpack.dumps("do not return this"),
+                }
+            ),
+        )
+
+    assert redis_client_with_live_instance.get("foo") == msgpack.dumps(
+        {"timestamp": time_first_set, "is_tombstone": False, "schema_version": 1, "value": msgpack.dumps("return this")}
+    )
```
risicle, line 525: Suggest also testing the case where it's successfully overwriting the already-existing key.

risicle: And perhaps a case where the existing key is not msgpack-decodable?

PR #1420, opened 2026-07-01
`InsensitiveDict`: repimplement using nested dict instead of inheritance 

risicle:
Individual commits have more information, but in short this gives us an easier time in several ways - beyond giving us a more complete/"correct" implementation it should make type hinting work better.

Some changes to `InsensitiveSet` are included to make it work more like the new `InsensitiveDict` (e.g. it now takes a type parameter that should allow us to be more precise about where the keys are `str | None` and where they are just `str`.

Included are a lot of typing fixups to make `RecipientCSV`, `Template` and `Field` use `InsensitiveSet` and `InsensitiveDict` correctly, being able to differentiate between types that have `str` keys and those that have `str | None` keys.

### DIFF
--- a/notifications_utils/recipients.py
+++ b/notifications_utils/recipients.py
@@ -1,10 +1,11 @@
 import csv
 import sys
+from collections.abc import Callable, Container, Iterable, Iterator, Mapping, MutableMapping, Sequence
 from contextlib import suppress
 from functools import lru_cache
 from io import StringIO
-from itertools import islice
-from typing import cast
+from itertools import chain, islice
+from typing import Any, cast
 
 from ordered_set import OrderedSet
 from werkzeug.utils import cached_property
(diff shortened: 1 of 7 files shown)


notifications_utils/recipients.py
```diff
@@ -377,26 +384,35 @@ def _get_error_for_field(self, key, value):  # noqa: C901
...
 
-class Row(InsensitiveDict):
-    message_too_long = False
-    message_empty = False
+class Row(InsensitiveDict[str | None, Any]):
+    message_too_long: bool = False
+    message_empty: bool = False
+
+    index: int
+    recipient_column_headers: Any  # TODO
```
quis: Not sure if the `TODO` is because there was something tricky here but following the rabbit hole I think this will always be
```suggestion
    recipient_column_headers: Sequence[str]
```

quis (approved):
Looks good to me. Type checker is happy and I don’t expect this to be any slower than the previous implementation.

notifications_utils/recipients.py
risicle: I totally forgot about this

quis (approved):

PR #1422, opened 2026-07-07
Add type hints for letter timings and postage

quis:
To test this locally:
```shell
make check-types
```

### DIFF
--- a/notifications_utils/countries/data.py
+++ b/notifications_utils/countries/data.py
@@ -1,6 +1,6 @@
 import json
 import os
-from typing import Any
+from typing import Any, Literal
 
 
 def _load_data(filename: str) -> Any:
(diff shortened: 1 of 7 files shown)


notifications_utils/countries/data.py
```diff
@@ -66,9 +66,25 @@ def find_canonical(item: dict, graph: dict, key: str):
 
 
 class Postage:
```
risicle, line 68: Gotta wonder if we should be turning all of these into `Enum`s

quis: I did try that, but then all the places that use this code have to start doing `postage.value` and it felt like a bit of a messy change.

risicle: Ick

risicle (approved):

PR #1423, opened 2026-07-07
`AbstractInsensitiveSet`, `AbstractInsensitiveDict`: split into mutable and immutable variants

risicle:
One being a subtype of the other, and, on concrete classes, each knowing about its mutable/immutable counterpart so it can special-case handling for performance.

Use of the immutable types should allow us to reduce the ambiguity of function call interfaces and prevent code from making unexpected mutations of possibly-shared data structures.

Have left aliases in here to allow `AbstractMutableInsensitiveSet` and `AbstractMutableInsensitiveDict` to still be imported as `AbstractInsensitiveSet` and `AbstractInsensitiveDict` respectively. And `InsensitiveSet` and `InsensitiveDict` still refer to their mutable variants so shouldn't lead to any behaviour change.

### DIFF
--- a/notifications_utils/insensitive_dict.py
+++ b/notifications_utils/insensitive_dict.py
@@ -417,10 +417,12 @@ class InsensitiveSet[T: str | None](ImmutableInsensitiveSet[T], AbstractMutableI
     _immutable_type: type[ImmutableInsensitiveSet[T]] = ImmutableInsensitiveSet
 
 
-ImmutableInsensitiveSet._mutable_type = InsensitiveSet._mutable_type = InsensitiveSet  # type: ignore[misc]
+ImmutableInsensitiveSet._mutable_type = InsensitiveSet  # type: ignore[misc]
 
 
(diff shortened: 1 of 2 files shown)


quis (commented):
Good change.

Probably need to parametrize/duplicate the tests for set/`del`/`pop`/`update` here to show that they raise the expected exception and avoid any mutation:
https://github.com/alphagov/notifications-utils/blob/8a1a23d49ad8a9f64f4c6bfe5176c2ad60be78f8/tests/test_insensitive_dict.py#L64-L164

risicle:
Hmm it looks like any tests I wrote for `ImmutableInsensitiveDict` seemed to get squashed out in a rebase... will (re-?)add.

rparke (approved):

rparke:
I clicked approve but ... do we need a version bump even though this shouldn't change any of the calls to InsensitiveDict at this point?

risicle:
I might follow up with one.

PR #1426, opened 2026-07-14
Removing reference to requirements.txt as utils doesn't have that

whpearson:
I'm guessing this just bought a long and isn't required here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ help:
 .PHONY: freeze-requirements
 freeze-requirements: ## Pin all test requirements including sub dependencies into requirements_for_test.txt
 	uv pip compile requirements_for_test.in pyproject.toml --output-file requirements_for_test.txt $(EXTRA_UV_PIP_COMPILE_FLAGS)
+	uv pip sync requirements_for_test.txt
 	uv pip install -e .
 
 .PHONY: refreeze-requirements
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,6 @@ refreeze-requirements: ## Upgrade unpinned requirements
 .PHONY: bootstrap
 bootstrap: ## Build project
 	uv pip install -r requirements_for_test.txt
-	uv pip sync requirements_for_test.txt
 	uv pip install -e .
 
 .PHONY: bootstrap-with-docker
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,6 @@ help:
 .PHONY: freeze-requirements
 freeze-requirements: ## Pin all test requirements including sub dependencies into requirements_for_test.txt
 	uv pip compile requirements_for_test.in pyproject.toml --output-file requirements_for_test.txt $(EXTRA_UV_PIP_COMPILE_FLAGS)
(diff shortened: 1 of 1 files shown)


spatel033 (approved):

Makefile
```diff
@@ -9,7 +9,6 @@ help:
 .PHONY: freeze-requirements
 freeze-requirements: ## Pin all test requirements including sub dependencies into requirements_for_test.txt
 	uv pip compile requirements_for_test.in pyproject.toml --output-file requirements_for_test.txt $(EXTRA_UV_PIP_COMPILE_FLAGS)
-	uv pip sync requirements.txt
```
quis, line 12: Reason is given in this commit: https://github.com/alphagov/notifications-utils/commit/295d5e9e66120caf4d77f706f5f35385d060b8a6

But yeah it should be `requirements_for_test.txt` not `requirements.txt`

Makefile
```diff
@@ -9,7 +9,6 @@ help:
 .PHONY: freeze-requirements
 freeze-requirements: ## Pin all test requirements including sub dependencies into requirements_for_test.txt
 	uv pip compile requirements_for_test.in pyproject.toml --output-file requirements_for_test.txt $(EXTRA_UV_PIP_COMPILE_FLAGS)
```
quis, line 11:
```suggestion
	uv pip compile requirements_for_test.in pyproject.toml --output-file requirements_for_test.txt $(EXTRA_UV_PIP_COMPILE_FLAGS)
    uv pip sync requirements_for_test.txt	
```

***

From 295d5e9:

> **Sync requirements after freezing**
>
>`sync` makes sure the specified dependencies – and only the specified dependencies – are installed. `install` installs the specified dependencies on top of whatever is already in the environment.
>
> It ensures that the freeze has produced a functiona environment.
>
> It’s what we do as the last part of the freeeze step in other repos, for
example: https://github.com/alphagov/notifications-admin/blob/9d22ced9dca1abc9567cc713d2986d12c8dc026c/Makefile#L79C2-L79C30

whpearson: Looks like you have spaces instead of tabs here.

quis:
```suggestion
	uv pip compile requirements_for_test.in pyproject.toml --output-file requirements_for_test.txt $(EXTRA_UV_PIP_COMPILE_FLAGS)
	uv pip sync requirements_for_test.txt
```

Makefile
```diff
@@ -19,6 +19,7 @@ refreeze-requirements: ## Upgrade unpinned requirements
 .PHONY: bootstrap
 bootstrap: ## Build project
 	uv pip install -r requirements_for_test.txt
+	uv pip sync requirements_for_test.txt
```
quis: There’s a case for removing `install` here and swapping for `sync` but for now we should just be consistent with [other repos](https://github.com/alphagov/notifications-admin/blob/e1ff400ca996edd9fbf4bdaece0632bb8d6c80d9/Makefile#L16) and do `install`. No point in doing both because that’s duplicative
```suggestion
```

quis (approved):

PR #1428, opened 2026-07-17
Rp revert insensitivedict changes

rparke:


quis:
This is running fine in prod now, not needed

PR #1433, opened 2026-07-22
 add `RelaxedContainerJSONEncoder` to handle encoding of alternative container types in encoding

risicle:
This is an annoyance exposed by #1420 making `InsensitiveDict` no longer inherit directly from concrete `dict` - the default `json` encoder refuses to encode it.

Sadly `requests` doesn't have a nice way of substituting the encoder used for its `json=` kwarg, so to make use of this encoder we'll have to perform the encoding ourselves, provide it to the `data=` kwarg and ensure the `Content-type: application/json` request header is used.

Flask, on the other hand, does allow supplying a custom encoder for its automatic json encoding of responses, and so we mix our `RelaxedContainerJSONEncoder` in to flask's default (already enhanced) json encoder to make `FlaskRelaxedContainerJSONEncoder` which can be used with a flask app's `json_provider_class` attribute.

### DIFF
--- a/notifications_utils/json.py
+++ b/notifications_utils/json.py
@@ -1,6 +1,8 @@
 from collections.abc import Iterator, Mapping, Sequence
 from json import JSONEncoder
 
+from flask.json.provider import DefaultJSONProvider as flask_DefaultJSONProvider
+
 type RelaxedJsonType = None | bool | int | float | str | Sequence["RelaxedJsonType"] | Mapping[str, "RelaxedJsonType"]
 
 
@@ -21,6 +23,7 @@ class RelaxedContainerJSONEncoder(JSONEncoder):
     A JSONEncoder subclass that will turn (almost) any encountered
     Sequence into a tuple and any Mapping into a plain dict.
     """
+
     def default(self, obj: RelaxedJsonType) -> StrictJsonType:
         match obj:
             case bool() | int() | float() | str() | None:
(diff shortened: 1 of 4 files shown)


notifications_utils/json.py
```diff
@@ -0,0 +1,47 @@
+from collections.abc import Iterator, Mapping, Sequence
+from json import JSONEncoder
+
+from flask.json.provider import DefaultJSONProvider as flask_DefaultJSONProvider
+
+type RelaxedJsonType = None | bool | int | float | str | Sequence["RelaxedJsonType"] | Mapping[str, "RelaxedJsonType"]
```
quis, line 6: Worth replacing this: 
https://github.com/alphagov/notifications-utils/blob/856f5ba4793d97e220604152176d83222ac99089/notifications_utils/clients/redis/request_cache.py#L9

With the version from this file?

(fine to do as a followup)

notifications_utils/json.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+type StrictJsonType = (
+    None
+    | bool
+    | int
+    | float
+    | str
+    | list["StrictJsonType"]
+    | tuple["StrictJsonType", ...]
```
quis, line 16: What is the ellipsis doing here?

quis (approved):

notifications_utils/json.py
risicle: Maybe. The version from this file is lifted directly from the requests source and intended to match that pretty exactly. Beware there are caveats about mypy's ability to handle recursive types properly though.

notifications_utils/json.py
risicle: Tuples interpret their generic types in a special way. `tuple[Foo]` means a tuple with a single element of type `Foo`. `tuple[Foo, Bar]` is a two element tuple with types `Foo` and `Bar` respectively. For a tuple of a single type but unspecified length, you have to give it a single element and an ellipsis.

PR #1435, opened 2026-07-23
add `test_celery` tests to a single xdist group

rparke:
when testing locally I've run into a few cases of these tests being flakey and I believe it's because multiple tests are trying to make calls to celery at the same time and causing ... some sort of issue (I haven't dug down into what it is).

At any rate adding these tests to a single xdist_group shouldn't negatively impact how they run and it doesn't seem to impact speed much.

### DIFF
--- a/tests/logging/test_celery.py
+++ b/tests/logging/test_celery.py
@@ -17,6 +17,7 @@ class Config:
         return getattr(self, key, default)
 
 
+@pytest.mark.xdist_group(name="test_celery")
 @patch("notifications_utils.logging.celery.dictConfig")
 @patch("notifications_utils.logging.celery.config", Config())
 def test_setup_logging_connect_success(mock_dict_config):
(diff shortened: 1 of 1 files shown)


tests/logging/test_celery.py
```diff
@@ -28,6 +29,7 @@ def test_setup_logging_connect_success(mock_dict_config):
     mock_dict_config.assert_called_once()
 
 
+@pytest.mark.xdist_group(name="test_celery")
 def assert_command_has_outputs(tmp_path, command, filename, expected_messages, unexpected_messages=None, env=None):
```
quis: Pytest won’t run this function  because it doesn’t start with `def test_`. It’s a helper which is used by the functions below. You don’t need to decorate it


```suggestion
def assert_command_has_outputs(tmp_path, command, filename, expected_messages, unexpected_messages=None, env=None):
```

rparke: I got carried away 😝

quis (approved):
I set the tests to run on a loop. After a few mins on `main` they failed with the Celery timeout. After 20 mins on this branch they still hadn‘t failed. I reckon you’ve fixed it.

PR #1437, opened 2026-07-23
`RequestCache`: use `RelaxedContainerJSONEncoder` to encode cached payloads

risicle:
We don't really want to run into any un-encodable values when caching.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 124.2.0
+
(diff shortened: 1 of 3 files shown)


quis (approved):
In theory this shouldn’t be needed because the data it’s caching will be deserialised from JSON. But it can’t hurt/might stop weird things from happening in the future.

risicle:
`RequestCache` isn't always used for caching raw requests as far as I can remember.

PR #1438, opened 2026-07-23
Remove `RecipientCSV.rows_as_list` and `RecipientCSV.get_rows()`

quis:
If you try to look at `RecipientCSV.rows_as_list` before `RecipientCSV.rows` has been accessed you will incorrectly see that the file has no rows. This change makes it so you will get an `AttributeError` instead.

We still need to store the list, but we mark the attribute as internal with an underscore.

***

`RecipientCSV.get_rows()` doesn’t cache its results, or make its processing interruptible, so code which consumes this library should not be calling it. We can imply this by prefixing its name with an underscore.

***

Any code which needs to loop through the rows should instead loop over `RecipientCSV.rows` or the `RecipientCSV` instance itself, rather than using these now-internal properties or methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 125.0.0
+
+* Removes `RecipientCSV.rows_as_list` and `RecipientCSV.get_rows()`. Use `for row in recipient_csv_instance:`, `len(recipient_csv_instance)`, etc. instead
+
 ## 124.2.1
 
 * Fix antivirus scan when the uploaded file is a Werkzeug `FileStorage` (requests 2.34)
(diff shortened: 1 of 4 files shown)


rparke:
I'm going to be slightly annoying and ask if we can get at least a draft of the version bump to the apps ready?

We are already 3 major versions behind on `notifications-api` (I haven't checked the other apps) and this would make it 4 and ... it's a long queue of stuff in utils that needs to go out 😆 

(otherwise LGTM though)

quis:
@rparke once https://github.com/alphagov/notifications-api/pull/4922 is merged we’ll be up-to-date in admin and API at least

rparke (approved):

PR #1440, opened 2026-07-28
fix: antivirus scan for uploaded files in admin

joybytes:
As part of vulnerability work, I was trying to upgrade notifications-admin
One of the dependencies that were bumped was `requests`.

It seems like the last time we tried to bump admin with this dependency, we ran into this issue [here](https://govuk-notify-gds.sentry.io/issues/7626487773/?project=4504949328838656&query=is%3Aunresolved&referrer=issue-stream):
`TypeError: a bytes-like object is required, not 'FileStorage'`

That's because after dependency upgrades, `requests 2.34` no longer accepts the way we were sending files chosen in the browser.

So with this change fixes how we build that HTTP upload for admin, without changing what each app passes in. Document-download-api still passes plain bytes; that path is unchanged.

https://tedboy.github.io/flask/generated/generated/werkzeug.FileStorage.html

https://trello.com/c/AY4TcgE0/718-vulnerability-review-replenishment-28-july-3-august-2026-pigeon

### DIFF
--- a/notifications_utils/clients/antivirus/antivirus_client.py
+++ b/notifications_utils/clients/antivirus/antivirus_client.py
@@ -1,6 +1,7 @@
 import requests
 from flask import current_app, request
 from flask.ctx import has_request_context
+from werkzeug.datastructures import FileStorage
 
 
 class AntivirusError(Exception):
@@ -18,6 +19,18 @@ class AntivirusError(Exception):
             status_code = 503
 
         return cls(message, status_code)
+
+
+def _multipart_file(document_stream):
+    """
+    Format the file for the HTTP library we use to call antivirus.
+
+    - Admin uploads arrive as a FileStorage, so we send requests the stream it wraps.
+    - Document-download-api sends plain bytes, which requests can take as-is.
+    """
+    if isinstance(document_stream, FileStorage):
+        return document_stream.stream
+    return document_stream
 
 
 class AntivirusClient:
@@ -41,7 +54,7 @@ class AntivirusClient:
             response = self.requests_session.post(
                 f"{self.api_host}/scan",
                 headers=headers,
-                files={"document": document_stream},
+                files={"document": _multipart_file(document_stream)},
             )
 
             response.raise_for_status()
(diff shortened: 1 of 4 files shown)


notifications_utils/clients/antivirus/antivirus_client.py
```diff
@@ -20,6 +20,22 @@ def from_exception(cls, e):
...
+
+    - Admin: user uploaded a file in the browser. We send name + contents + type as a tuple.
+      (Sending the upload object as-is broke staging after the requests upgrade.)
+    - Document-download-api: plain bytes — send them unchanged.
+    """
+    filename = getattr(document_stream, "filename", None)
+    if filename is not None:
+        stream = getattr(document_stream, "stream", document_stream)
+        content_type = getattr(document_stream, "content_type", None)
+        return filename, stream, content_type
```
quis: This matches what’s in the Requests docs:
> files – (optional) Dictionary of `'name': file-like-objects` (or `{'name': file-tuple}`) for multipart encoding upload. file-tuple can be a 2-tuple (`'filename', fileobj`), 3-tuple (`'filename', fileobj, 'content_type'`) or a 4-tuple (`'filename', fileobj, 'content_type', custom_headers`), where `'content_type'` is a string defining the content type of the given file and `custom_headers` a dict-like object containing additional headers to add for the file.

— https://requests.readthedocs.io/en/latest/api/

quis: The previous behaviour would pass the whole `FileStorage` object to Requests, and because `FileStorage` [quacks like a duck](https://en.wikipedia.org/wiki/Duck_typing):
> All the attributes of the wrapper stream are proxied by the file storage so it’s possible to do storage.read() instead of the long form storage.stream.read().

…older versions of Requests would accept it.

We could make this simpler and replicate the previous behaviour by doing:
```suggestion
    if isinstance(document_stream, FileStorage):
        return document_stream.stream
```

Your existing tests would cover this, whereas your existing tests don’t cover the full range of permutations possible with the use of `getattr`.

quis (approved):
Approving this based on:
- replicating the problem locally
- proving that this PR, as-is, fixes the problem

I think there’s a potential improvement (see comments) but up to you if you want to spend the time on this

notifications_utils/clients/antivirus/antivirus_client.py
joybytes: great suggestion, I have updated the code adding your suggestion

joybytes:
@quis Thanks for the review,
I applied your suggestion.

It will probably need another thumbs up for approval (if you agree with the changes of course)

quis (approved):

PR #1442, opened 2026-08-10
Correctly count complex unicode characters in text messages

quis:
Our existing character count makes the assumption that 1 character in a message will be encoded and charged as 1 character by our providers.

This is true for the limited number of non-GSM characters we currently support, such as `Ŵ` and `ŷ`.

Once we start supporting a wider range of unicode characters, this assumption breaks because some characters require multiple code points to encode.

An extreme example is 👨‍👩‍👧‍👦, which is effectively:
> `👨` + `[joiner]` + `👩` + `[joiner]` + `👧` + `[joiner]` + `👦`

Each emoji takes 2 UTF-16 code points, and each joiner takes 1, so:
> 2 + 1 + 2 + 1 + 2 + 1 + 2 = 11

We encode the content as `utf-16-le` (little endian) instead of `utf-16` because that gives bytes without a preceding BOM (byte-order mark). Each code point in UTF-16 takes 2 bytes, so we floor divide by 2 to get the number of code points. This is what we are charged for.

---

All the test examples come from [📝 Multilingual SMS - feature doc](https://docs.google.com/document/d/1SpY9ecLJoA2zIvvxKqBFUXBRba6QreI0ihtkVX1ogWQ/edit?tab=t.23t60h9vu6k0#heading=h.dbyeyt1vcq4m)

I compared them against [QuickSMS SMS & RCS Character Counter](https://www.quicksms.com/sms-character-counter) (which was recommended by one of our providers) and verified that the counts match.

---

An alternative way of doing this would be:
```python
def content_count(self) -> int:
    # >= 65536 is unicode plane 2 
    sum(1 if ord(c) < 65536 else 2 for c in self.content_with_placeholders_filled_in)
```

However when I tested this it was 2.5× slower.

---

This pull request also fixes a number of bugs with the way we count characters, these are detailed in separate commits.

---

This pull request does not change how we encode messages, so it doesn’t yet change what we actually send.

https://trello.com/c/D0gIzzY6/683-prepare-the-billing-code-and-fragment-counter-for-unicode-support-in-sms

### DIFF
--- a/notifications_utils/sanitise_text.py
+++ b/notifications_utils/sanitise_text.py
@@ -71,9 +71,12 @@ class SanitiseText:
             # For a full list of the types, see here: https://www.compart.com/en/unicode/decomposition.
             # If it's got a mapping, we're not sure how best to downgrade it, so just see if it's in the
             # REPLACEMENT_CHARACTERS map. If not, then it's probably a letter with a modifier, eg á
-            # ASSUMPTION: The first character of a combined unicode character (eg 'á' == '0061 0301')
-            # will be the ascii char
-            return cls.get_unicode_char_from_codepoint(decomposed.split()[0])
+            # If this is the case then the first character of a combined unicode character (eg 'á' == '0061 0301')
+            # will be an ASCII char in ALLOWED_CHARACTERS
+            first_character_of_decomposition = cls.get_unicode_char_from_codepoint(decomposed.split()[0])
+            if first_character_of_decomposition in cls.ALLOWED_CHARACTERS:
+                return first_character_of_decomposition
+            return None
         else:
             # try and find a mapping (eg en dash -> hyphen ('–': '-')), else return None
             return cls.REPLACEMENT_CHARACTERS.get(c)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 125.0.1
+
+* Fixes a number of bugs with counting characters in text message templates
+
 ## 125.0.0
 
 * Removes `RecipientCSV.rows_as_list` and `RecipientCSV.get_rows()`. Use `for row in recipient_csv_instance:`, `len(recipient_csv_instance)`, etc. instead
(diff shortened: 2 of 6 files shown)


notifications_utils/sanitise_text.py
```diff
@@ -71,9 +71,12 @@ def downgrade_character(cls, c: str) -> str | None:
             # For a full list of the types, see here: https://www.compart.com/en/unicode/decomposition.
             # If it's got a mapping, we're not sure how best to downgrade it, so just see if it's in the
             # REPLACEMENT_CHARACTERS map. If not, then it's probably a letter with a modifier, eg á
-            # ASSUMPTION: The first character of a combined unicode character (eg 'á' == '0061 0301')
-            # will be the ascii char
-            return cls.get_unicode_char_from_codepoint(decomposed.split()[0])
+            # If this is the case then the first character of a combined unicode character (eg 'á' == '0061 0301')
+            # will be an ASCII char in ALLOWED_CHARACTERS
+            first_character_of_decomposition = cls.get_unicode_char_from_codepoint(decomposed.split()[0])
```
joybytes, line 76: One thing I noticed, which is pre-existing

In `get_unicode_char_from_codepoint` we assume 4-digits, but some CJK breakdowns are longer (e.g. 242EE). Those would raise here instead of falling through to `?`.

joybytes (approved):

notifications_utils/sanitise_text.py
quis: It doesn’t raise because `𤋮` doesn’t decompose into multiple characters like `á` does:
```python
>>> import unicodedata
>>> unicodedata.decomposition("𤋮")
''
```

I had a look at https://www.unicode.org/Public/UNIDATA/UnicodeData.txt

There is exactly one example in all of unicode where a character decomposes to characters having more than 4 hex digits.

That is `105E4` ‘Todhri Letter U’ which decomposes to `105E6` ’Todhri Letter O’ and `0307` ‘combining dot above’.

Todhri is an obsolete Albanian script which was introduced to unicode in 2024. It’s not yet made it into Python so we can’t write a test case for it.

PR #1443, opened 2026-08-10
Allow a wider range of unicode characters in text messages 

quis:
There is some refactoring and bug fixing in this PR so I recommend reviewing commit-by-commit. The substance of the change is as follows.

# What’s changing

This changes our text message encoding from:
- allowing GSM characters
- allowing a small set of replaceable characters
- converting everything else to '?'

To:
- allowing any character
- replacing the same set of replaceable characters

This is so we can support languages which use characters not in GSM, such as Arabic, Cyrillic, etc.

For simplicity of proposition we are choosing to allow any unicode character, rather than a subset such as only alphabets (and not, for example emoji).

## Notes on decomposition

We are choosing not to decompose characters like `á` to `a` like we did previously. While this might save money in some select cases, it will also change the meaning of the message. So we are prioritising correctness over cost savings in these case.

There are still some replacements we can make where:
- it doesn’t really change the meaning
- the use of a character might be the only difference between the message being GSM-encodable or not

Examples of these are em dash (—) to hyphen (-) or typographic quotes (“) to straight quotes (").

For consistency we make these replacements in all cases because:
- this matches what we do at the moment
- it means including another non-GSM character like á in a message doesn’t cause other characters in the message to start changing

# How to roll this out

1. Update the API: https://github.com/alphagov/notifications-api/pull/4937
2. Update the admin app: https://github.com/alphagov/notifications-admin/pull/6037

There will be a short window between 1. and 2. where the API messages sends messages with the new encoding but the admin app displays them with the current encoding. I think this is acceptable.

## Breaking changes

These properties and methods are no longer needed so this PR removes them. App code may need updating accordingly:
* `SanitiseSMS.downgrade_character()` and `SanitiseSMS.encode_char()` (should not be used anywhere)
* `SanitiseSMS.WELSH_DIACRITICS` and `SanitiseSMS.ALLOWED_CHARACTERS` (should not be used anywhere)
* `SanitiseSMS.get_non_gsm_characters()` (use `template.non_gsm_characters` instead)


notifications_utils/sanitise_text.py
```diff
@@ -55,6 +50,72 @@ def get_unicode_char_from_codepoint(codepoint: str) -> str:
...
+
+    CHARACTERS_NOT_REQUIRING_UNICODE: Set[str] = GSM_CHARACTERS | set(SanitiseText.REPLACEMENT_CHARACTERS)
+
+    @classmethod
+    def encode(cls, content: str) -> str:
+        return "".join(cls.REPLACEMENT_CHARACTERS.get(c, c) for c in content)
+
+    @classmethod
+    def get_non_compatible_characters(cls, content: str) -> Set:
+        return set()
```
joybytes: I assume that with this change we'll need to remove the usage on-api and -admin, is that correct?

joybytes: ah yes, just saw the roll out plan

quis: Both will work without any code changes. We may want to remove this function entirely though.

I think the only question is whether we should keep logging for non-GSM characters here:
https://github.com/alphagov/notifications-api/blob/9b93761197e8aa2b84a54c3cea57719272a3332d/app/delivery/send_to_providers.py#L63-L74

joybytes (dismissed):
I honestly think this looks good.

I appreciated the thorough comments around the code, it made things easier to understand.

risicle (approved):

PR #1449, opened 2026-08-18
dependency bump

rparke:


quis (approved):
You don’t technically need to make a new version here because these changes aren’t going to affect any of the code which runs in production.

PR #1461, opened 2026-09-03
add scripts/reduce_ofcom_prefix_list.py

risicle:
This is intended for use in extracting the relevant information from an OFCOM csv of phone number prefixes and preparing it for efficient prefix matching.

See its argparse help strings for more information.

### DIFF
--- a/scripts/reduce_ofcom_prefix_list.py
+++ b/scripts/reduce_ofcom_prefix_list.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python
+
+import csv
+import re
+import sys
+import typing
+from collections.abc import Callable, Container, Iterable, Sequence
+from io import TextIOBase
+from itertools import groupby
+
+
+def _compressed_sorted_prefixes_iter(prefixes: Iterable[str]) -> Iterable[str]:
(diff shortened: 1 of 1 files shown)


scripts/reduce_ofcom_prefix_list.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+if __name__ == "__main__":
+    import argparse
+
+    parser = argparse.ArgumentParser(
+        description="Provided with an OFCOM csv of phone number prefixes, will filter to those "
+        "of a type matching FILTER, normalize the number prefixes and if possible "
+        "compress the list to its minimal equivalent prefix list before outputting "
+        "the remaining prefixes to stdout, newline-separated and sorted."
```
rparke, line 85: We are assuming that this is ordered and only contains the blocks you want to compress right? ("protected" and maybe "free")

risicle: This script will do all that for you, as long as you use e.g. `--filter Protected`

rparke: wonderful

rparke (approved):

PR #1463, opened 2026-09-04
`PhoneNumber`: add `is_number_in_S7_protected_range()` method

risicle:
This uses the script added in #1461, to bundle a reduced list of "protected" phone prefixes based on OFCOM's `S7.csv` file.

It also includes a new method in `PhoneNumber` to efficiently load & query it.


notifications_utils/ofcom.py
```diff
@@ -0,0 +1,32 @@
...
+        importlib_resources.files("notifications_utils")
+        .joinpath("data/ofcom/S7_protected_prefixes.txt")
+        .read_text()
+        .split()
+    )
+    prefixes.sort()
+    return tuple(prefixes)  # ensure cached result can't be mutated
+
+
+def is_number_in_S7_protected_range(number: str) -> bool:
```
quis: I feel like this function would go nicely alongside:
https://github.com/alphagov/notifications-utils/blob/91da6f8f67ae0ea2823c790133d44a3d0909fda7/notifications_utils/recipient_validation/phone_number.py#L224-L229

Unless the way this function will be used makes that inconvenient

risicle: Umm... I guess it would.

Would I still give it the same name?

quis: Assuming the usage will be similar to https://github.com/alphagov/notifications-api/pull/4957/changes#diff-62c0fd99f002d23e05e91c9f98d00928a1ab89aa52e985f91b1adb9803e9b75aL183

I would name it like:

```
if number.is_in_S7_protected_range():
```

quis (approved):

notifications_utils/ofcom.py
risicle: Meh well I've left the name the same because all the other methods keep the word `number` in them.

quis (approved):

tests/recipient_validation/test_phone_number.py
```diff
@@ -623,6 +639,39 @@ def test_validation_fails_for_unsupported_country_codes(self):
...
+            ("07011100876", False),
+            ("07034700000", True),
+            ("07034701000", True),
+            ("07034710000", False),
+            ("07034777777", True),
+            ("07074971099", False),
+            ("07075971077", True),
+            ("07999999999", False),
+            # non-uk number
+            ("+1 202-483-3000", False),
```
DilwoarH, line 655: do you want to test `+447...`, `447...`, and `7...`?

risicle: That's kinda handled by the rest of `PhoneNumber` I think...

PR #1465, opened 2026-09-10
Add sadd and expire methods to redis client

BlessedDev:
These are useful redis functionality to add to the `RedisClient` and would help facilitate the ongoing work to log API key usage.

https://trello.com/c/eiKDOLcB/764-record-api-key-usage-in-redis

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 131.1.0
+
+* `Redis-client`: Add `sadd` and `expire` methods.
+
 ## 131.0.0
 
 * Removes `RecipientCSV.rows`, `RecipientCSV._rows_as_list` (use, for example `for r in RecipientCSV(…)` instead)
--- a/notifications_utils/clients/redis/redis_client.py
+++ b/notifications_utils/clients/redis/redis_client.py
@@ -446,6 +446,24 @@ class RedisClient:
 
         return None
 
+    def expire(
+        self,
+        key: str,
+        seconds: int,
+        raise_exception: bool = False,
+        always_raise: tuple[type[BaseException], ...] | type[INSTANCE_DEFAULT] = INSTANCE_DEFAULT,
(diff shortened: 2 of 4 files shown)


risicle:
Hold on, I think I need a refresher on what we're doing with the key usage logging story. Last I knew, the only way it was going to use redis was through the normal cache decorator.

BlessedDev:
> Hold on, I think I need a refresher on what we're doing with the key usage logging story. Last I knew, the only way it was going to use redis was through the normal cache decorator.

The aim is to have hourly buckets in `redis` to store a service's API key at each endpoint usage for each hour. A `redis` set would help to prevent duplicates. These commands would help with that.

I am going to have a secondary kickoff as there seems to be some issue with clarity on the caching implementation specific to API key logging.

However I think this functionality are good options to have.

risicle:
Ok this sounds like it's completely different to how @rparke and I planned it

risicle:
> However I think this functionality are good options to have.

I'd generally prefer not to add methods that we don't use.

rparke:
> Ok this sounds like it's completely different to how @rparke and I planned it

This should just be a thin caching layer reusing what we already have, yeh. @BlessedDev we can have a call

rparke:
> > Hold on, I think I need a refresher on what we're doing with the key usage logging story. Last I knew, the only way it was going to use redis was through the normal cache decorator.
> 
> 
> 
> The aim is to have hourly buckets in `redis` to store a service's API key at each endpoint usage for each hour. A `redis` set would help to prevent duplicates. These commands would help with that.
> 
> 
> 
> I am going to have a secondary kickoff as there seems to be some issue with clarity on the caching implementation specific to API key logging.
> 
> 
> 
> However I think this functionality are good options to have.

The key should have the hour-truncated timestamp in it. And should just have a 1 hour expiry.

BlessedDev:
The chosen implementation for logging API keys doesn't need these methods so I will be closing the PR.

PR #1466, opened 2026-09-10
Update `is_number_in_S7_protected_range` to return False if number is a TV number or smoke test number

rparke:
TV numbers and smoke test numbers are in the protected blocks. We don't actually want to block sending to these as they will break our users integration/smoke tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 131.1.0
+
+* Update `is_number_in_S7_protected_range` to return False if number is a TV number or smoke test number
+
 ## 131.0.0
 
 * Removes `RecipientCSV.rows`, `RecipientCSV._rows_as_list` (use, for example `for r in RecipientCSV(…)` instead)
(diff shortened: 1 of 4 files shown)


risicle:
I think we should keep this method doing what it says on the tin, not slip surprising logic into it. If we need more predicates we can add them separately and leave the caller to marry them together.

risicle:
Perhaps we should remove the underscore from `_is_tv_number` and expose it as a public method.

The smoke test numbers _are_ TV numbers, so I'm not sure we need to match them explicitly do we?

rparke:
> Perhaps we should remove the underscore from `_is_tv_number` and expose it as a public method.
> 
> The smoke test numbers _are_ TV numbers, so I'm not sure we need to match them explicitly do we?

it was redundancy in case anything changed, but you're probably right that we can just leave them as TV numbers

rparke:
> I think we should keep this method doing what it says on the tin, not slip surprising logic into it. If we need more predicates we can add them separately and leave the caller to marry them together.

happy to close this then and just add that into the API

PR #1467, opened 2026-09-10
make `PhoneNumber._is_tv_number` a public method

rparke:
we want `notifications-api` to be able to use this method directly to assist with AIT mitigations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 131.1.0
+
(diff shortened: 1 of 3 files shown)


risicle (approved):

risicle:
You'll probably want to bump/changelog.

rparke:
> You'll probably want to bump/changelog.

didn't think it was strictly necessary at first - but you're right it DOES strictly change the API even though nothing was using it

risicle:
Well, just so you're able to pull it in to apps at all.

risicle (approved):

PR #1468, opened 2026-09-11
`PhoneNumber.validate` should reject numbers to ofcom protected block when `block_ofcom_protected_block = True`

rparke:
Add `block_ofcom_protected_block` as an optional argument to `PhoneNumber.validate`

our current pattern for all service permission specific validation is to put it in the `PhoneNumber.validate` method. We should do the same here for permission-based blocking of notifications to protected blocks, it's consistend vs using the new public methods (`is_number_in_S7_protected_range` and `is_tv_number` directly whenever we try to validate.

### DIFF
--- a/notifications_utils/recipient_validation/phone_number.py
+++ b/notifications_utils/recipient_validation/phone_number.py
@@ -103,9 +103,24 @@ class PhoneNumber:
         if str(self.number.country_code) not in COUNTRY_PREFIXES:
             raise InvalidPhoneError(code=InvalidPhoneError.Codes.UNSUPPORTED_COUNTRY_CODE)
 
-    def validate(self, allow_international_number: bool = False, allow_uk_landline: bool = False) -> None:
+    def _raise_if_service_who_blocked_it_tries_sending_to_ofcom_protected_ranges(
+        self, block_ofcom_protected_ranges: bool = False
+    ):
+        if block_ofcom_protected_ranges:
+            if self.is_number_in_S7_protected_range() and not self.is_tv_number(self.number):
+                raise InvalidPhoneError(code=InvalidPhoneError.Codes.INVALID_NUMBER)
+
+    def validate(
+        self,
+        allow_international_number: bool = False,
+        allow_uk_landline: bool = False,
+        block_ofcom_protected_ranges: bool = False,
+    ) -> None:
         self._raise_if_service_cannot_send_to_international_but_tries_to(allow_international=allow_international_number)
         self._raise_if_service_cannot_send_to_uk_landline_but_tries_to(allow_uk_landline=allow_uk_landline)
+        self._raise_if_service_who_blocked_it_tries_sending_to_ofcom_protected_ranges(
+            block_ofcom_protected_ranges=block_ofcom_protected_ranges
+        )
         self._raise_if_unsupported_country()
 
     @staticmethod
--- a/tests/recipient_validation/test_phone_number.py
+++ b/tests/recipient_validation/test_phone_number.py
@@ -259,6 +259,21 @@ mock_S7_prefixes = (
 )
 
 
+@pytest.fixture(scope="function")
+def mock_get_S7_protected_prefixes(mocker):
+    return mocker.patch(
+        "notifications_utils.recipient_validation.phone_number.get_S7_protected_prefixes",
+        return_value=(
+            "70346",
+            "703470",
+            "703477",
+            "70348",
+            "703490",
+            "7075",
+        ),
+    )
+
+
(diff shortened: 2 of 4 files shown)


notifications_utils/recipient_validation/phone_number.py
```diff
@@ -103,9 +103,22 @@ def _raise_if_unsupported_country(self) -> None:
         if str(self.number.country_code) not in COUNTRY_PREFIXES:
             raise InvalidPhoneError(code=InvalidPhoneError.Codes.UNSUPPORTED_COUNTRY_CODE)
 
-    def validate(self, allow_international_number: bool = False, allow_uk_landline: bool = False) -> None:
+    def _raise_if_service_has_block_ofcom_protected_block_but_tries_to(self, block_ofcom_protected_block: bool = False):
```
CrystalPea: could we rename it to `_raise_if_service_who_blocked_it_tries_to_send_to_ofcom_protected_ranges`

Then we don't use block twice, so it's easier to read, and general sentence structure is clearer 🙌🏼

CrystalPea (commented):
Hey, we could rename `ofcom_protected_block` to `ofcom_protected_ranges` throughout, then we don't have double block, and the code will be easier to read.

notifications_utils/recipient_validation/phone_number.py
rparke: ofcom uses the language of "blocks" and not ranges.

It's a balance between readability and accuracy of language used - I'm not super bothered which we use so long as it's consistent.

tests/recipient_validation/test_phone_number.py
```diff
@@ -468,6 +483,35 @@ def test_rejects_valid_uk_landlines_if_allow_landline_is_false(self, phone_numbe
...
+            ("07034777777", True),
+            ("07074971099", False),
+            ("07075971077", True),
+            ("07999999999", False),
+            # non-uk number
+            ("+1 202-483-3000", False),
+        ),
+    )
+    @pytest.mark.parametrize("block_ofcom_protected_block", [True, False])
+    def test_rejects_valid_uk_mobiles_if_in_ofcom_protected_block(
```
CrystalPea: This test name currently reads weird - as there is third person singular verb, but no subject for it. We could rename it to the same grammar structure as the test above it:`test_reject_valid_uk_mobiles_if_in_ofcom_protected_block` or add a subject `test_PhoneNumber_rejects_valid_uk_mobiles_if_in_ofcom_protected_block`

rparke: done

notifications_utils/recipient_validation/phone_number.py
CrystalPea: I would go for ranges, I think in this case readability will save us more time/energy than accuracy - `block the block` sounds confusing.

rparke: done

notifications_utils/recipient_validation/phone_number.py
```diff
@@ -103,9 +103,24 @@ def _raise_if_unsupported_country(self) -> None:
         if str(self.number.country_code) not in COUNTRY_PREFIXES:
             raise InvalidPhoneError(code=InvalidPhoneError.Codes.UNSUPPORTED_COUNTRY_CODE)
 
-    def validate(self, allow_international_number: bool = False, allow_uk_landline: bool = False) -> None:
+    def _raise_if_service_has_block_ofcom_protected_ranges_but_tries_to(
```
CrystalPea: thanks a lot for changing the block to ranges 🙌🏼 . Sorry to be a nitpick, but could we also change the sentence structure a bit for improved grammar, like so? 

`def _raise_if_service_who_blocked_it_tries_sending_to_ofcom_protected_ranges`

tests/recipient_validation/test_phone_number.py
```diff
@@ -468,6 +483,35 @@ def test_rejects_valid_uk_landlines_if_allow_landline_is_false(self, phone_numbe
...
+            ("07034777777", True),
+            ("07074971099", False),
+            ("07075971077", True),
+            ("07999999999", False),
+            # non-uk number
+            ("+1 202-483-3000", False),
+        ),
+    )
+    @pytest.mark.parametrize("block_ofcom_protected_ranges", [True, False])
+    def test_PhoneNumber_rejects_valid_uk_mobiles_if_in_ofcom_protected_block(
```
CrystalPea: if we want to be consistent throughout we need to change block to ranges in the test title, too

CrystalPea (commented):
Last two little things 🙌🏼

CrystalPea (approved):

PR #1469, opened 2026-09-15
Escape HTML in branding options of `HTMLEmailTemplate`

quis:
`brand_text`, `brand_colour`, etc can contain user-supplied data, either from the database or as query string arguments to [this endpoint in the admin app](https://github.com/alphagov/notifications-admin/blob/b325e53ff18bb9f05f725aefea2bd2391fe1d729/app/main/views/index.py#L69-L117).

We should escape things like `>` to prevent injection of malicious content. We can do this with the `escape_html` function which is already used on stuff like the `subject` and `preheader` (albeit via the `html` argument of `Field`):
https://github.com/alphagov/notifications-utils/blob/191f0c92af0d68e0adac8c278937e40248cb66f3/notifications_utils/template.py#L590

***

Some of these variables are inserted into attributes of HTML elements, for example:
https://github.com/alphagov/notifications-utils/blob/191f0c92af0d68e0adac8c278937e40248cb66f3/notifications_utils/jinja_templates/email_template.jinja2#L128

We therefore also need to escape quotes, so malicious content can’t break out of the attribute and mangle the HTML. Our `escape_html` function does not do this:
https://github.com/alphagov/notifications-utils/blob/191f0c92af0d68e0adac8c278937e40248cb66f3/notifications_utils/formatters.py#L178

This pull request adds a `quote` argument to `escape_html`, matching [Python’s built in `html.escape`](https://docs.python.org/3/library/html.html#html.escape). It defaults to `False` to preserve existing behaviour.

***

To make sure the escaping is working properly we do an extra check for valid HTML with `html5lib`, which the admin app [uses already](https://github.com/alphagov/notifications-admin/blob/b325e53ff18bb9f05f725aefea2bd2391fe1d729/tests/conftest.py#L3028-L3030).


rparke:
I think flakey test that I'm just re-running. Otherwise LGTM

rparke (approved):

PR #1472, opened 2026-09-18
Dependency update

whpearson:
Three things

- Removing a test that changes due to an underlying issue. Zimbawean processing is different since v9.0.38 https://github.com/google/libphonenumber/blob/master/release_notes.txt  and the error changed to too long from invalid. As we have tests for too long phone numbers already, this seemed redundant.
- Version bump
- Make refreeze-requirments


klssmith (approved):
We don't really need to update the version since we're only changing test requirements so there are no changes for people using the package, but since it's only used internally this doesn't affect anyone so I'm happy to approve

klssmith (approved):

